PR #6, opened 2020-05-04
Add Algolia search

dcramer:


dcramer:
Vercel isn't letting me set any environment variables, but I had already configured prod with (most of) the needed vars. Going to deploy so search is at least landed, but we'll have to add ALGOLIA_INDEX=1 to production yet.

PR #9, opened 2020-05-04
Enable bors

dcramer:


dcramer:
not really sure why its not working at this point

PR #13, opened 2020-05-04
Override default Link/anchor tag to automatically do external

dcramer:


dcramer:
This simply determines routing based on the URL, and automatically adds the external icon when needed.

PR #14, opened 2020-05-05
fix: Broken alerts and links

HazAT:


HazAT:
@dcramer not sure what's happening here but something is wrong with the permissions
image
(Even after approval)

HazAT:
image

mitsuhiko:
Wonder if this is bors doing.

dcramer:
disabled bors requirement til it works

PR #33, opened 2020-05-13
Feat/add using samples

markstory:
I often use these commands to create sample events locally for platforms I don't have applications for and to ensure I haven't broken minidump UI.


markstory:
> Couldn't we just load these up from the mock data?

Potentially. load-mocks currently only makes events for ruby, php, python, java and javascript.

PR #36, opened 2020-05-22
Add environment troubleshooting guide from notion

markstory:


src/docs/environment.mdx
```diff
@@ -164,3 +164,122 @@ If you would like to be able to run `devserver` outside of your root checkout, y
   <em>and</em> port are required in order for DNS and some pages urls to be
   displayed correctly.
 </Alert>
+
+
+## Troubleshooting
+
+**Problem:** something like `pkg_resources.DistributionNotFound: The 'some_dependency<0.6.0,>=0.5.5' distribution was not found and is required by sentry`
+
+**Solution:** Your virtualenv needs to be updated. In sentry, type `make install-sentry-dev`. In getsentry, type `make install-develop-requirements`.
```
joshuarli, line 173: note to self: I'm gonna unify those two commands to just be `install-dev`. There's no reason they should be different.

src/docs/environment.mdx
```diff
@@ -164,3 +164,122 @@ If you would like to be able to run `devserver` outside of your root checkout, y
...
+**Solution:**  You likely need to upgrade your Python dependencies. Go to the git root directory and run `make install-sentry-dev`.
+
+---
+
+**Problem:**  "out of memory" issues after running the sentry or getsentry devserver, or during webpacking.
+
+**Solution:**
+
+```bash
+export NODE_OPTIONS="--max-old-space-size=4096"
```
joshuarli, line 188: Additionally, we should plug direnv here. @dcramer wdyt?

src/docs/environment.mdx
```diff
@@ -164,3 +164,122 @@ If you would like to be able to run `devserver` outside of your root checkout, y
...
+  ...
+django.core.exceptions.ImproperlyConfigured:
+Application labels aren't unique, duplicates: okta
+make: *** [develop] Error 1
+```
+
+**Solution:**
+
+```jsx
+pip uninstall sentry-auth-github sentry-auth-saml2
```
joshuarli, line 239: This was a workaround mentioned in an email long ago during the great sentry-plugins migration, I'm fairly sure at this point that no one would have this problem. @scefali you think this can be removed?

src/docs/environment.mdx
```diff
@@ -164,3 +164,122 @@ If you would like to be able to run `devserver` outside of your root checkout, y
...
+00:51:27 server  | Python auto-reloader enabled
+00:51:27 server  |   File "/Users/dfuller/code/sentry/src/sentry/wsgi.py", line 11, in <module>
+00:51:27 server  |     from django.conf import settings
+00:51:27 server  |   File "/Users/dfuller/code/sentry/.venv/lib/python2.7/site-packages/django/conf/__init__.py", line 14, in <module>
+00:51:27 server  |     from django.core.exceptions import ImproperlyConfigured
+00:51:27 server  |   File "/Users/dfuller/code/sentry/.venv/lib/python2.7/site-packages/django/core/exceptions.py", line 5, in <module>
+00:51:27 server  |     from django.utils.encoding import force_text
+00:51:27 server  |   File "/Users/dfuller/code/sentry/.venv/lib/python2.7/site-packages/django/utils/encoding.py", line 10, in <module>
+00:51:27 server  |     from django.utils.functional import Promise
+00:51:27 server  |   File "/Users/dfuller/code/sentry/.venv/lib/python2.7/site-packages/django/utils/functional.py", line 1, in <module>
```
joshuarli, line 256: @markstory it might be good to `-- snip --` the lines above here, in the interest of leaner docs.

src/docs/environment.mdx
scefali: @joshuarli yea, we no longer use those libs

src/docs/environment.mdx
markstory: I think we should be leading to people to the paved roads. direnv seems to be working well for many people now. Using it lets us avoid the oom issue with node entirely and not have to worry about troubleshooting it ever again.

markstory:
> Looking at what's left in the dev handbook, are you planning on following up with a "howto" and/or "hacks" page (e.g. for the travis caches thing)?

I don't think this content belongs in the environment docs. Instead I would rather see us have a chapter on CI where we cover CI the various CI tools we use (travis, percy, github actions, zeus) and relevant troubleshooting information there.

Same for deployments. I feel that is another chapter that can live beside environment and CI in the handbook.  I'm leaning towards adding a new sidebar section for development lifecycle workflow but don't have a pithy name for it just yet, nor enough content put together to justify a separate sidebar section just yet.

src/docs/environment.mdx
joshuarli: @markstory ^ just in case you missed this

src/docs/environment.mdx
joshuarli: Cool. I haven't seen anyone bitten by this for a while now, so you can just remove this then.

src/docs/environment.mdx
joshuarli: https://github.com/getsentry/sentry/pull/19017
https://github.com/getsentry/getsentry/pull/3944

Will resolve this when you make the edits.

src/docs/environment.mdx
scefali: @joshuarli My bad, I read this over again. Even though we don't use the packages, the error in question only happens when you have duplicate installs. It's definitely possible that someone could still have this problem I think, though maybe less likely.

joshuarli: Probably way less likely, if I had to guess. We could remove this in the near future then.

src/docs/environment.mdx
```diff
@@ -164,3 +164,108 @@ If you would like to be able to run `devserver` outside of your root checkout, y
...
+Traceback (most recent call last):
+  ...
+django.core.exceptions.ImproperlyConfigured:
+Application labels aren't unique, duplicates: okta
+make: *** [develop] Error 1
+```
+
+**Solution:**
+
+```jsx
```
dcramer, line 238:
```suggestion
```shell
```

dcramer:
Can we put this as a different page? One issue this page had that I was solving for was the bloat. It tried to cover every random edge case vs just assuming a developer could handle their 9-5.

Think of this like an FAQ. You wouldn't embed the FAQ inside of the core doc when its boundless and noisy.

src/docs/environment.mdx
markstory: Fixed now.

src/docs/environment.mdx
markstory: Cool I'll remove this.

src/docs/environment.mdx
```diff
@@ -164,3 +164,80 @@ If you would like to be able to run `devserver` outside of your root checkout, y
...
+
+**Problem:** something like `pkg_resources.DistributionNotFound: The 'some_dependency<0.6.0,>=0.5.5' distribution was not found and is required by sentry`
+
+**Solution:** Your virtualenv needs to be updated. run `make install-py-dev`.
+
+---
+
+**Problem:** You see `Error occured while trying to proxy to: dev.getsentry.net:8000/`
+
+**Solution:**  You likely need to upgrade your Python dependencies. Go to the git root directory and run `make install-sentry-dev`.
```
joshuarli, line 179: You forgot to update to `install-py-dev` here too.

markstory: Whoops :blush:

joshuarli:
> Can we put this as a different page? One issue this page had that I was solving for was the bloat. It tried to cover every random edge case vs just assuming a developer could handle their 9-5.
> 
> Think of this like an FAQ. You wouldn't embed the FAQ inside of the core doc when its boundless and noisy.
> 
> (Also just saw Mark's comment, oops!)

> I'm leaning towards adding a new sidebar section for development lifecycle workflow but don't have a pithy name for it just yet, nor enough content put together to justify a separate sidebar section just yet.

@markstory How about "ERRATA", all the way down the sidebar, and then move Troubleshooting under there? Not the best name, but you can link to Troubleshooting in the dev environment docs.

markstory:
I think the troubleshooting section is ok where it is now. If it continues to grow I think we should split it off into a separate page, perhaps not in the sidebar that is linked to from the environment docs.

PR #44, opened 2020-05-28
feat: Improve testing docs

markstory:
Include content on the datetime helpers that make building events much easier in tests.

### DIFF
--- a/src/docs/testing.mdx
+++ b/src/docs/testing.mdx
@@ -106,7 +106,7 @@ This will help you simulate success and failure scenarios with relative ease.
 
 When writing tests related to ingesting events we have to operate within the
 constraint of events cannot be older than 30 days. Because all events must
-be recent, we cannot use traditional time freezin strategies to get consistent
+be recent, we cannot use traditional time freezing strategies to get consistent
 data in tests. Instead of choosing arbitrary points in time we work backwards
 from the present, and have a few helper functions to do so:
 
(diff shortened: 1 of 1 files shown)


src/docs/testing.mdx
```diff
@@ -102,6 +102,25 @@ def test_success(self):
 Use the `responses` library to add stub responses for an outbound API requests your code is making.
 This will help you simulate success and failure scenarios with relative ease.
 
+### Working with time reliably
+
+When writing tests related to ingesting events we have to operate within the
+constraint of events cannot be older than 30 days. Because all events must
+be recent, we cannot use traditional time freezin strategies to get consistent
```
Zylphrex, line 109: freezin -> freezing

src/docs/testing.mdx
```diff
@@ -102,6 +102,25 @@ def test_success(self):
...
+data in tests. Instead of choosing arbitrary points in time we work backwards
+from the present, and have a few helper functions to do so:
+
+```python
+from sentry.testutils.helpers.datetime import before_now, iso_format
+
+five_min_ago = before_now(minutes=5)
+iso_timestamp = iso_format(five_min_ago)
```
Zylphrex, line 117: Since we're talking about testing event ingestion, would it be reasonable to include how this `iso_timestamp` is used to set the time on an event with a `store_event` example?

markstory: I was hoping people would look for examples of store_event() I'm not sure its something we need in the docs right now.

PR #48, opened 2020-05-29
feat: Add AM guideline docs

HazAT:
CleanShot 2020-05-29 at 10 17 48


src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -0,0 +1,60 @@
...
+[https://github.com/getsentry/sentry-javascript/tree/master/packages/apm](https://github.com/getsentry/sentry-javascript/tree/master/packages/apm) for a reference implementation
+
+- New option for the SDK `tracesSampleRate`
+    - Defaults to `0.0`, which means send no transactions (0% delivery rate)
+    - `1.0` (100% delivery rate) means send all transactions
+    - Value is expressed in `0.0` - `1.0` float, ranged from 0% to 100% rate
+    - Default makes AM an opt-in feature
+- Introduce `Sentry.startTransaction`
+    - Internally this creates a `Transaction` and returns the instance
+    - Users need to interaction with the instance and keep track of it themselves
```
rhcarvalho, line 22:
```suggestion
    - Users need to interact with the instance and keep track of it themselves
```

src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -0,0 +1,60 @@
...
+    - Internally we can use the `Scope` to hold on the a reference of it if needed
```
rhcarvalho, line 23:
```suggestion
    - Internally we can use the `Scope` to hold on to a reference of the current `Transaction` if needed
```

src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -0,0 +1,60 @@
...
+- Introduce the new type `Span` → [https://develop.sentry.dev/sdk/event-payloads/span/](https://develop.sentry.dev/sdk/event-payloads/span/)
+- Introduce new type `Transaction` which inherits from `Span`
+- `Span` Interface
+    - When a `Span` is created, set the `startSimestamp` to the current time
```
rhcarvalho, line 27:
```suggestion
    - When a `Span` is created, set the `startTimestamp` to the current time
```

src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -0,0 +1,60 @@
...
+    - If there is already a `Span` on the current `Scope`, the created `Span` should be its child
+    - This method deals with sampling, and therefore it should take the `tracesSampleRate` option into account:
+        - If it's a child `Span`, we inherit the decision from the parent
+        - If it's a root span, we need to roll the dice here
+        - Depending on the outcome, the sample decision should be stored in the `Span`'s `sampled` property and again forwarded to its children
+- `Hub` → Introduce a method called `traceHeaders`
+    - This function returns a header (string) `sentry-trace`
+    - The value should be the trace header string of the `Span` that is currently on the `Scope`
+- `Scope` Introduce `setSpan`
+    - This can be used internally to pass a `Span` / `Transaction` around so integration can attach children to it
```
rhcarvalho, line 46:
```suggestion
    - This can be used internally to pass a `Span` / `Transaction` around so that integrations can attach children to it
```

src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -0,0 +1,60 @@
...
+    - The `transaction` property on the `Scope` (legacy) should overwrite the name of the `Transaction` if we send one. With that we give the user the option to change the name downstream if they don't have access to the instance of the `Transaction` directly.
```
rhcarvalho, line 47:
```suggestion
    - Setting the `transaction` property on the `Scope` (legacy) should overwrite the name of the `Transaction` stored in the `Scope`, if there is one. With that we give users the option to change the transaction name even if they don't have access to the instance of the `Transaction` directly.
```

src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -0,0 +1,60 @@
...
+- `Span.finish()`
+    - Just set `endTimestamp` to the current time (in payload `timestamp`)
+- `Transaction.finish()`
+    - `super.finish()` (call finish on Span)
+    - Send it to Sentry only if `sampled != true`
```
rhcarvalho, line 52:
```suggestion
    - Send it to Sentry only if `sampled == true`
```

src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -0,0 +1,60 @@
...
+    - `Transaction` needs to be wrapped in an `Envelope` and send through the [Envelope Endpoint](https://develop.sentry.dev/sdk/envelopes/)
```
rhcarvalho, line 53:
```suggestion
    - A `Transaction` needs to be wrapped in an `Envelope` and sent to the [Envelope Endpoint](https://develop.sentry.dev/sdk/envelopes/)
```

src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -0,0 +1,60 @@
...
+    - The `Transport` should use the same internal queue for `Transactions` / `Events`
+    - The `Transport` should implement category-based rate limiting →
```
rhcarvalho, line 55: Looks like this is missing a link after / instead of the arrow.

src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -0,0 +1,60 @@
...
+    - The `Transport` should deal with wrapping a `Transaction` in an `Envelope` internally
+- A `Transaction` should **not** go through `beforeSend` (but `EventProcessors` are fine for now, since they are considered internal)
+    - The distinctive feature of a `Transaction` is `type: "transaction"`
+    - Also, a few other properties see → Link to develop docs
```
rhcarvalho, line 59: Pending the link?

rhcarvalho:
@HazAT thanks for putting it all together ❤️

src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -0,0 +1,60 @@
...
+    - `Transaction` has additionally a `setName` method the set the name of the transaction
+    - `Transaction` receives a `TransactionContext` on creation (new property vs. `SpanContext` is `name`)
+    - Since a `Transaction` inherits a `Span` it has all functions available and can be interacted with like it was a `Span`
+- `Hub` → Introduce a method called `startSpan`
+    - Creates a new `Span` instance
+    - If there is already a `Span` on the current `Scope`, the created `Span` should be its child
+    - This method deals with sampling, and therefore it should take the `tracesSampleRate` option into account:
+        - If it's a child `Span`, we inherit the decision from the parent
+        - If it's a root span, we need to roll the dice here
+        - Depending on the outcome, the sample decision should be stored in the `Span`'s `sampled` property and again forwarded to its children
```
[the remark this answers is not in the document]
HazAT: It needs to be a property so if a XHR creates a span we need the decision on the Span in the header but not in the payload.

src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -0,0 +1,59 @@
...
+- `Scope` Introduce `setSpan`
+    - This can be used internally to pass a `Span` / `Transaction` around so that integrations can attach children to it
+    - Setting the `transaction` property on the `Scope` (legacy) should overwrite the name of the `Transaction` stored in the `Scope`, if there is one. With that we give users the option to change the transaction name even if they don't have access to the instance of the `Transaction` directly.
+- `Span.finish()`
+    - Just set `endTimestamp` to the current time (in payload `timestamp`)
+- `Transaction.finish()`
+    - `super.finish()` (call finish on Span)
+    - Send it to Sentry only if `sampled == true`
+    - A `Transaction` needs to be wrapped in an `Envelope` and sent to the [Envelope Endpoint](https://develop.sentry.dev/sdk/envelopes/)
+    - The `Transport` should use the same internal queue for `Transactions` / `Events`
```
bruno-garcia, line 54: Could you please elaborate on this? Why do we must use the same?
Although likely simpler, it would now require that we have a single max value that now affects tiny transactions and big events. If I understand correctly, errors in a tight loop could fill up the queue and start dropping trace data and break distribute traces.

src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -0,0 +1,59 @@
...
+    - The `Transport` should implement category-based rate limiting →
+    - The `Transport` should deal with wrapping a `Transaction` in an `Envelope` internally
+- A `Transaction` should **not** go through `beforeSend` (but `EventProcessors` are fine for now, since they are considered internal)
```
bruno-garcia, line 57: I'm afraid avoiding BeforeSend and going through EventProcessor might cause confusion.
Until today we've announced: "Events always go through BeforeSend" so the user has a chance to change it. Perhaps this just shows that Transactions are not Events?

A typed EventProcessor is our public API for anyone who wants to use an instance of a type that implements an interface and doesn't care to be executed only once the event is totally formed (only BeforeSend gives that guarantee). But saying Transaction events go through those callbacks could cause confusion since they'll be "weird events" and could behave strange in users current EventProcessors.

src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -0,0 +1,60 @@
...
+- `Hub` → Introduce a method called `startSpan`
+    - Creates a new `Span` instance
+    - If there is already a `Span` on the current `Scope`, the created `Span` should be its child
+    - This method deals with sampling, and therefore it should take the `tracesSampleRate` option into account:
+        - If it's a child `Span`, we inherit the decision from the parent
+        - If it's a root span, we need to roll the dice here
+        - Depending on the outcome, the sample decision should be stored in the `Span`'s `sampled` property and again forwarded to its children
+- `Hub` → Introduce a method called `traceHeaders`
+    - This function returns a header (string) `sentry-trace`
+    - The value should be the trace header string of the `Span` that is currently on the `Scope`
```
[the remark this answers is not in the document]
HazAT: This is only relevant for inter-integration communication. In JS we don't use it anywhere (right now) but for python we need it to attach this info for background queue integration.

And that doesn't mean we need to add it since the Spans also have the `toTraceparent` function which should be used to attach to outgoing requests.

src/docs/sdk/unified-api/tracing.mdx
HazAT: This is not ideal I know and maybe for now we need an internal "beforeTransaction" (we don't want to expose this either). But integrations need a way to still attach data or discard outgoing Transactions before sending. The workaround for this is to still let them go through eventProcessors but just not call the user function beforeSend.

src/docs/sdk/unified-api/tracing.mdx
HazAT: Yep, it's intended, if an SDK is so spammy that the queue can't keep up with sending it, it's fine to drop whatever.

Keep in mind, this is not offline caching.

src/docs/sdk/unified-api/tracing.mdx
bruno-garcia: The problem is that we'll trigger user defined `EventProcessor` too. Users do use that in .NET which is much preferable in cases of dependency injection over an inlined function (beforeSend).
Sounds to me that a Transaction isn't an Event but I'll resolve this as clarified.

src/docs/sdk/unified-api/tracing.mdx
bruno-garcia: Yes but we could drop the errors that are too spammy but keep the transaction flowing without any problems.
At that point it's possible the errors are getting rate-limited too, while the transactions are not.
If we use a single queue here, it won't matter that we have different rate limits, we'll drop transactions because the internal queue is full of errors.

I just wanted to voice my concern but if you're aware and have decided this, I'll resolve this too.

HazAT:
We can start iterating on this document.

Issue #51, opened 2020-06-03
Dev environment: zsh documentation


robinrendle:
I’m getting my dev environment setup and I’m using zsh so the very first step is confusing: 

![Screenshot 2020-06-03 12 40 25](https://user-images.githubusercontent.com/1900676/83681529-6c123a00-a597-11ea-9f93-ebf7320687c5.png)

My familiarity with the differences between zsh/bash are limited to say the least but I feel like I should still be able to setup my env properly. What file should I be updating here? I understand that I need to update `~/.zshrc` with `eval "$(pyenv init -)"` but then how do I reload the command line if I’m using zsh?

dcramer:
Just needs updated to have a zsh example, but you can completely ignore this, as there's only `.zshrc`

PR #52, opened 2020-06-03
fix: use correct pyenv command

AbhiPrasad:
We now have to use `make setup-pyenv` when we are setting up our development environment. 

See more details with Josh's PR on why we do this: https://github.com/getsentry/sentry/pull/18906

Also took the time to add a `zsh` snippet ~and add a troubleshooting prompt for python version issues.~

Closes https://github.com/getsentry/develop/issues/51 as it deals with the main issue.

### DIFF
--- a/src/docs/environment.mdx
+++ b/src/docs/environment.mdx
@@ -27,7 +27,7 @@ You can verify that Docker is running by running `docker ps`. If it doesn't erro
 
 We utilize [pyenv](https://github.com/pyenv/pyenv) to install and manage python versions. It should have already been installed earlier when you ran `brew bundle`.
 
-To install the required versions of Python you'll need to run `pyenv install`. This will take a while, since your computer is actually compiling python!
+To install the required versions of Python you'll need to run `make setup-pyenv`. This will take a while, since your computer is actually compiling python!
 
 After this, if you type `which python`, you should see something like `/usr/bin/python`... this means `python` will resolve to the system's python. You'll need to make some manual changes to your shell initialization files, if you want your shell to see pyenv's python.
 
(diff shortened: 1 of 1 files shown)


src/docs/environment.mdx
```diff
@@ -180,6 +185,12 @@ If you would like to be able to run `devserver` outside of your root checkout, y
 
 ---
 
+**Problem:** Some kind of python version issues like `python-build: definition not found ...` or `pyenv: version 'x.x.x' is not installed` 
+
+**Solution** Make sure you have the correct python versions installed. Go to the git root directory and run `make setup-pyenv`.
```
joshuarli, line 190: Since the dev env docs are now updated, there's a very low chance anyone would run into this issue. So I'd recommending just removing this to keep things slimmer.

Or, you could just change the solution to "Run `make setup-pyenv`." People know make is in git root.

src/docs/environment.mdx
```diff
@@ -91,7 +90,14 @@ exec "$SHELL"
...
 
-```bash
+```bash {filename: ~/.bashrc}
+eval "$(pyenv init -)"
+
+export VOLTA_HOME="~/.volta"
+grep --silent "$VOLTA_HOME/bin" <<< $PATH || export PATH="$VOLTA_HOME/bin:$PATH"
+```
+
+```bash {filename: ~/.zshrc} {tabTitle: Zsh}
```
joshuarli, line 100: cc @robinrendle There's your change. Thanks Abhijeet!

src/docs/environment.mdx
AbhiPrasad: Good point, I can remove it, we can always add it back if lot's of people end up hitting the issue.

PR #53, opened 2020-06-05
Add Sentry plugin

dcramer:
This is pulled in from https://github.com/dcramer/dcramer.github.com, and is based on the official gatsby-plugin-sentry (which cannot support APM as built).


dcramer:
oops forgot deps

PR #54, opened 2020-06-05
Enable typescript

dcramer:


.eslintrc.js
```diff
@@ -0,0 +1,41 @@
+module.exports = {
```
EvanPurkhiser, line 1: Would be nice to use the sentry eslint config

PR #60, opened 2020-06-09
build: upgrade node to match node version in sentry (v10.16.3)

joshuarli:
Not too keen on downloading another node version to develop, so bump node to match the version I already use for Sentry.

This doesn't break anything locally for me.

Right now, I don't know if this needs to be bumped elsewhere.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "react-test-renderer": "^16.13.1"
   },
   "volta": {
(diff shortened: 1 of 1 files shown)


dcramer:
tho to be fair, it basically does this for you and you don't have to think about it, so I wouldn't worry about this in the future when versions change

PR #73, opened 2020-07-03
Slack integration guide

phyzical:
* added quickstart to readme
* added links to new slack integration doc page
* added images incase they are removed from discord cdn
* added guide contents and added chunk on missing config
* added init command to simplify things further

Addresses things mentioned in https://github.com/getsentry/sentry/pull/19446


src/docs/integrations/slack/index.mdx
```diff
@@ -0,0 +1,89 @@
...
+Here’s where you’ll connect your on-premise Sentry instance to your newly created Slack app.
+<dl>
+<dt><code>system.admin-email</code></dt>
+Declared in <code>config.yml</code>.
+
+<dd>
+<markdown>
+Copy your Client ID, Client Secret, and Verification Token.
+
+Set slack.legacy-app to False unless you are supporting an old bot integration (Pre June 2020)
```
BYK, line 24: Now that I think about this, I think this setting should default to `false` on Sentry. @MeredithAnya, thoughts?

src/docs/integrations/slack/index.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+```yaml
+slack.client-id: <'client id'>
```
BYK, line 27: Why do we have quotes around this and not others? 😀

phyzical: i copied verbatim, but i remember this actually helped me initially as this comes in the format 'asd3h3.534b453hg4' so without the quotes it would violate the config (i think?..)

src/docs/integrations/slack/index.mdx
phyzical: i thought the same, but as others mentioned its to avoid breaking existing installs

phyzical:
no worries :) figured i could do something to give back only took 10 minutes of my time after all

src/docs/integrations/slack/index.mdx
MeredithAnya: hmm if there is a risk at all of breaking existing installs, I would want to avoid that. I almost think it would be clearer to say, if you are creating a slack app after June 2020 set it to False, cause I don't think its about the bot app being old, its about the scopes that were requested during installation. Slack's "old" bot requests the bot scope which is like old old and maybe confuse people?

BYK: Yeah, that's true. I assumed a fresh Sentry installation meant a fresh Slack bot so we could at least have this set to `False` by default in the example config of onpremise. That said the Slack bot can still (and likely?) to be older, right?

BYK:
On your court @MeredithAnya or @scefali as the Ecosystem team.

src/docs/integrations/slack/index.mdx
```diff
@@ -0,0 +1,89 @@
...
+slack.client-secret: <client secret>
+slack.verification-token: <verification token>
```
scefali, line 29: @phyzical So we should actually now be telling people to use the `slack.signing-secret` instead of `slack.verification-token` since it's more secure. It's something that we added recently. Do you mind making this change?
 
Screen Shot 2020-07-08 at 2 33 18 PM

scefali:
Everything looks good to me! Just left one comment about the signing secret

src/docs/integrations/slack/index.mdx
phyzical: i suppose your right but we already merged that change in the config example 😄 Maybe we remember next time were touching the config example we can default to false

ill rejig the wording to match the config example

src/docs/integrations/slack/index.mdx
phyzical: hmm i did not notice that deprecation warning. i will give it a test later today if i get a chance then update the docs accordingly

src/docs/integrations/slack/index.mdx
phyzical: infact ill make another pr to add the `signing-secret` and enforce a default of false at same time over at `onpremise`

src/docs/integrations/slack/index.mdx
phyzical: i gave it a quick test, seemed to sync the integration again, so updated docs accordingly

phyzical:
hey @MeredithAnya @scefali 

Ready for review whenever  :)

src/docs/integrations/slack/index.mdx
phyzical: alight pr open over there :) this should be save to merge if all are happy

scefali:
@phyzical Thanks for your help! Merged!

PR #76, opened 2020-07-06
Add Mermaid + basic architecture diagram

dcramer:


dcramer:
I've made it 20 years without learning to flow chart.. probably not gonna start now

That said:

image

PR #77, opened 2020-07-06
Label routing endpoints

dcramer:


dcramer:
fwiw im not quite sure how this applies to self-hosted.. is it just a path prefix that goes to ingest?

src/docs/architecture.mdx
```diff
@@ -7,8 +7,8 @@ title: Architecture
 ```mermaid
 graph TD
 app[Your Application] --> |SDK| lb{{Load Balancer}}
-lb --> relay
-lb --> sentry_web["Sentry (web)"]
+lb --> |"ingest.sentry.example.com"| relay
```
jan-auer, line 10: For onpremise and local dev, we usually have the same domain for ingest and the web API. There is an optional setting to add the ingest domain. I'd propose to leave the same domain here. 

Separately, it's Relay that handles the store endpoint, so probably we should add that here instead of web:

```suggestion
lb --> |"sentry.example.com/api/\d+/store/"| relay
```

src/docs/architecture.mdx
```diff
@@ -7,8 +7,8 @@ title: Architecture
...
+lb --> |"sentry.example.com/api/\d+/store/"| sentry_web["Sentry (web)"]
```
jan-auer, line 11: Counterpart to the above comment: This is what handles the web API and serves static routes, so we can show the root URL here:

```suggestion
lb --> |"sentry.example.com"| sentry_web["Sentry (web)"]
```

dcramer:
:+1: thanks folks

PR #83, opened 2020-07-13
chore: Make it clear when not to use our test base classes

untitaker:


src/docs/testing.mdx
```diff
@@ -38,11 +38,15 @@ managing services.
 For python tests we use [pytest](https://docs.pytest.org/en/latest/) and testing tools
 provided by Django. On top of this foundation we've added a few base test cases (in `sentry.testutils.cases`).
 
-Endpoint integration tests is where the bulk of our test suite is focused. These tests help us ensure that
+Endpoint integration tests based on `APITestCase` is where the bulk of our test suite is focused. These tests help us ensure that
 the APIs our customers, integrations and front-end application continue to work in expected ways. You should
 endeavour to include tests that cover the various user roles, and cross organization/team access scenarios,
 as well as invalid data scenarios as those are often overlooked when manually testing.
 
+* **Use `unittest.TestCase` if you do not need access to the database.** Those tests will set up and teardown very quickly.
```
dcramer, line 46: I'm not in love with this advice as its a great foot gun. Its like a "if you are smart af and know the ins and outs, you can use unittest, if you're unsure, stay far away from it".

Also is the database the _only_ case? Are redis/caches/etc coupled to TestCase or just generic hooks?

joshuarli: > Also is the database the only case? Are redis/caches/etc coupled to TestCase or just generic hooks?

That's a good question.

Regarding `unittest.TestCase`...would those concerns be alleviated if https://github.com/getsentry/sentry/pull/19845 switched to https://docs.djangoproject.com/en/1.11/topics/testing/tools/#django.test.SimpleTestCase?

dcramer: Looks much safer

untitaker: I feel like it's really hard to shoot yourself into the foot with this. All stuff on `self` is missing (postWithHeader, `self.<model>`, factories) and all the regular db access like creating models is explicitly blocked by pytest-django (with a misleading suggestion but still):

>Failed: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.

If you want I can try to figure out how to rewrite that suggestion to link to dev docs.

---

wrt caches/redis, they're not really coupled but definetly much more of a potential footgun. redis in particular is currently not flushed consistently in any test base I could find.

untitaker: btw we have quite a bit of unittest.TestCase usage already in the suite.

untitaker:
@dcramer feel free to close if you think it's not necessary

untitaker:
closing because I am not sure what to do about redis setup/teardown. I still think we should avoid wiping the db if we're not going to use it. It saves a lot of time.

PR #86, opened 2020-07-15
ci(license): Add automatic BSL date update via GHA

BYK:


.github/workflows/license.yml
```diff
@@ -0,0 +1,22 @@
+name: update-license
+on:
+  repository_dispatch:
+    types: [update-license]
+  schedule:
+    # This is the typical CalVer release time
+    - cron:  '0 18 15 * *'
```
jan-auer, line 7: May I suggest to move that to `0 17` just like other services? If we choose to bundle docs into the release, we want this to come in ahead of time.

BYK: We don't really bundle them in but sure. This aligns with onpremise release time, not the others.

PR #87, opened 2020-07-15
feat: add sync-getsentry into new workflow section

joshuarli:
Also, some minor refactor + shuffling around.

### DIFF
--- a/src/docs/workflow.mdx
+++ b/src/docs/workflow.mdx
@@ -2,14 +2,5 @@
 title: "Development Workflow"
 ---
 
-## Settings/Configuration
+You may also be interested in <Link to="/sentry-vs-getsentry/">Sentry vs Getsentry</Link>.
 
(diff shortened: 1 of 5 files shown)


src/docs/workflow.mdx
```diff
@@ -0,0 +1,19 @@
+---
+title: "Workflow"
+---
+
+You may also be interested in <Link to="/sentry-vs-getsentry/">Sentry vs Getsentry</Link>.
+
+
+## Testing Sentry commits in Getsentry
```
untitaker, line 8: I feel like this might as well go into the CI section, otherwise this page feels kinda empty

joshuarli: Ah, that's an equally suitable place. I wanted to expand the workflow section in the near future though (how to bump python dependencies).

joshuarli: I'll merge this first, then decide later.

PR #88, opened 2020-07-16
fix: event.logger attribute is optional

rhcarvalho:
Depending on platform, the logger is not always available/known. It has been an optional attribute in practice, document it as such.


rhcarvalho:
See also https://github.com/getsentry/sentry-javascript/issues/1851#issuecomment-658159196

PR #96, opened 2020-08-03
doc: PoC of loading event schema docs from Relay

untitaker:
~This currently fetches the JSON Schema straight from Relay docs when loading the page, while in practice we should find a way to make builds more reproducible.~

Adds relay as submodule


untitaker:
I have requested access for dependabot to this repo, some admin got bugged now

PR #97, opened 2020-08-06
fix(docs): Add to getsentry troubleshooting

maheskett:
- adds `mock-subscription` to the getsentry setup docs and troubleshooting section since this is a common question
- updates to getsentry's Makefile are in [#4303](https://github.com/getsentry/getsentry/pull/4303), which adds `mock-subscription` to `make bootstrap` so it is run on initial setup, instead of only `make reset-db`

### DIFF
--- a/src/docs/environment/index.mdx
+++ b/src/docs/environment/index.mdx
@@ -209,10 +209,16 @@ Note that you **cannot** have both sentry and getsentry devserver running at the
 After the server warms up for a little while, you should be able to access it
 at [http://dev.getsentry.net:8000](http://dev.getsentry.net:8000/).
 
+If you see `DoesNotExist: Subscription matching query does not exist` in your dev server,
+run the following in getsentry:
+```bash
+./bin/mock-subscription
+```
+
 You can set your local instance's org to use a business plan by running the
 following in getsentry:
 
-```go
+```bash
 ./bin/mock-subscription <org_slug> --plan mm2_a_500k
 ```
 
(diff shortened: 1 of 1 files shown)


src/docs/environment/index.mdx
```diff
@@ -208,10 +208,16 @@ Note that you **cannot** have both sentry and getsentry devserver running at the
...
+If you see `DoesNotExist: Subscription matching query does not exist` in your dev server,
+run the following in getsentry:
+```bash
+./bin/mock-subscription
+```
+
 You can set your local instance's org to use a business plan by running the
 following in getsentry:
 
-```go
```
joshuarli, line 214: lol

src/docs/environment/index.mdx
```diff
@@ -292,3 +298,24 @@ In your sentry virtualenv:
...
+
+```bash
+./bin/mock-subscription <org_slug>
+```
+
+---
+
+**Problem:** Something like `Error: No such container: sentry_postgres`
+
+**Solution:** Review the [bootstrap services](http://localhost:8000/environment/#bootstrap-services)
```
joshuarli, line 316:
```suggestion
**Solution:** You need <Link to="/services/devservices/">devservices</Link>.
```

joshuarli: After you merge this PR I think it's better to just bake the link to develop docs inside of sentry to make that error more helpful. I can follow up with that.

joshuarli: Oh wait, I didn't see the remaining text. Just change the link away from localhost.

maheskett: lol will fix

src/docs/environment/index.mdx
```diff
@@ -293,3 +299,24 @@ In your sentry virtualenv:
...
+
+```bash
+./bin/mock-subscription <org_slug>
+```
+
+---
+
+**Problem:** Something like `Error: No such container: sentry_postgres`
+
+**Solution:** Review the [bootstrap services](/environment/#bootstrap-services)
```
joshuarli, line 317: Dang, I remember markdown link notation didn't work before, but it works on the preview.

PR #104, opened 2020-08-10
fix(dev environment): Add brew flag so progress is shown

lobsterkatie:
This addresses two problems I ran into when running through the dev environment instructions while setting up a new laptop.

### DIFF
--- a/src/docs/environment/index.mdx
+++ b/src/docs/environment/index.mdx
@@ -19,7 +19,7 @@ You're going to be working out of this repository for the remainder of the setup
 
 You'll need to first install Xcode cli tools. (If you try running `git` or `make`, you'll get an xcrun error. Run `xcode-select --install` and follow the instructions.
 
-Once you have Xcode installed, install [Homebrew](http://brew.sh), and then run `brew bundle` to install the various system packages as listed in sentry's `Brewfile`.
+Once you have Xcode installed, install [Homebrew](http://brew.sh), and then run `brew bundle` to install the various system packages as listed in sentry's `Brewfile`. This may take a few minutes, and brew isn't great about showing a progress bar, so it may look like the process has hung. Have faith and let it run.
 
 One thing that requires manual attention is `docker`, which should have just been installed. Open up Spotlight, search for "Docker" and start it. You should soon see the docker icon in your macOS menubar. Docker will automatically run on system restarts, so this should be the only time you do this.
 
@@ -112,7 +112,9 @@ Now, if you try and run `volta`, you should see some help text, meaning volta is
 node -v
 ```
 
-Volta intercepts this and automatically downloads and installs the node version in sentry's `package.json`.
+Volta intercepts this and automatically downloads and installs the node and yarn versions in sentry's `package.json`.
+
+Note that if you didn't already have node and yarn installed on your computer prior to this step, you'll need to run `volta install node@x.x.x` and `volta install yarn@x.x.x` when outside the `sentry` directory in order to use `node` and `yarn` elsewhere.
 
 ## Bootstrap Services
 
--- a/src/docs/environment/index.mdx
+++ b/src/docs/environment/index.mdx
@@ -114,7 +114,6 @@ node -v
 
 Volta intercepts this and automatically downloads and installs the node and yarn versions in sentry's `package.json`.
 
-Note that if you didn't already have node and yarn installed on your computer prior to this step, you'll need to run `volta install node@x.x.x` and `volta install yarn@x.x.x` when outside the `sentry` directory in order to use `node` and `yarn` elsewhere.
 
 ## Bootstrap Services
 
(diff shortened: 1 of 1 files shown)


src/docs/environment/index.mdx
```diff
@@ -19,7 +19,7 @@ You're going to be working out of this repository for the remainder of the setup
 
 You'll need to first install Xcode cli tools. (If you try running `git` or `make`, you'll get an xcrun error. Run `xcode-select --install` and follow the instructions.
 
-Once you have Xcode installed, install [Homebrew](http://brew.sh), and then run `brew bundle` to install the various system packages as listed in sentry's `Brewfile`.
+Once you have Xcode installed, install [Homebrew](http://brew.sh), and then run `brew bundle` to install the various system packages as listed in sentry's `Brewfile`. This may take a few minutes, and brew isn't great about showing a progress bar, so it may look like the process has hung. Have faith and let it run.
```
joshuarli, line 22: Oh actually, there's a progress bar with `brew bundle --verbose`. Could you just change to that?

joshuarli, line 22:
```suggestion
Once you have Xcode installed, install [Homebrew](http://brew.sh), and then run `brew bundle --verbose` to install the various system packages as listed in sentry's `Brewfile`. 
```

src/docs/environment/index.mdx
```diff
@@ -112,7 +112,9 @@ Now, if you try and run `volta`, you should see some help text, meaning volta is
 node -v
 ```
 
-Volta intercepts this and automatically downloads and installs the node version in sentry's `package.json`.
+Volta intercepts this and automatically downloads and installs the node and yarn versions in sentry's `package.json`.
+
+Note that if you didn't already have node and yarn installed on your computer prior to this step, you'll need to run `volta install node@x.x.x` and `volta install yarn@x.x.x` when outside the `sentry` directory in order to use `node` and `yarn` elsewhere.
```
joshuarli, line 117: Why is this necessary? Ideally I'd like to only provide documentation relevant to the dev environment. This is just how to use volta which anyone can read about at their leisure.

If there's other sentry-related repositories you needed non-volta node/yarn on, I can take a look.

joshuarli, line 117:
```suggestion
```

src/docs/environment/index.mdx
dcramer: Why does that matter? Let's not be nit picky about this kind of thing. If folks want to change behavior of things they can figure it out for themselves. We don't need to explain all of technology in our Sentry-specific docs.

src/docs/environment/index.mdx
lobsterkatie: Yeah, fair, I guess. Was just a hiccup I encountered, so thought it might be nice to fix it. But I see your point.

src/docs/environment/index.mdx
dcramer: I'll use this as a reminder in that I already re-wrote these pages once to remove peoples opinions (and a LOT of exces boilerplate text) about what others should do. Let's keep this to just straight forward "heres how you do something" without adding a ton of extra content for a persons particular flavor of behavior.

src/docs/environment/index.mdx
dcramer: Same thing here, why does anyone need to care about this. Its not related to Sentry and this isnt generic "how to use node" or "how to use Volta" docs.

src/docs/environment/index.mdx
lobsterkatie: This is hardly "explaining all of technology." The vast majority of CLI tools provide progress bars for long processes, so when one doesn't, it's possible it's hung. Is there any harm in adding `--verbose`, and making it so someone and can just cut and paste the commands and still see immediately that it's not hung up?

dcramer: The docs should be concise. They should explain the minimal steps needed to get up and working correctly, without adding additional complexity or repeating themselves (unless absolutely needed). This page used to be a mess of both of those problems. My point is we should not be re-adding that kind of flavor text. For example, I've never once had an issue running 'brew install' because it "didn't show a progress bar", so why do we feel we must force this knowledge on people? How's that any different (or better) than putting a big old "hey almost no CLIs show progress bars, so just wait until they're finished"? That's explaining technology, not explaining how to setup Sentry.

dcramer: Aside, I was reviewing the old PR I guess, but my point is still the same when we make changes to docs.

lobsterkatie: I see your point in general, but I also feel like if adding a flag makes the process more seamless for people, it's worth it. Each time I've gone through this process, I've had the, "Wait, is it frozen? Hmmm... Oh, right, damn, brew is terrible about progress bars. Okay, fine, I'll wait." conversation with myself. A blocker? Hardly. But it's a bump, so would be nice, IMHO, to smooth it out for folks.

dcramer: I'm fine with the `--verbose` flag (though I'd argue this is still not needed - again, assume people understand how to computer), my bigger feedback was when there was much more boilerplate.

lobsterkatie: Okay. We can agree to disagree about the value of the flag. 🙂

The rest is gone now, so no issue there.

PR #117, opened 2020-08-28
fix: Remove span.transaction from API docs

rhcarvalho:
This was never a proper part of the protocol as understood and
implemented in Relay.

The transaction name is sent as a top-level field in the event payload,
it is not part of a span.

### DIFF
--- a/src/docs/sdk/event-payloads/properties/status.mdx
+++ b/src/docs/sdk/event-payloads/properties/status.mdx
@@ -1,7 +1,6 @@
 `status`
 
-: _Optional_. Describing the `status` of the Span. We use this property to determine if the [Transaction](/sdk/event-payloads/properties/transaction/)
-ended in an error or not.
+: _Optional_. Describes the `status` of the Span/Transaction.
 
 | State                        | Description                                                                                                  | HTTP status code equivalent |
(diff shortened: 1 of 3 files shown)


src/docs/sdk/event-payloads/properties/status.mdx
```diff
@@ -1,7 +1,6 @@
 `status`
 
-: _Optional_. Describing the `status` of the Span. We use this property to determine if the [Transaction](/sdk/event-payloads/properties/transaction/)
-ended in an error or not.
+: _Optional_. Describes the `status` of the Span/Transaction.
```
rhcarvalho, line 3: Had to remove the link because the property now doesn't exist. This appears on both Spans and Transactions and is not only about "errors", so I've trimmed the description.

AbhiPrasad: Can we add a note similar to https://github.com/getsentry/relay/blob/569678659bac656d6c4a6edcbdf3381677b5eaa2/relay-common/src/constants.rs#L153?

AbhiPrasad: Oh wait merged 😄 we can get to this later then.

PR #126, opened 2020-09-02
ref(feature-flags): Document the process for graduation

BYK:
See https://www.notion.so/sentry/When-to-turn-on-features-for-OSS-c8859ec556a94d88901054e12569592a


src/docs/feature-flags.mdx
```diff
@@ -227,9 +230,10 @@ class TestFeatureFeatureHandler(features.FeatureHandler, PercentageRollout):
 
 ## After launch
 
-After your feature has been mainlined and is available for all hosted customers.
-You should go back and enable the feature in `conf/server.py` which will enable
-the feature for on-premise users.
+After your feature has been mainlined and is available for all hosted customers:
```
BYK, line 233: This is the main change, rest is mostly the formatter and some grammatical fixes.

src/docs/feature-flags.mdx
```diff
@@ -2,7 +2,7 @@
 title: 'Feature Flags'
 ---
 
-Feature flags are controlled in Sentry at the commit and configuration level. At this time there is built-in way to modify them in production in real-time.
+Feature flags are controlled in Sentry at the commit and configuration level. At this time there is no built-in way to modify them in production in real-time.
```
jan-auer, line 5:
```suggestion
Feature flags are controlled in Sentry at the commit and configuration level. At this time, there is no built-in way to modify them in production in real-time.
```

src/docs/feature-flags.mdx
```diff
@@ -118,12 +119,20 @@ a organization/project
...
+const toRender = (
+  <Feature features={['test-feature']}>
+    <MyComponentToFlag />
+  </Feature>
+);
+=======
 <Feature features={['test-feature']}>
   <MyComponentToFlag />
 </Feature>;
+>>>>>>> master
```
jan-auer, line 132: This looks like an unresolved merge conflict, or is this intended?

src/docs/feature-flags.mdx
```diff
@@ -227,9 +236,10 @@ class TestFeatureFeatureHandler(features.FeatureHandler, PercentageRollout):
...
-the feature for users of self-hosted Sentry.
+After your feature has been mainlined and is available for all hosted customers:
 
-If your feature switch is doesn't have an 'off' mode you should go back and
-remove the feature switch checks and feature flag as feature flags are not free.
+If your feature flag doesn't have an 'off' mode, remove it and all related checks.
+Otherwise, disable the feature by default in getsentry in `getsentry/conf/settings/defaults.py`,
+add the necessary conditional in `getsentry/features.py`, and then enable it by default in
+[`conf/server.py`](https://github.com/getsentry/sentry/blob/master/src/sentry/conf/server.py) to enable
+the feature for users of self-hosted Sentry.
```
jan-auer, line 245: I think we can give some more guidance on the getsentry logic. I'm thinking something like this:

---

- If the feature cannot be disabled, or if no application specific overrides are required, remove it and all related checks from the Sentry code base. If necessary, also remove references to the feature from the onpremise and getsentry repositories.
- If the feature controls behavior specific to Sentry SaaS, leave it disabled in Sentry and add appropriate feature handlers in getsentry (see below). In this case, the feature is also disabled during development.
- If the feature is generally available but is constrained to specific organizations or projects on Sentry SaaS, enable the feature by default in in [`conf/server.py`](https://github.com/getsentry/sentry/blob/master/src/sentry/conf/server.py) and add appropriate feature handlers in getsentry. This ensures that the feature is enabled in development and self-hosted Sentry installations.

To override features in getsentry, follow these steps:

1. Disable the feature in `getsentry/conf/settings/defaults.py` by updating `SENTRY_FEATURES`.
2. Add a new feature handler class in `getsentry/features.py` that determines availability of the feature based on the organization or actor.
3. Register the handler at the bottom of `getsentry/features.py`.

BYK: Replaced the paragraph with your suggestion verbatim. Thanks a lot for the great write-up!

src/docs/feature-flags.mdx
BYK: Done.

PR #134, opened 2020-09-07
feat(pii): Import PII docs from Relay

jan-auer:
Imports the PII format docs from Relay into the "Application" section.

### DIFF
--- a/src/components/sidebar.js
+++ b/src/components/sidebar.js
@@ -1,17 +1,17 @@
-import {withPrefix} from 'gatsby';
-import React from 'react';
-import {useLocation} from '@reach/router';
+import { withPrefix } from "gatsby";
+import React from "react";
+import { useLocation } from "@reach/router";
 
-import SmartLink from './smartLink';
+import SmartLink from "./smartLink";
 
-const NavLink = ({to, title, children, ...props}) => {
+const NavLink = ({ to, title, children, ...props }) => {
   const location = useLocation();
 
-  let className = 'toc-item';
+  let className = "toc-item";
   if (location && location.pathname.indexOf(withPrefix(to)) === 0) {
-    className += ' toc-active';
+    className += " toc-active";
   }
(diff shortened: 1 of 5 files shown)


src/components/sidebar.js
```diff
@@ -77,6 +77,11 @@ const Sidebar = () => (
         <NavLink to="/config/">Configuration</NavLink>
         <NavLink to="/feature-flags/">Feature Flags</NavLink>
         <NavLink to="/serializers/">Serializers</NavLink>
+        <NavLink to="/pii/" title="PII and Data Scrubbing">
```
untitaker, line 80: I think this is too high-level, can we add a Relay service page and nest it there?

jan-auer: I was thinking about this and decided against, but happy to reconsider. There will be a Relay page in the future, but it seems to me that the PII config format is a larger concern than Relay.

untitaker (approved):

src/components/sidebar.js
untitaker: i don't think it's on the same level as serializers and featureflags, nobody needs this stuff in their day-to-day work

jan-auer: It seems like there's no good place for this at the moment, overall. Especially, the Relay page will be focussed around the service nature of Relay (see where the menu item is located at the moment). 

Also, if we do this, then there will be 3 levels of navigation, which is a bit much, wdyt?

jan-auer: > nobody needs this stuff in their day-to-day work
Fair, let me see how it looks in a Relay page.

jan-auer: Okay, I agree that this will need another place. Shipping here now to get the docs live. In a follow-up, we'll document other concerns of event ingestion:
- endpoints and limitations
- rate limits
- filters
- normalization
- _move PII into that structure_

PR #137, opened 2020-09-08
ref(services): Update/resolve relay port conflict

billyvg:
This updates and resolves the relay port conflict. See https://github.com/getsentry/sentry/pull/20668


billyvg:
@untitaker should we add a "project" column too?

untitaker:
would link in the second column. layout breaks with too many columns

untitaker (approved):

PR #142, opened 2020-09-16
fix: Inclusive language on dependencies page

jan-auer:


marandaneto (approved):

src/docs/python-dependencies.mdx
```diff
@@ -12,7 +12,7 @@ So here are the rules we worked with so far:
 
 ## Rules on Dependencies
 
-1. Any new dependency needs to be thoroughly reviewed and whitelisted
+1. Any new dependency needs to be thoroughly reviewed and allowed
```
rhcarvalho, line 15:
```suggestion
1. Any new dependency needs to be thoroughly reviewed and selected
```

I think "allow" doesn't work so well in this context.

jan-auer: It seems like it should mean rather "approved" than "selected", let me change to that.

jan-auer: Updated.

mitsuhiko (approved):

PR #147, opened 2020-09-22
feat: Add from/to traceparent

HazAT:
This came up in this PR and in Discord
https://github.com/getsentry/sentry-php/pull/1092

We should document to have this.

### DIFF
--- a/src/docs/sdk/unified-api/tracing.mdx
+++ b/src/docs/sdk/unified-api/tracing.mdx
@@ -113,7 +113,7 @@ The sampling section is optional. The format is not `flags` to simplify processi
 ```
 
 Differently than `b3` a simple sampling decision as a value to `sentry-trace` should not be implemented. [There are reasons to always include](https://github.com/apache/incubator-zipkin-b3-propagation/blob/bc937b6854ea30e46b3e85fbf147d8f4de685dd5/README.md#why-send-trace-ids-with-a-reject-sampling-decision) the `trace-id` and `span-id` regardless of sampling having been decided by the caller. And also this will simplify the implementation.
-Besides the [usual reasons to use *defer](https://github.com/apache/incubator-zipkin-b3-propagation/blob/bc937b6854ea30e46b3e85fbf147d8f4de685dd5/README.md#why-defer-a-sampling-decision),* in the case of Sentry, a reason would be if a downstream system captures an error event with Sentry. The decision could be done at that point to *not* sample that trace in order to have tracing data available for the reported crash. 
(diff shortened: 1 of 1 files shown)


src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -70,6 +70,8 @@ tree as well as the unit of reporting to Sentry.
     - When a `Span` is created, set the `startTimestamp` to the current time
     - `SpanContext` is the attribute collection for a `Span` (Can be an implementation detail)
     - The relation between parent - child is captured in the property `parentSpanId`
+    - `Span` should have a method called `toTraceparent` which returns a string `sentry-trace` that could be sent as a header
+    - Similar `SpanContext` should have a static method called `fromTraceparent` which prefills a `SpanContext` with data received from a `sentry-trace` string
```
rhcarvalho, line 74: - `SpanContext` is an implementation detail, as indicated above, so I'd not define required methods on it. `from/toTraceparent` should be in `Span` and `Transaction`. JS defines both methods on `Span` and they are inherited by `Transaction`.
- We need to document the format of the string used for the `sentry-trace` header.

HazAT: I would say `SpanContext` while implementation detail we can come with an opinion to have this there.

src/docs/sdk/unified-api/tracing.mdx
```diff
@@ -88,6 +88,37 @@ tree as well as the unit of reporting to Sentry.
...
+```
+  - No value means defer
+
+0 - Don't sample
+
+1 - Sampled
+```
+
+Differently than `b3` a simple sampling decision as a value to `sentry-trace` should not be implemented. [There are reasons to always include](https://github.com/apache/incubator-zipkin-b3-propagation/blob/bc937b6854ea30e46b3e85fbf147d8f4de685dd5/README.md#why-send-trace-ids-with-a-reject-sampling-decision) the `trace-id` and `span-id` regardless of sampling having been decided by the caller. And also this will simplify the implementation.
+Besides the [usual reasons to use *defer](https://github.com/apache/incubator-zipkin-b3-propagation/blob/bc937b6854ea30e46b3e85fbf147d8f4de685dd5/README.md#why-defer-a-sampling-decision),* in the case of Sentry, a reason would be if a downstream system captures an error event with Sentry. The decision could be done at that point to *not* sample that trace in order to have tracing data available for the reported crash.
```
rhcarvalho, line 116: The logic here is reversed -- sample == report data.

```suggestion
Besides the [usual reasons to use *defer](https://github.com/apache/incubator-zipkin-b3-propagation/blob/bc937b6854ea30e46b3e85fbf147d8f4de685dd5/README.md#why-defer-a-sampling-decision),* in the case of Sentry, a reason would be if a downstream system captures an error event with Sentry. The decision could be done at that point to sample that trace in order to have tracing data available for the reported crash. 
```

rhcarvalho (approved):
Thanks for filling in the gaps :)

PR #148, opened 2020-09-22
Add: note about stack frame order

MimiDumpling:
Customers are writing in about the stack frame order and an added note might help with that communication. There's another sentence about stack framer order in the introductory paragraph, but perhaps that gets overlooked.

### DIFF
--- a/src/docs/sdk/event-payloads/stacktrace.mdx
+++ b/src/docs/sdk/event-payloads/stacktrace.mdx
@@ -24,6 +24,11 @@ follow this rule of thumb:
 : **Required**. A non-empty list of stack frames (see below). The list is
   ordered from caller to callee, or oldest to youngest. The last frame is the
   one creating the exception.
+  <Note>
+  <markdown>
+  Note: The stack frame order lists the oldest frame to the most recent (where the exception happened).
+  </markdown>
+  </Note>
 
 `registers`
 
--- a/src/docs/sdk/event-payloads/stacktrace.mdx
+++ b/src/docs/sdk/event-payloads/stacktrace.mdx
@@ -24,10 +24,11 @@ follow this rule of thumb:
 : **Required**. A non-empty list of stack frames (see below). The list is
   ordered from caller to callee, or oldest to youngest. The last frame is the
   one creating the exception.
+  
   <Note>
-  <markdown>
-  Note: The stack frame order lists the oldest frame to the most recent (where the exception happened).
-  </markdown>
+    <markdown>
+      Note: The stack frame order lists the oldest frame to the most recent (where the exception happened).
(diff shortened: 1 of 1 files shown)


MimiDumpling:
@rhcarvalho I tried applying <Note> for consistency, but it's not rendering as expected. Let me know if there's something I'm missing. If not, we can remove the <Note> tag

src/docs/sdk/event-payloads/stacktrace.mdx
```diff
@@ -24,6 +24,12 @@ follow this rule of thumb:
 : **Required**. A non-empty list of stack frames (see below). The list is
   ordered from caller to callee, or oldest to youngest. The last frame is the
   one creating the exception.
+  
+  <Note>
+    <markdown>
+      Note: The stack frame order lists the oldest frame to the most recent (where the exception happened).
```
PeloWriter, line 30:
```suggestion
      Note: The stack frame order lists the oldest frame to the most recent (where the exception happened).
```
```suggestion
      The stack frame order lists the oldest frame to the most recent (where the exception happened).
```

PeloWriter (commented):
LGTM.

src/docs/sdk/event-payloads/stacktrace.mdx
```diff
@@ -24,6 +24,12 @@ follow this rule of thumb:
 : **Required**. A non-empty list of stack frames (see below). The list is
   ordered from caller to callee, or oldest to youngest. The last frame is the
   one creating the exception.
```
rhcarvalho, line 26: @MimiDumpling the paragraph you're adding is replicating this description here, it doesn't look right to have both.

src/docs/sdk/event-payloads/stacktrace.mdx
```diff
@@ -24,6 +24,12 @@ follow this rule of thumb:
...
+  
+  <Note>
+    <markdown>
```
rhcarvalho, line 29: I think these tags / React components do not exist in the "develop" repository, that's why it doesn't produce the desired effect.

src/docs/sdk/event-payloads/stacktrace.mdx
MimiDumpling: @rhcarvalho Yes, good point. We discussed adding the Note to the #attributes section of this page. I looked for other places for the Note, but there's no good fit. Perhaps we don't need the Note?

src/docs/sdk/event-payloads/stacktrace.mdx
MimiDumpling: Ah, that makes sense. Thanks. 😄

PR #149, opened 2020-09-22
docs: adds on-prem docs for Microsoft Teams integration

scefali:
Trying to attach an image of my change but I think Github is having a stroke 

Microsoft Teams Integration _ Sentry Developer Documentation


MimiDumpling (approved):
LGTM 🌻

PR #156, opened 2020-09-24
feat(test) Improve docs on saving events in tests

markstory:
Setting a timestamp helps prevent clock alignment problems, and add example of storing an event.

cc @Zylphrex


Zylphrex (approved):
Thanks!

PR #161, opened 2020-10-01
fix: Change Message.formatted to optional

philipphofmann:
The formatted attribute of the Message for the EventPayload is not required. The comment states
that if it is missing, Sentry will try to interpolate the message. Furthermore, Sentry accepts an empty
formatted attribute.


jan-auer (changes requested):
This is required. The only reason it is in fact optional is because of legacy payloads we're still receiving from old SDKs. This **must** be sent by every new SDK.

PR #172, opened 2020-10-16
Add documentation for adding LB health checks to the on-premise docs

ThomasWunderlich:
Add documentation around the health check endpoint to the on-premise documentation.

The corresponding code for this documentation is the health middleware  https://github.com/getsentry/sentry/blob/master/src/sentry/middleware/health.py


BYK (approved):
Sorry we missed this @ThomasWunderlich and thanks a lot for the PR!

PR #174, opened 2020-10-19
perf(prefetching): Disable prefetching to preserve bw

BYK:


dcramer (approved):

BYK:
Just verified that this disables the immediate prefetching of all sidebar links.

PR #175, opened 2020-10-20
feat(GitLab): Add on prem user docs

ceorourke:
This PR adds GitLab setup instructions - however it's the same as installing GitLab on SaaS so I just followed the format for creating the app as we have for other local integrations and then link to the setup docs on SaaS. 

Screen Shot 2020-10-19 at 5 28 44 PM


scefali (approved):
LGTM

PR #179, opened 2020-10-21
docs: update github.mdx

sumit4613:
`github-app.id` should be an int value otherwise it raises error.


dashed:
@BYK do you know if this is correct?

BYK (approved):
Seems legit @dashed: https://github.com/getsentry/sentry/blob/814a7ff3636522ed4d64fb44dee7851e5fbf1483/src/sentry/options/defaults.py#L127

Thanks a lot @sumit4613!

PR #185, opened 2020-10-27
feat(envelope): Change user report to user feedback

philipphofmann:
The user facing wording of user report is now user feedback on Sentry. When implementing
user feedback on SDKs you have to impliclictly know that user feedback was once called user
report. Therefore UserReport is now renamed to User Feedback for the envelopes so devs can
find it quickly. Furthermore, information about the JSON of the envelope item is added.

### DIFF
--- a/src/docs/sdk/envelopes.mdx
+++ b/src/docs/sdk/envelopes.mdx
@@ -400,14 +400,37 @@ update to an existing session for Release Health.
 
 *None*
 
-### UserReport
+### User Feedback
 
-Item type `"user_report"`. This Item contains a user report JSON payload.
+User Feedback was called User Report in the beginning. Therefore the envelope item type is `"user_report"`. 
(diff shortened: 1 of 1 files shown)


src/docs/sdk/envelopes.mdx
```diff
@@ -400,14 +400,37 @@ update to an existing session for Release Health.
...
+
+: *String, recommended.* The email of the user.
+
+`name`
+
+: *String, recommended.* The name of the user.
+
+`comments`
+
+: *String, recommended.* Comments of the user about what happened.
```
bruno-garcia, line 428: Isn't this required? It seems that minimum is eventId, commments and (name or email)

philipphofmann: No, `eventId` plus any of the other is required. It's fine to send user feedback with only a comment for example.

bruno-garcia (approved):

PR #186, opened 2020-10-27
Fix length header with UTF8 content

Tyrrrz:
"\xef\xbb\xbfHello\r\n\n" has length of 11 as a string (or 10 if you don't count the last \n), but is actually 14 (or 13) bytes in UTF8

### DIFF
--- a/src/docs/sdk/envelopes.mdx
+++ b/src/docs/sdk/envelopes.mdx
@@ -193,7 +193,7 @@ payload which is included in `length`:
 
 ```
 {"event_id":"9ec79c33ec9942ab8353589fcb2e04dc","dsn":"https://e12d836b15bb49d7bbf99e64295d995b:@sentry.io/42"}\n
-{"type":"attachment","length":10,"content_type":"text/plain","filename":"hello.txt"}\n
+{"type":"attachment","length":13,"content_type":"text/plain","filename":"hello.txt"}\n
 \xef\xbb\xbfHello\r\n\n
 {"type":"event","length":41,"content_type":"application/json","filename":"application.log"}\n
 {"message":"hello world","level":"error"}\n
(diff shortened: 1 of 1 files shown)


Tyrrrz:
cc @bruno-garcia

bruno-garcia (approved):
Makes sense to me.

bruno-garcia:
Makes sense to me (it's also what the counting bytes in C# says) but tagging @HazAT and @Swatinem to confirm from their side if there's no misunderstanding here

src/docs/sdk/envelopes.mdx
```diff
@@ -193,7 +193,7 @@ payload which is included in `length`:
 
 ```
 {"event_id":"9ec79c33ec9942ab8353589fcb2e04dc","dsn":"https://e12d836b15bb49d7bbf99e64295d995b:@sentry.io/42"}\n
-{"type":"attachment","length":10,"content_type":"text/plain","filename":"hello.txt"}\n
+{"type":"attachment","length":14,"content_type":"text/plain","filename":"hello.txt"}\n
 \xef\xbb\xbfHello\r\n\n
```
Swatinem, line 197:
```
\xef\xbb\xbfHello\r\n\n
|   |   |   |||||| | <
1   2   3   456789 10<
```

Swatinem (changes requested):
I count 10 bytes. This is not really about utf-8. As the paragraph above says, this is only about the notation.

Swatinem:
Also open the browser console real quick: `"\xef\xbb\xbfHello\r\n".length == 10`

bruno-garcia:
Yeah that makes sense

bruno-garcia:
Thanks @Swatinem

PR #197, opened 2020-11-08
Update overview.mdx

ovcharenko-di:
Removed `Secret Key` from L129, as it was absent in the example dsn above


untitaker:
Thanks!

PR #199, opened 2020-11-09
feat(azure devops): Add on prem user docs

ceorourke:
Add instructions for installing Azure DevOps in an on prem instance of Sentry. Also alphabetize the sidebar and index page. 

Screen Shot 2020-11-09 at 12 45 24 PM

### DIFF
--- a/src/components/sidebar.tsx
+++ b/src/components/sidebar.tsx
@@ -245,12 +245,13 @@ export default () => {
 
         <ul className="list-unstyled" data-sidebar-tree>
           <SidebarLink to="/integrations/">Overview</SidebarLink>
+          <SidebarLink to="/integrations/azuredevops/">Azure DevOps</SidebarLink>
+          <SidebarLink to="/integrations/bitbucket/">Bitbucket</SidebarLink>
           <SidebarLink to="/integrations/github/">GitHub</SidebarLink>
+          <SidebarLink to="/integrations/gitlab/">GitLab</SidebarLink>
+          <SidebarLink to="/integrations/msteams/">Microsoft Teams</SidebarLink>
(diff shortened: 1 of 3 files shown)


src/components/sidebar.tsx
```diff
@@ -245,12 +245,13 @@ export default () => {
...
+          <SidebarLink to="/integrations/azuredevops/">Azure DevOps</SidebarLink>
+          <SidebarLink to="/integrations/bitbucket/">Bitbucket</SidebarLink>
           <SidebarLink to="/integrations/github/">GitHub</SidebarLink>
+          <SidebarLink to="/integrations/gitlab/">GitLab</SidebarLink>
+          <SidebarLink to="/integrations/msteams/">Microsoft Teams</SidebarLink>
           <SidebarLink to="/integrations/slack/">Slack</SidebarLink>
           <SidebarLink to="/integrations/vercel/">Vercel</SidebarLink>
-          <SidebarLink to="/integrations/msteams/">Microsoft Teams</SidebarLink>
-          <SidebarLink to="/integrations/gitlab/">GitLab</SidebarLink>
-          <SidebarLink to="/integrations/bitbucket/">Bitbucket</SidebarLink>
```
manuzope, line 253: Love it!

I don't _think_ you can add a comment in a `return` statement but would be so nice if we could add a comment saying this is alphabetical!

manuzope (approved):

src/components/sidebar.tsx
ceorourke: @manuzope `{/* A JSX comment */}` should work fine I believe

manuzope: Dang, nice. TIL!

manuzope (approved):

PR #201, opened 2020-11-10
fix(github-sso): Fix GitHub SSO app form

BYK:
This was broken in #181


BYK:
I think it is a good idea to setup Sentry for this repo too 😅

untitaker (approved):

PR #207, opened 2020-11-14
feat(integrations): Add develop docs for Jira Integration

NisanthanNanthakumar:
## Objective
Add instructions for how to setup Jira integration when running Sentry locally.

## UI
Screen Shot 2020-11-13 at 4 18 40 PM


src/docs/integrations/jira/index.mdx
```diff
@@ -0,0 +1,18 @@
+---
+title: "Jira Integration"
+---
+
+Create an [Atlassian Cloud Instance](http://go.atlassian.com/cloud-dev)
+
+Configure Jira settings.
+  1. Navigate to Jira settings (cog icon) > Apps > Manage apps.
+  2. Scroll to the bottom of the Manage apps page, and click Settings (1).
+  3. Select Enable development mode and Enable private listings (2), and click Apply.
```
manuzope, line 10: Would be lovely if we could just link to JIRA's docs on how to enable developer mode. Do they have a page like that?

manuzope (approved):

src/docs/integrations/jira/index.mdx
NisanthanNanthakumar: @manuzope they have https://developer.atlassian.com/cloud/jira/platform/getting-started/. That was what I used to write this PR, but we only need steps 1 and 2. Not sure how to specify that.

manuzope: Hmmm, that's fair. What we have right now is probably better. Ship it.

PR #211, opened 2020-11-25
Python3 by default & some clean up

armenzg:


joshuarli:
I'll take a look at this on Monday, wasn't able to get to this today + next 2 days are holiday.

src/docs/environment/index.mdx
```diff
@@ -17,36 +21,82 @@ You're going to be working out of this repository for the remainder of the setup
...
+toolchain up to date by occasionally running:
 
 ```bash
 rustup update stable
 ```
 
 ## Python
 
-We utilize [pyenv](https://github.com/pyenv/pyenv) to install and manage python versions. It should have already been installed earlier when you ran `brew bundle`.
+**NOTE**: We've recently switched our development environment to use Python 3 by default. Temporarily you will still be able
```
EvanPurkhiser, line 78: Don't we have an actual `Note` component or something we can use here @dcramer ?

Or do we hate those callout boxes, I feel like I've heard that somewhere

src/docs/environment/index.mdx
```diff
@@ -17,36 +21,82 @@ You're going to be working out of this repository for the remainder of the setup
...
+the various system packages as listed in Sentry's `Brewfile`.
+
+```shell
+brew bundle --verbose
+```
+
+### Docker
 
-Once you have Xcode installed, install [Homebrew](http://brew.sh), and then run `brew bundle --verbose` to install the various system packages as listed in sentry's `Brewfile`.
+NOTE: There's an issue with the latest Docker, thus, we need to uninstall what `brew bundle` installed and then downgrade with:
```
joshuarli, line 43: I've just verified that https://github.com/getsentry/sentry/pull/22250 works on 2.3.0.5 as well as the most recent 2.5.0.1 - we can remove these instructions.

src/docs/environment/index.mdx
joshuarli: Yes, grep for stuff like this: `<Alert title="Note" level="info">`.

@armenzg could you switch to using those?

src/docs/environment/index.mdx
```diff
@@ -17,36 +21,82 @@ You're going to be working out of this repository for the remainder of the setup
...
 ```
 
 ## Python
 
-We utilize [pyenv](https://github.com/pyenv/pyenv) to install and manage python versions. It should have already been installed earlier when you ran `brew bundle`.
+**NOTE**: We've recently switched our development environment to use Python 3 by default. Temporarily you will still be able
+to set things up with Python 2 by running this command (it will be read by the setup scripts):
+
+```
+export SENTRY_PYTHON2=1
```
joshuarli, line 82: This means this PR shouldn't be merged until https://github.com/getsentry/sentry/pull/22284 lands.

src/docs/environment/index.mdx
```diff
@@ -17,36 +21,82 @@ You're going to be working out of this repository for the remainder of the setup
...
+```
 
-To install the required versions of Python you'll need to run `make setup-pyenv`. This will take a while, since your computer is actually compiling python!
+**NOTE**: If you are switching your development environment from the default Python 2 set up you will need to run `make clean`.
```
joshuarli, line 85: We should just add `clean` to `develop`.

src/docs/environment/index.mdx
```diff
@@ -2,7 +2,11 @@
 title: "Environment"
 ---
 
-This guide steps you through configuring a local development environment for the Sentry server on macOS. If you're using another operating system the instructions are still roughly the same, but we don't maintain any official documentation for anything else for now.
+*NOTE*: There's some documented known issues at the bottom of this page.
```
joshuarli, line 5: Are you referring to the troubleshooting section here? If so, best to just add a link. And move it to the end of the same intro paragraph, the link is enough for a visual distinguishment vs. _NOTE_

joshuarli (changes requested):

src/docs/environment/index.mdx
armenzg: Removed.

src/docs/environment/index.mdx
armenzg: Correct.

src/docs/environment/index.mdx
armenzg: Done.

src/docs/environment/index.mdx
```diff
@@ -17,36 +21,77 @@ You're going to be working out of this repository for the remainder of the setup
...
+toolchain up to date by occasionally running:
 
 ```bash
 rustup update stable
 ```
 
 ## Python
 
-We utilize [pyenv](https://github.com/pyenv/pyenv) to install and manage python versions. It should have already been installed earlier when you ran `brew bundle`.
+<Alert title="Python 2 notice" level="info">
```
joshuarli, line 71:
```suggestion
<Alert title="Python 2 notice" level="warning">
```

Dunno why info is green, but this should be yellow to catch attention.

joshuarli (approved):
lgtm, blocked by https://github.com/getsentry/sentry/pull/22330 and friends

src/docs/environment/index.mdx
```diff
@@ -17,36 +21,77 @@ You're going to be working out of this repository for the remainder of the setup
...
 
-To install the required versions of Python you'll need to run `make setup-pyenv`. This will take a while, since your computer is actually compiling python!
+We've recently switched our development environment to use Python 3 by default. Temporarily you will still be able
```
joshuarli, line 73: Hmm, thinking about this more, actually not sure this belongs at this location in the document, if at all. People who read this document are setting up entirely new devenvs, we shouldn't talk about transitional workarounds.

armenzg: I removed this.

PR #228, opened 2020-12-17
feat(payload): Specify event errors

jan-auer:
Documents the previously internal error list on the event payload, which allows SDKs to attach general errors with handling an event. This list should not be used for annotating individual fields, as this will be done automatically during ingestion, or otherwise should be done via `_meta`.

image

### DIFF
--- a/src/docs/sdk/event-payloads/index.mdx
+++ b/src/docs/sdk/event-payloads/index.mdx
@@ -250,8 +250,8 @@ Fingerprints](https://docs.sentry.io/data-management/event-grouping/).
   declared in the Sentry [EventError] model. If there is no applicable type
   variant, consider [opening an issue] to suggest the addition.
 
-  Apart from types, any attribute is valid. There is a set of conventions for
-  common error attributes:
+  Apart from types, any attribute is valid. There are conventions for the
+  semantics of common error attributes, if they are included:
 
 - `name`: A string declaring the path to the payload field that causes or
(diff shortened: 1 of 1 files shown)


src/docs/sdk/event-payloads/index.mdx
```diff
@@ -236,6 +236,43 @@ import "./properties/tags.mdx"
...
+{
+  "errors": [
+    {
+      "type": "unknown_error",
+      "path": "/var/logs/errors.log.1",
+      "details": "Failed to read attachment"
+    }
+  ]
+}
+```
```
philipphofmann, line 271: `l`: Above we state that `name` and `value` are conventions, but we miss them in here.

philipphofmann (approved):
Really good docs. 🚀

src/docs/sdk/event-payloads/index.mdx
jan-auer: Good point. This was intentional to show that there is no obligation to send them (and it often doesn't make sense), but rather if they are specified, they should have a certain meaning.

jan-auer: @philipphofmann updated the sentence introducing conventions. Does it read better now?

philipphofmann: Yes, is better now. Thanks 👍🏼

PR #238, opened 2021-01-07
fix: Specify valid content types

philipphofmann:
Specify that any valid MIME type can be used as a content type. Brought up in here https://github.com/getsentry/sentry-docs/pull/2826#issuecomment-755500370.


jan-auer (approved):
👌

Issue #221, opened 2020-12-10
Improve transaction & span docs


Tyrrrz:
As discussed on the call with @bruno-garcia and @maciejwalkowiak, the docs on transaction & span (i.e. the performance feature) are lacking. Currently, it's really hard to implement this feature correctly as it requires a lot of guessing, reverse-engineering and asking people questions, all of which should ideally be answered by the documentation.

Issues:

- [ ] Transaction docs doesn't contain an example of a full valid transaction payload. It has an example showing `contexts.trace` and `spans` but not the rest of the fields. Ideally, there should be an example of a minimum valid transaction payload (with only required fields set) and a maximum valid transaction payload (with all of the fields set).
- [ ] Transaction/span schemas are missing from https://github.com/getsentry/sentry-data-schemas/blob/main/relay. Adding them would be very helpful. (Issue raised here https://github.com/getsentry/sentry-data-schemas/issues/2 )
- [ ] The docs on transaction mention that "Transactions are Events enriched with Span data". Does it imply that *all* event fields are valid and make sense on transaction? For example, would having exception data make sense on transaction? Currently, it's not clear what is the actual contract of the transaction payload.
- [ ] The docs should mention that relay may return 200 OK on invalid transactions (not sure if by design or a bug). For example, a transaction that doesn't have `contexts.trace` will get accepted by the relay but won't show up on the portal. Generally speaking, a list of common issues and solutions would be nice.
- [ ] The docs say that "transactions are events with span data", but only briefly mention that the span data is actually stored inside `contexts.trace`. I think this part should be emphasized because it makes no sense from the consumer perspective, but at the same time has critical implications on the design of the SDK.
- [ ] The docs say that the span data is included within `contexts.trace`, but it's not completely true. Some fields like start/end timestamps and description are included on the object itself, despite also being part of "span data". There is an example of `contexts.trace` in the docs, but again, it's not clear whether it's exhaustive or not. What about `tags` and `data`, are they also on `contexts`?
- [ ] The docs on spans don't cover the `span_id`/`parent_span_id`/`sampled` fields and potentially some others. What is the list of fields required/allowed on spans?
[...]

PR #240, opened 2021-01-13
SDK documentation for Performance Monitoring

bruno-garcia:
In this PR we want to address some of the points of this issue https://github.com/getsentry/develop/issues/221

Goals:

* Better general guidelines for SDK developers to implement performance monitoring support
* FAQ section at the end with common questions that come up
* Links to source code of different SDKs to demonstrate how it was done so far
* Clearly communicate some differences (EventProcessor is called in JS but not in others SDKs necessarily).

### DIFF
--- a/src/docs/sdk/performance.mdx
+++ b/src/docs/sdk/performance.mdx
@@ -1,8 +1,8 @@
 ---
-title: "Guidelines for Tracing Support"
+title: "Guidelines for Performance Monitoring"
 ---
 
-This document covers how SDKs should add support for [Distributed
+This document covers how SDKs should add support for Performance Monitoring  with [Distributed
 Tracing](https://docs.sentry.io/product/performance/distributed-tracing/).
 
 This should give an overview of the APIs that SDKs need to implement, without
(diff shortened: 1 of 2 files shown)


src/docs/sdk/performance.mdx
```diff
@@ -1,8 +1,8 @@
 ---
-title: "Guidelines for Tracing Support"
+title: "Guidelines for Performance Monitoring"
```
rhcarvalho, line 2: Idea: if one of the problems is this page is hard to find, try adding `tags: [...]` here in the frontmatter, I heard it is picked up by Algolia search.

rhcarvalho (approved):
The redirects don't seem to be working in the preview deployment, but otherwise nothing is broken so feel free to merge and iterate on it later.

@bruno-garcia we could update https://github.com/getsentry/develop/blob/master/src/docs/sdk/index.mdx to make the Perf Mon page part of the "home page" for SDK docs, makes it easier to navigate to it.

src/docs/sdk/performance.mdx
bruno-garcia: OK merged, but lets follow up on this an iterate

Issue #223, opened 2020-12-10
Document Attachments API implementation


bruno-garcia:
A reference issue is on the [sentry-cocoa repository](https://github.com/getsentry/sentry-cocoa/issues/871) which describes thing in detail.

It also links related PRs:

getsentry/sentry-python#856
getsentry/sentry-java#1082

One more related PR: https://github.com/getsentry/sentry-dotnet/pull/640

PR #241, opened 2021-01-14
feat: Document attachments API

philipphofmann:
Add documentation to expected features for the implementation of the attachments API.

Fixes GH-223

### DIFF
--- a/src/docs/sdk/features.mdx
+++ b/src/docs/sdk/features.mdx
@@ -105,6 +105,23 @@ Envelope item:
  * [Java](https://github.com/getsentry/sentry-java/blob/b2e8dd43ea2ef37201b215ec409655e49e3231c8/sentry/src/main/java/io/sentry/SentryEnvelopeItem.java#L131-L155)
  * [C#](https://github.com/getsentry/sentry-dotnet/blob/3bc6c66a14e653982a6a25df033d68d8ad12c4af/src/Sentry/Protocol/Envelopes/EnvelopeItem.cs#L170-L178)
  
+## Attachments
+
+Attachments are files stored alongside an event or transaction. To send an attachment, add it as an [envelope item](/sdk/envelopes/#attachment)
+to the corresponding event or transaction.
+
+We recommend implementing two types of attachments, one with a path and another with a byte array.
+If the programming language allows it, create one class with multiple constructors to keep things simple and guess the content type
+of the attachment via the filename. Use the implementations of [Java](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Attachment.java),
+[Objective-C](https://github.com/getsentry/sentry-cocoa/blob/master/Sources/Sentry/SentryAttachment.m), or
+[Python](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/attachments.py) as a reference for the API.
+
+
+### Max Attachment Size
+Alongside the implementation of attachments, add `maxAttachmentSize` to the options and set the default to 20 MiB. When converting an attachment to an
+envelope item, the SDK must discard items larger than the `maxAttachmentSize`. Especially on SDKs with offline caching, typical on mobile, this is
+useful because attachments could quickly eat up the users' disk space. Furthermore, Relay has a
+[maximum size for attachments](https://docs.sentry.io/product/relay/options/), and we want to reduce unnecessary requests.
 
 ## Before-Send Hook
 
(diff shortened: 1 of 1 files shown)


philipphofmann:
I added you @mitsuhiko because you implemented the API in Python, so you might have something to add.

src/docs/sdk/features.mdx
```diff
@@ -105,6 +105,23 @@ Envelope item:
  * [Java](https://github.com/getsentry/sentry-java/blob/b2e8dd43ea2ef37201b215ec409655e49e3231c8/sentry/src/main/java/io/sentry/SentryEnvelopeItem.java#L131-L155)
  * [C#](https://github.com/getsentry/sentry-dotnet/blob/3bc6c66a14e653982a6a25df033d68d8ad12c4af/src/Sentry/Protocol/Envelopes/EnvelopeItem.cs#L170-L178)
  
+## Attachments
+
+Attachments are files stored alongside an event or transaction. To send an attachment, add it as an [envelope item](/sdk/envelopes/#attachment)
```
bruno-garcia, line 110: Can you link an attachment to a transaction?

src/docs/sdk/features.mdx
```diff
@@ -105,6 +105,23 @@ Envelope item:
...
+to the corresponding event or transaction.
+
+We recommend implementing two types of attachments, one with a path and another with a byte array.
+If the programming language allows it, create one class with multiple constructors to keep things simple and guess the content type
+of the attachment via the filename. Use the implementations of [Java](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Attachment.java),
+[Objective-C](https://github.com/getsentry/sentry-cocoa/blob/master/Sources/Sentry/SentryAttachment.m), or
+[Python](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/attachments.py) as a reference for the API.
+
```
bruno-garcia, line 118:
```suggestion

The overload that takes a `path` should consider:
* Read the file at the time an event gets captured, and not at the time the attachment is added to the scope.
* If reading the attachment fails, fail gracefully (don't drop the whole envelope)
* If the SDK is in debug mode (`debug=true`) log out errors to help debugging
```

src/docs/sdk/features.mdx
rhcarvalho: > event or transaction

This reads weird as atm transactions are also events.

philipphofmann: @bruno-garcia, yes, you can add attachments to transactions. We add them in [Java](https://github.com/getsentry/sentry-java/blob/931464f007ed852548c3b671311dc98d113df960/sentry/src/main/java/io/sentry/SentryClient.java#L359-L360).
In Python, we also add them to transactions if I recall correctly. @mitsuhiko added [`add_to_transaction`](https://github.com/getsentry/sentry-python/blob/de54b4f99bf9bf746d75f48f2a63a27a2cd6eec2/sentry_sdk/attachments.py#L18) in Python, but I don't think we are checking that somewhere: https://github.com/getsentry/sentry-python/search?q=add_to_transactions.

Maybe we should also add `add_to_transaction` to Java, as reading a large attachment for transactions might cause a huge overhead. If so, we should document that here too.

src/docs/sdk/features.mdx
philipphofmann: Thanks for the input. I slightly modified it.

src/docs/sdk/features.mdx
bruno-garcia: We talked on a call and agreed on so just writing down here for the record:
None of us (mobile team) understand in what case it would make sense to add an attachment to a transaction. Better keep it out until we understand the trade offs.

dcramer: video replays

bruno-garcia: Thanks for clarifying @dcramer

philipphofmann: @bruno-garcia, shouldn't we rather add `add_to_transaction` and set the default to `false`?

bruno-garcia: @philipphofmann Sounds good.

philipphofmann: @bruno-garcia I added two issues to add this flag to [Java](https://github.com/getsentry/sentry-java/issues/1185) and [.NET](https://github.com/getsentry/sentry-dotnet/issues/750).

bruno-garcia (approved):

Issue #252, opened 2021-01-21
SDK Performance/Tracing docs are broken


bruno-garcia:
This change: https://github.com/getsentry/develop/pull/240 seem to have broken things.

It moved it to the sdk folder (note performance.mdx there, but not item in the menu showed up)

![image](https://user-images.githubusercontent.com/1633368/105401418-3092b180-5bf4-11eb-84ce-70751ecc8c84.png)

I added a redirect to the old URL: https://develop.sentry.dev/sdk/unified-api/tracing which returns 404 but renders the content at the bottom of the page

PR #255, opened 2021-01-26
fix(SDKs): Update performance link in sidebar

lobsterkatie:
Sidebar links are hardcoded, so when a page is moved (as happened in https://github.com/getsentry/develop/pull/240), the sidebar has to be updated by hand.

Fixes https://github.com/getsentry/develop/issues/252.


bruno-garcia (approved):
TIL

PR #256, opened 2021-01-26
Terminology change for where to place the tag

armenzg:
Since Github does not call the "first comment" a "description".

### DIFF
--- a/src/docs/python-dependencies.mdx
+++ b/src/docs/python-dependencies.mdx
@@ -21,10 +21,10 @@ So here are the rules we worked with so far:
 In order to avoid ever breaking master, you can:
 
 1. Open PR in getsentry to update requirements
-2. Open PR in sentry with the same branch name, put [`#sync-getsentry`](/workflow/) in PR description
+2. Open PR in sentry with the same branch name, put [`#sync-getsentry`](/workflow/) in the PR's first comment
 3. Merge both when both PRs are green
 
-Unfortunately you'll have to revert the SHA changes made by the sync bot before merging, so this process is elaborate and time-consuming, but by far the safest that never breaks master.
+Unfortunately, you'll have to revert the SHA changes made by the sync bot before merging, so this process is elaborate and time-consuming, but by far the safest that never breaks master.
 
 ## Rules on Licenses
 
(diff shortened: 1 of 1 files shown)


src/docs/python-dependencies.mdx
```diff
@@ -21,10 +21,10 @@ So here are the rules we worked with so far:
 In order to avoid ever breaking master, you can:
 
 1. Open PR in getsentry to update requirements
-2. Open PR in sentry with the same branch name, put [`#sync-getsentry`](/workflow/) in PR description
+2. Open PR in sentry with the same branch name, put [`#sync-getsentry`](/workflow/) in the PR's first comment
```
tonyo, line 24: I'm not sure about this. This still leaves it a bit ambiguous, I would read this as "add another comment after creating the PR". 
What about clarifying what description means ("not the title, but the text you add when creating the PR"?) And if we want to be extra clear, we could also add a hint that the tag can be added later, but for the change to be picked up new changes have to be pushed to the branch.

Thinking about this now, someone could also update the bot to support GitHub labels. If the 'sync-getsentry' label is there => bot syncs. This will hopefully remove the description vs comment vs title ambiguity for good.

armenzg:
I wouldn't worry too much about the label.
We can keep "description". I added an extra comment that hopefully will make it more obvious for the next person coming along.

src/docs/python-dependencies.mdx
```diff
@@ -21,10 +21,11 @@ So here are the rules we worked with so far:
 In order to avoid ever breaking master, you can:
 
 1. Open PR in getsentry to update requirements
-2. Open PR in sentry with the same branch name, put [`#sync-getsentry`](/workflow/) in PR description
+2. Open PR in sentry with the same branch name, put [`#sync-getsentry`](/workflow/) in the PR's description
+   - You can also add it in any other comment, however, you will need to push new changes to trigger the bot
```
tonyo, line 25: @armenzg Double-checked, the bot looks at the PR body only: https://github.com/getsentry/sentry-deploy-sync-hook/blob/master/deployhook.py#L170
So the tag has to go to the PR description (body) specifically, sorry for the confusion.

armenzg:
How about now? FYI you can make suggestions directly in the PR wrt to the wording.

tonyo (approved):

PR #258, opened 2021-01-29
feat(self-hosted): Add docs around backup/restore

BYK:
Fixes getsentry/onpremise#506, fixes getsentry/onpremise#364.

Related: https://app.asana.com/0/1169344595888357/1190341230301185

### DIFF
--- a/src/docs/self-hosted/backup.mdx
+++ b/src/docs/self-hosted/backup.mdx
@@ -1,0 +1,41 @@
+---
+title: "Self-Hosted Backup & Restore"
+---
+
+## Backup via export
+
+If you need a quick way to backup and restore your Sentry instance and you don't need historical event data, you can use the built in `export` and `import` commands. These commands will save and load all project and user data, but will not contain any event data.
+
+```shell
+docker-compose run --rm -T -e SENTRY_LOG_LEVEL=CRITICAL web export > backup.json
+```
+
+If you omit the `-T` or `-e SENTRY_LOG_LEVEL=CRITICAL` parts, your backup file will have logging lines mixed in and you'll have to remove them somehow.
+
+## Restore via import
+
+Once you have a backup via the `export` command, the easiest way to restore it is to placing it under the `sentry` directory in your main `onpremise` repo, next to the config files. This directory automatically gets mounted to `/etc/sentry` so you can run the following to restore your data:
+
+```shell
+docker-compose run --rm -T web import /etc/sentry/backup.json
+```
+
+If you don't see any errors and the process exits with code `0`, congratulations, you have just restored your data.
+
+It is strongly recommended to restore the data on the same version of Sentry on a fresh install (empty database but migrations are run), as otherwise you are very likely to hit errors and even corrupt your database.
+
+## Total backup/restore via Docker volumes
+
+The ideal way to backup and restore Sentry is to backup and restore all Docker volumes used by it. All volumes that hold critical long-term data are already defined as global volumes at install time and are prefixed with `sentry-`:
+
+- `sentry-data`
+- `sentry-postgres`
+- `sentry-redis`
+- `sentry-zookeeper`
+- `sentry-kafka`
+- `sentry-clickhouse`
+- `sentry-symbolicator`
+
+Only backing up and restoring these volumes should bring back all persisted data. If you also need to back up in-flight data, we recommend backing up any project-specific volumes that `docker-compose` automatically creates, typically with the `sentry_onpremise_sentry-` prefix.
+
+Docker documents [how to backup and restore volumes](https://docs.docker.com/storage/volumes/#backup-restore-or-migrate-data-volumes) on their documentation. You may use different methods too as long as the volumes can be read back without issues.
--- a/src/docs/self-hosted/backup.mdx
+++ b/src/docs/self-hosted/backup.mdx
@@ -2,29 +2,35 @@
 title: "Self-Hosted Backup & Restore"
 ---
 
-## Backup With `export`
+## Quick Backup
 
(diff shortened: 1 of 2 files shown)


BYK:
Ping @PeloWriter

src/docs/self-hosted/backup.mdx
```diff
@@ -0,0 +1,41 @@
+---
+title: "Self-Hosted Backup & Restore"
+---
+
+## Backup via export
```
PeloWriter, line 5:
```suggestion
## Backup With `export`
```

src/docs/self-hosted/backup.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+If you need a quick way to backup and restore your Sentry instance and you don't need historical event data, you can use the built in `export` and `import` commands. These commands will save and load all project and user data, but will not contain any event data.
+
+```shell
+docker-compose run --rm -T -e SENTRY_LOG_LEVEL=CRITICAL web export > backup.json
+```
+
+If you omit the `-T` or `-e SENTRY_LOG_LEVEL=CRITICAL` parts, your backup file will have logging lines mixed in and you'll have to remove them somehow.
+
+## Restore via import
```
PeloWriter, line 15:
```suggestion
## Restore With `import`
```

src/docs/self-hosted/backup.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+Once you have a backup via the `export` command, the easiest way to restore it is to placing it under the `sentry` directory in your main `onpremise` repo, next to the config files. This directory automatically gets mounted to `/etc/sentry` so you can run the following to restore your data:
```
PeloWriter, line 17:
```suggestion
Once you have a backup by using the `export` command, the easiest way to restore it is to place it under the `sentry` directory in your main `onpremise` repo, next to the config files. This directory automatically gets mounted to `/etc/sentry` so you can run the following to restore your data:
```

src/docs/self-hosted/backup.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+```shell
+docker-compose run --rm -T web import /etc/sentry/backup.json
+```
+
+If you don't see any errors and the process exits with code `0`, congratulations, you have just restored your data.
+
+It is strongly recommended to restore the data on the same version of Sentry on a fresh install (empty database but migrations are run), as otherwise you are very likely to hit errors and even corrupt your database.
```
PeloWriter, line 25:
```suggestion
We strongly recommended that you restore data on the **same version of Sentry** on a fresh install (empty database but migrations are run). Otherwise, you are very likely to hit errors and may corrupt your database.
```

src/docs/self-hosted/backup.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+## Total backup/restore via Docker volumes
```
PeloWriter, line 27:
```suggestion
## Total Backup and Restore With Docker Volumes
```

src/docs/self-hosted/backup.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+The ideal way to backup and restore Sentry is to backup and restore all Docker volumes used by it. All volumes that hold critical long-term data are already defined as global volumes at install time and are prefixed with `sentry-`:
```
PeloWriter, line 29:
```suggestion
The ideal way to backup and restore Sentry is to backup and restore all Docker volumes that it uses. All volumes that hold critical long-term data are already defined as global volumes at install time and are prefixed with `sentry-`:
```

PeloWriter (approved):
LGTM

BYK:
Thanks a lot @PeloWriter! Will still wait for @chadwhitacre's input before merging.

chadwhitacre:
Great doc to have! Some thoughts:

1. The three h2 headings are not parallel. I think the basic structure we're aiming for is between a **partial** backup/restore and a **full** backup/restore. I suggest reflecting that in the heading structure. 
2. I think "[full](https://www.google.com/search?q=full+backup)" backup is a better-established term than "[total](https://www.google.com/search?q=total+backup)."
3. I think it'd be good to use consistent, unambiguous terms for the two types of data (e.g., "you can run the following to restore your data" is confusing—not obvious without mental effort what "data" refers to there). Maybe **configuration** and **data**? Maybe **metadata** and **events**?
4. Does the "same version of Sentry" recommendation apply to both types of backup/restore (i.e., total/full and partial)? If so consider moving it to the beginning or end of the document.
5. Also, consider putting the "same version of Sentry" recommendation in a warning/note box.
6. Which is the more normal case? Total/full or partial? Whichever is more normal should be first, imo.

chadwhitacre (changes requested):
Great doc to have! Some thoughts:

1. The three h2 headings are not parallel. I think the basic structure we're aiming for is between a **partial** backup/restore and a **full** backup/restore. I suggest reflecting that in the heading structure. 
2. I think "[full](https://www.google.com/search?q=full+backup)" backup is a better-established term than "[total](https://www.google.com/search?q=total+backup)."
3. I think it'd be good to use consistent, unambiguous terms for the two types of data (e.g., "you can run the following to restore your data" is confusing—not obvious without mental effort what "data" refers to there). Maybe **configuration** and **data**? Maybe **metadata** and **events**?
4. Does the "same version of Sentry" recommendation apply to both types of backup/restore (i.e., total/full and partial)? If so consider moving it to the beginning or end of the document.
5. Also, consider putting the "same version of Sentry" recommendation in a warning/note box.
6. Which is the more normal case? Total/full or partial? Whichever is more normal should be first, imo.

chadwhitacre (approved):

PR #259, opened 2021-02-01
Fix: context name not required when raw_description is given

rhcarvalho:
Fixes https://github.com/getsentry/sentry-docs/issues/1590.

### DIFF
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -25,7 +25,7 @@ For more details about sending additional data with your event, see the [full do
 ## Device Context
 
 Device context describes the device that caused the event. This is most
-appropriate for mobile applications. 
+appropriate for mobile applications.
 
 The `type` and default key is `"device"`.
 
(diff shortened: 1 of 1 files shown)


src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -147,7 +147,7 @@ The `type` and default key is `"os"`.
 
 `name`
 
-: **Required**. The name of the operating system.
+: _Optional_. The name of the operating system. It might be derived from `raw_description`. It is **required** if `raw_description` is not provided.
```
jan-auer, line 150: These should at least be recommended. Even with `raw_description`, SDKs are encouraged to try obtain better values.

```suggestion
: _Recommended_. The name of the operating system. It might be derived from `raw_description`. It is **required** if `raw_description` is not provided.
```

jan-auer (approved):

src/docs/sdk/event-payloads/contexts.mdx
rhcarvalho: That's fair, updated both.

jan-auer (approved):

PR #260, opened 2021-02-01
fix(Note): Expose `--light-text` CSS var, fixups

BYK:
`Note` component was using `--light-text` CSS var, which was not exposed here. This made `Note`s look identical with regular content. This patch exposes the variable, fixing the issue.

It also converts some notes to `Alert`s to get more attention and revises some grammar in them.

Old:

image

New:

image


BYK:
Ping @PeloWriter

dcramer:
Note is intentionally different than Alert, as its supposed to be less "in your face". Not sure what isn't working, but it's something thats needed in docs.

BYK:
>Note is intentionally different than Alert, as its supposed to be less "in your face". Not sure what isn't working, but it's something thats needed in docs.

I mean, it was just looking like regular text and it was only used in 2 pages. Doesn't `<Alert level="info">` cover the same "less in your face" scenario while being more consistent?

BYK:
@dcramer updated the PR summary with screenshots. I can remove the "Note" titles.

dcramer:
Alert is a big blue box.

Note is supposed to be a diff font color. Maybe we're must missing the styles for some reason

dcramer:
image

gray text here is note

PeloWriter:
Note in our user docs is used extensively. Like so:
Screen Shot 2021-02-01 at 9 58 55 AM

BYK:
>Note is supposed to be a diff font color. Maybe we're must missing the styles for some reason

Possibly, I'll look into that.

>gray text here is note

I don't find that very readable or eye-catching (maybe because I'm color-blind?) but sure I'll restore `<Note>` and try to get its styles in as replacing `<Note>` with `<Alert level="info">` seems like a larger discussion.

BYK:
@dcramer @PeloWriter fixed, ready for another review.

src/docs/config.mdx
```diff
@@ -113,9 +110,9 @@ LOGGING['default_level'] = 'WARNING'
 
 If logging in a particular module is not showing up when you expect it to, you should check the log level for that module in `src/sentry/conf/server.py` in the `LOGGING` variable.
 
-<Note>
-If you are a Sentry employee who is looking for logs in Kibana and they aren't showing up when you expect them to, one potential reason is that a field in the log message does not match Elasticsearch's schema. See the index patterns <a href="https://kibana.getsentry.net/app/kibana#/management/kibana/indices/0ba261d0-97ca-11e9-90db-dfacda4da93b?_g=()&_a=(tab:indexedFields)">here</a>.
-</Note>
+<Note><markdown>
```
PeloWriter, line 113: I haven't found it necessary to use the <markdown></markdown> elements in user docs, FYI, but I don't think they'll cause issues.

src/docs/config.mdx
```diff
@@ -113,9 +110,9 @@ LOGGING['default_level'] = 'WARNING'
...
+If you are a Sentry employee who is looking for logs in Kibana and they aren't showing up when you expect them to, one potential reason is that a field in the log message not matching Elasticsearch's schema. Check <a href="https://kibana.getsentry.net/app/kibana#/management/kibana/indices/0ba261d0-97ca-11e9-90db-dfacda4da93b?_g=()&_a=(tab:indexedFields)">the index patterns</a> to see whether this is the reason.
```
PeloWriter, line 114:
```suggestion
If you are a Sentry employee looking for logs in Kibana and they aren't showing up when as expected, one potential reason is that a field in the log message does not match Elasticsearch's schema. Check <a href="https://kibana.getsentry.net/app/kibana#/management/kibana/indices/0ba261d0-97ca-11e9-90db-dfacda4da93b?_g=()&_a=(tab:indexedFields)">the index patterns</a> to see whether this is the reason.
```

src/docs/config.mdx
```diff
@@ -269,9 +266,9 @@ The secret used to verify GitHub App Webhook event.
 
 Your app's private key.
 
-<Note>
-Make sure to replace newlines them with `\n` to preserve them.
-</Note>
+<Alert level="warning">
+Make sure to replace the newlines with `\n` to preserve them.
```
PeloWriter, line 270:
```suggestion
Replace the newlines with `\n` to preserve them.
```

PeloWriter (approved):
You may not need the " markdown  /markdown" elements

src/docs/config.mdx
BYK: Eh, will leave them for now. May do another sweep to fix them.

PR #266, opened 2021-02-10
Warn folks about the true meaning of CalVer

chadwhitacre:
Arrived here from self-hosted TSC conversation today. Small mitigation ahead of https://github.com/getsentry/onpremise/issues/854.

### Screenshot

Screen Shot 2021-02-10 at 4 25 23 PM

### DIFF
--- a/src/docs/self-hosted/releases.mdx
+++ b/src/docs/self-hosted/releases.mdx
@@ -4,7 +4,7 @@ title: 'Self-Hosted Versions & Releases'
 
 ![CalVer: YY.MM.MICRO](https://img.shields.io/badge/calver-YY.MM.MICRO-ad6caa.svg)
 
-Self-hosted Sentry follows a monthly release schedule using the [CalVer](https://calver.org/#scheme) versioning scheme. A new version is released on the [15th of each month](https://github.com/getsentry/onpremise/blob/704e4c3b5b7360080f79bcfbe26583e5a95ae675/.github/workflows/release.yml#L20-L24), and occasionally in between, as well. You can find the [latest release](https://github.com/getsentry/onpremise/releases/latest) over at the [releases section of our onpremise repository](https://github.com/getsentry/onpremise/releases/).
+Self-hosted Sentry follows a monthly release schedule using the [CalVer](https://calver.org/#scheme) versioning scheme. A new version is released on the [15th of each month](https://github.com/getsentry/onpremise/blob/704e4c3b5b7360080f79bcfbe26583e5a95ae675/.github/workflows/release.yml#L20-L24), and follow-up releases are done when necessary in between. You can find the [latest release](https://github.com/getsentry/onpremise/releases/latest) over at the [releases section of our onpremise repository](https://github.com/getsentry/onpremise/releases/).
 
 <Alert title="Why CalVer?" level="info">
   In short, this is to keep the self-hosted Sentry as close to the live version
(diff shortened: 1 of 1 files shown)


src/docs/self-hosted/releases.mdx
```diff
@@ -4,14 +4,18 @@ title: 'Self-Hosted Versions & Releases'
 
 ![CalVer: YY.MM.MICRO](https://img.shields.io/badge/calver-YY.MM.MICRO-ad6caa.svg)
 
-Self-hosted Sentry follows a monthly release schedule using the [CalVer](https://calver.org/#scheme) versioning scheme. A new version is released on the [15th of each month](https://github.com/getsentry/onpremise/blob/704e4c3b5b7360080f79bcfbe26583e5a95ae675/.github/workflows/release.yml#L20-L24), and patch releases are done when needed in-between. You can find the [latest release](https://github.com/getsentry/onpremise/releases/latest) over at the [releases section of our onpremise repository](https://github.com/getsentry/onpremise/releases/).
```
chadwhitacre, line 7: Removing "patch" here since that implies "stability" more strongly than we'd like.

BYK: How about "follow-up releases"?

src/docs/self-hosted/releases.mdx
```diff
@@ -4,14 +4,18 @@ title: 'Self-Hosted Versions & Releases'
...
+Self-hosted Sentry follows a monthly release schedule using the [CalVer](https://calver.org/#scheme) versioning scheme. A new version is released on the [15th of each month](https://github.com/getsentry/onpremise/blob/704e4c3b5b7360080f79bcfbe26583e5a95ae675/.github/workflows/release.yml#L20-L24), and occasionally in between, as well. You can find the [latest release](https://github.com/getsentry/onpremise/releases/latest) over at the [releases section of our onpremise repository](https://github.com/getsentry/onpremise/releases/).
 
 <Alert title="Why CalVer?" level="info">
   In short, this is to keep the self-hosted Sentry as close to the live version
   hosted at <Link to="https://sentry.io">sentry.io</Link>. There are more details for the
   curious over at our <Link to="https://blog.sentry.io/2020/06/22/self-hosted-sentry-switching-to-calver">blog post announcing the switch</Link>.
 </Alert>
 
+<Alert title="Warning" level="warning">
```
BYK, line 15: I think a warning is too much here. Info should suffice?
```suggestion
<Alert level="info">
```

BYK (changes requested):

src/docs/self-hosted/releases.mdx
chadwhitacre: Mmmm ... gets lost next to other info. I think warning is the right level, this is a fundamentally different model than SemVer and the N.N.N release number and the monthly cadence can mask that. Interested in @jan-auer's input.

src/docs/self-hosted/releases.mdx
chadwhitacre: Wfm, done in b3969a7d63482da5e8aeefedbdaa6495c22f1fb0.

src/docs/self-hosted/releases.mdx
BYK: Eh, fair enough. Will unblock for now. If we get complaints (from users or from @PeloWriter, we can revisit).

chadwhitacre: Also, the link is to https://sentry.io/pricing/, definitely want to "warn" people there. 😁

BYK (approved):

PR #268, opened 2021-02-11
fix: remove stray **s in pycharm docs

alxu-sentry:


untitaker (commented):
feel free to merge in this repo without approval

what you're fixing is likely an artifact of notion export

PR #270, opened 2021-02-12
fix: Use official cask upgrade command

dashed:
Running `brew cask upgrade chromedriver` outputs the following:

```
Error: Calling `brew cask upgrade` is disabled! Use brew upgrade [--cask] instead.
```

Ref: https://stackoverflow.com/questions/31968664/upgrade-all-the-casks-installed-via-homebrew-cask

https://github.com/Homebrew/discussions/discussions/340


armenzg (commented):
lgtm

PR #275, opened 2021-02-16
Clarify effects of empty DSN

rhcarvalho:
Addresses https://github.com/getsentry/sentry-docs/issues/1635

See also https://github.com/getsentry/sentry-docs/pull/3096

### DIFF
--- a/src/docs/sdk/overview.mdx
+++ b/src/docs/sdk/overview.mdx
@@ -39,7 +39,7 @@ Additionally, the following features are highly encouraged:
 
 Generally, using an SDK consists of three steps for the end user, which should look almost identical no matter the language:
 
-1.  Creation of the SDK (sometimes this is hidden from the user):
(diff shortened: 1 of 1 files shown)


src/docs/sdk/overview.mdx
```diff
@@ -78,7 +79,10 @@ Sentry.init({
 ```
 
 <Alert title="Note" level="info">
-The SDK should accept an empty DSN as valid configuration.
+SDKs should accept an empty DSN as valid configuration.
+
+If an SDK is not initialized or if it is initialized with an empty DSN, the SDK should not send any data over the network, such as captured exceptions.
```
bruno-garcia, line 84:
```suggestion
If an SDK is not initialized or if it is initialized with an empty string, the SDK should not send any data over the network, such as captured exceptions.
```

bruno-garcia (approved):

PR #280, opened 2021-03-03
fix(docs): Update "Debugging with PyCharm"

RyanSkonnord:
Devserver's `--skip-daemons` flag has been removed and entirely replaced
by `--debug-server`. Update and simplify all associated instructions.
Delete or trim down tips that are no longer as relevant.


alxu-sentry (approved):
seems consistent with my experience over the past couple days. it'd be good if we could fix SIGINT but I guess it's not a high priority.

untitaker (approved):
would eventually remove SENTRY_USE_UWSGI flag from code as well

PR #287, opened 2021-03-09
Python dependencies hard pinned in Sentry (no more getsentry)

armenzg:
We're moving toward treating `sentry` as a requirement of `getsentry`,
thus, dependencies on `sentry` do not need to be defined on `getsentry` as well.

We also won't need to synchronize PRs on both repositories.

### DIFF
--- a/src/docs/python-dependencies.mdx
+++ b/src/docs/python-dependencies.mdx
@@ -2,34 +2,26 @@
 title: "Python Dependencies"
 ---
 
-Sentry as you might know for the most part runs on an old version of Python (2.7). This version has reached end of life which has some consequences in how we develop and pick dependencies.
+Unlike our frontend JavaScript story, where we're generally very happy pulling in dependencies, we're much more conservative on
+the backend. Any dependency we pull in, might require us to eventually (temporarily) fork and vendor it if the upstream
+project no longer supports our version of Python.
 
-Unlike our frontend JavaScript story where we're generally very happy pulling in dependencies we're much more conservative on the backend. Any dependency we pull in might require us to eventually (temporarily) fork and vendor it if the upstream project no longer supports our version of Python.
-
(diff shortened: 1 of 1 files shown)


armenzg:
@joshuarli I should probably make a comment wrt to exceptions to the norm (e.g. the sentry-sdk URL); or should we assume that those folks that are doing something different know what they're doing?

src/docs/python-dependencies.mdx
```diff
@@ -2,34 +2,26 @@
...
 
-Sentry uses BSD/MIT/ISC and Apache 2 licenses. Whatever we used needs to be compatible with this. This means an absolute hard no on GPL/AGPL and a soft no on LGPL unless absolutely necessary. Acceptable uses of LGPL are swappable components like database drivers.
+Sentry uses BSD/MIT/ISC and Apache 2 licenses. Whatever we used needs to be compatible with this. This means an absolute
+hard no on GPL/AGPL and a soft no on LGPL unless absolutely necessary. Acceptable uses of LGPL are swappable components
+like database drivers.
 
 ## Unclear?
 
-If you have questions about dependencies feel free to reach out to Armin Ronacher or Matt Robenolt with questions.
+If you have questions about dependencies feel free to reach out to Armin Ronacher with questions.
```
joshuarli, line 27:
```suggestion
If you have questions about dependencies feel free to reach out to [owners-python-build](https://github.com/orgs/getsentry/teams/owners-python-build/members) with questions.
```

joshuarli (approved):

joshuarli:
> @joshuarli I should probably make a comment wrt to exceptions to the norm (e.g. the sentry-sdk URL); or should we assume that those folks that are doing something different know what they're doing?

Oh yeah, I'd add a rule saying something like "Sentry dependencies that have a URL need to be specified in getsentry for X reason." (i forget the reason)

armenzg:
I added the note. Thanks Josh!

PR #288, opened 2021-03-11
Consider sampler callback returning null

bruno-garcia:
Is this fair to say?

### DIFF
--- a/src/docs/sdk/performance.mdx
+++ b/src/docs/sdk/performance.mdx
@@ -110,8 +110,9 @@ When there's the potential for more than one of these to come into play, the fol
 
 1. If a sampling decision is passed to `startTransaction` (`startTransaction({name: "my transaction", sampled: true})`), that decision will be used, regardlesss of anything else
 2. If `tracesSampler` is defined, its decision will be used. It can choose to keep or ignore any parent sampling decision, or use the sampling context data to make its own decision or choose a sample rate for the transaction.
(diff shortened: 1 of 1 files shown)


src/docs/sdk/performance.mdx
```diff
@@ -110,8 +110,9 @@ When there's the potential for more than one of these to come into play, the fol
 
 1. If a sampling decision is passed to `startTransaction` (`startTransaction({name: "my transaction", sampled: true})`), that decision will be used, regardlesss of anything else
 2. If `tracesSampler` is defined, its decision will be used. It can choose to keep or ignore any parent sampling decision, or use the sampling context data to make its own decision or choose a sample rate for the transaction.
-3. If `tracesSampler` is not defined, but there's a parent sampling decision, the parent sampling decision will be used.
-4. If `tracesSampler` is not defined and there's no parent sampling decision, `tracesSampleRate` will be used.
+3. If `tracesSampler` returns `null` and `tracesSampleRate` is defined, `tracesSampleRate` will be used.
```
bruno-garcia, line 113:
```suggestion
3. If `tracesSampler` is defined but returns `null`, `tracesSampleRate` will be used.
```

lobsterkatie:
I think this needs a larger conversation, including about how client-side dynamic sampling is going to interact with server-side dynamic sampling. Right now, JS and Python have no concept of "return null" from `traces_sampler`, nor any infrastructure for the decision-deferring case. There's also an open question of whether an unsampled transaction should be sending headers in the first place. Perhaps we can discuss this in a future SDK evolution meeting?

PR #289, opened 2021-03-11
Clarify that 0.0 sampling can sample

bruno-garcia:
I'm sure it can be better described, but I believe currently its a bit misleading

### DIFF
--- a/src/docs/sdk/performance.mdx
+++ b/src/docs/sdk/performance.mdx
@@ -20,7 +20,7 @@ Tracing is enabled by setting either one of two new SDK config options, `tracesS
 
 ### `tracesSampleRate`
 
-This should be a float/double between `0.0` and `1.0` (inclusive) and represents the percentage chance that any given transaction will be sent to Sentry. (`0.0` is a 0% chance, effectively disabling tracing, and `1.0` is a 100% chance, meaning all transactions will be sent.) This rate applies equally to all transactions; in other words, each transaction should have the same random chance of ending up with `sampled = true`, equal to the `tracesSampleRate`.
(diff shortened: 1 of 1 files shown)


src/docs/sdk/performance.mdx
```diff
@@ -20,7 +20,7 @@ Tracing is enabled by setting either one of two new SDK config options, `tracesS
 
 ### `tracesSampleRate`
 
-This should be a float/double between `0.0` and `1.0` (inclusive) and represents the percentage chance that any given transaction will be sent to Sentry. (`0.0` is a 0% chance, effectively disabling tracing, and `1.0` is a 100% chance, meaning all transactions will be sent.) This rate applies equally to all transactions; in other words, each transaction should have the same random chance of ending up with `sampled = true`, equal to the `tracesSampleRate`.
+This should be a float/double between `0.0` and `1.0` (inclusive) and represents the percentage chance that any given transaction will be sent to Sentry. (`0.0` is a 0% chance, effectively disabling tracing unless the callback is also defined, or a parent trace was sampled. And `1.0` is a 100% chance, meaning all transactions will be sent.) This rate applies equally to all transactions; in other words, each transaction should have the same random chance of ending up with `sampled = true`, equal to the `tracesSampleRate`.
```
lobsterkatie, line 23: A callback _shouldn't_ also be defined, though - they're mutually exclusive options. I do think it's worthwhile to be clearer that `0` is not an absolute disabler.

```suggestion
This should be a float/double between `0.0` and `1.0` (inclusive) and represents the percentage chance that any given transaction will be sent to Sentry. So, barring [outside influence](https://develop.sentry.dev/sdk/performance/#sampling), `0.0` is a 0% chance (none will be sent) and `1.0` is a 100% chance (all will be sent). This rate applies equally to all transactions; in other words, each transaction should have the same random chance of ending up with `sampled = true`, equal to the `tracesSampleRate`.
```

bruno-garcia: > A callback shouldn't also be defined, though - they're mutually exclusive options. I do think it's worthwhile to be clearer that 0 is not an absolute disabler.

They are two independent options, nothing stops a user from setting up both.

The choice for us is: Do we ignore the random sampling option if the callback is defined or do we use it as a fallback if the tracing sampler callback can return null (which we implement in some SDKs).

bruno-garcia:
Good edit! thanks @lobsterkatie

lobsterkatie (approved):
LGTM! Thanks for suggesting it. 🙂

PR #292, opened 2021-03-19
Update tracesSampler docs about returning null

maciejwalkowiak:
Java & .NET SDKs already allow returning `null` from `tracesSampler`.

### DIFF
--- a/src/docs/sdk/performance.mdx
+++ b/src/docs/sdk/performance.mdx
@@ -30,7 +30,7 @@ This should be a callback, called when a transaction is started, which will be g
 
 Optionally, the `tracesSampler` callback can also return a boolean to force a sampling decision (with `false` equivalent to `0.0` and `true` equivalent to `1.0`). If returning two different datatypes isn't an option in the implementing language, this possibility can safely be omitted.
 
-Optionally, the `tracesSampler` callback can also return `null` causing SDK to fallback to a sample rate set with `tracesSampleRate` property.
+Optionally, in some SDKs, the `tracesSampler` callback can also return `null` causing it to behave like not having the callback at all. For example using the sample rate set with `tracesSampleRate` property and inheriting the sampling decision of a trace.
 
 See more about how sampling should be performed below.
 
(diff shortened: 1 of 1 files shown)


maciejwalkowiak:
cc @bruno-garcia

src/docs/sdk/performance.mdx
```diff
@@ -30,6 +30,8 @@ This should be a callback, called when a transaction is started, which will be g
 
 Optionally, the `tracesSampler` callback can also return a boolean to force a sampling decision (with `false` equivalent to `0.0` and `true` equivalent to `1.0`). If returning two different datatypes isn't an option in the implementing language, this possibility can safely be omitted.
 
+Optionally, the `tracesSampler` callback can also return `null` causing SDK to fallback to a sample rate set with `tracesSampleRate` property.
```
bruno-garcia, line 33:
```suggestion
Optionally, in some SDKs, the `tracesSampler` callback can also return `null` causing it to behave like not having the callback at all. For example using the sample rate set with `tracesSampleRate` property and inheriting the sampling decision of a trace.
```

bruno-garcia (approved):

marandaneto (approved):

marandaneto:
I don't have vercels permission :(

rhcarvalho:
Are we still pursuing this change even though #288 was dropped?

marandaneto:
> Are we still pursuing this change even though #288 was dropped?

I guess so, the closed PR was changing the spec so all SDKs should follow it, this one is making it a bit more flexible, some SDKs can return null, but it's an optional behavior.

bruno-garcia:
I second @marandaneto here. @rhcarvalho and @lobsterkatie green light from u to add this note?

rhcarvalho:
I understood our agreement was to not differ when not obviously advantageous to do so. The `tracesSampleRate` and `tracesSampler` init options where meant to be mutually exclusive, if you use the latter you don't use the former.

I think this is a too subtle behavior difference that needs to be documented without being able to reuse content, and it confuses support specialists that will need to keep the difference in mind, etc. From our conversations I didn't find the argument for it compelling enough that we'd change all SDKs in that direction either.

We've already updated user docs to include checking the parent decision (as we do in other platforms) -- https://github.com/getsentry/sentry-docs/pull/3231.

marandaneto:
@bruno-garcia should we close this? that's how it works on Java/NET but ideally should not be part of the spec

PR #302, opened 2021-03-31
feat(analytics): Add docs on analytics

NisanthanNanthakumar:
## Objective:
We should document how we can instrument our code with our current analytics infrastructure. 

I added docs on Big Query and Datadog with an implementation example.

## UI:
localhost_8000_analytics_

### DIFF
--- a/src/docs/analytics.mdx
+++ b/src/docs/analytics.mdx
@@ -2,7 +2,7 @@
 title: "Analytics"
 ---
 
-This guide steps you through instrumenting your code with Sentry's analytics infrastructure.
+This guide steps you through instrumenting your code with Sentry's 3rd-party analytics infrastructure.
 
 ## Big Query
 
--- a/src/docs/analytics.mdx
+++ b/src/docs/analytics.mdx
@@ -42,7 +42,7 @@ analytics.record(
 
 ## Datadog
 
-Track aggregrate stats with Datadog. An example be to track aggregate response codes for an endpoint.
+Track aggregrate stats with Datadog. For example, this can be used to track aggregate response codes for an endpoint.
 
 Import the metrics library and use the `metrics.inc` function. The key needs to be unique.
 
--- a/src/docs/analytics.mdx
+++ b/src/docs/analytics.mdx
@@ -1,0 +1,58 @@
+---
+title: "Analytics"
+---
+
+This guide steps you through instrumenting your code with Sentry's analytics infrastructure.
+
+## Big Query
+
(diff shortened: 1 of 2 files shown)


src/docs/analytics.mdx
```diff
@@ -0,0 +1,58 @@
...
+    organization_id=project.organization_id,
+    project_id=project.id,
+    group_id=group.id,
+    # more attributes
+)
+```
+
+## Datadog
+
+Track aggregrate stats with Datadog. An example be to track aggregate response codes for an endpoint.
```
ceorourke, line 45: > An example be to track aggregate response codes for an endpoint.

maybe "For example, this can be used to track aggregate response codes for an endpoint.

src/docs/analytics.mdx
```diff
@@ -0,0 +1,58 @@
+---
+title: "Analytics"
+---
+
+This guide steps you through instrumenting your code with Sentry's analytics infrastructure.
```
ceorourke, line 5: > This guide steps you through instrumenting your code with Sentry's analytics infrastructure.

maybe "This guide steps you through instrumenting your code with Sentry's preferred analytics infrastructure." since we didn't create Redash and/or Datadog? or "...Sentry's choice of ... " idk something like that.

NisanthanNanthakumar: @ceorourke or maybe `This guide steps you through instrumenting your code with Sentry's 3rd-party analytics infrastructure.`

src/docs/analytics.mdx
```diff
@@ -0,0 +1,58 @@
...
+analytics.record(
+    "codeowners.assignment", # replace with your key
+    organization_id=project.organization_id,
+    project_id=project.id,
+    group_id=group.id,
+    # more attributes
+)
+```
+
+## Datadog
```
manuzope, line 43: I don't think this necessarily needs to be Datadog: https://develop.sentry.dev/services/metrics/.

We support a bunch of metrics engines.

ceorourke (approved):

src/docs/analytics.mdx
NisanthanNanthakumar: @manuzope hmm I didn't know that. I still think its relevant if you want an example of sending events to Datadog. But I think we should link to that page, with some context that metrics.incr can be used with any backend.

manuzope: Yea, I think the section is used. I would just make the header "StatsD" or "Metrics". And avoid using "Datadog".

manuzope: I'd go with metrics.

PR #323, opened 2021-04-30
Guidelines for sessions in SDKs

rhcarvalho:
https://github.com/getsentry/develop/pull/323#issuecomment-842332513

### DIFF
--- a/src/docs/sdk/sessions.mdx
+++ b/src/docs/sdk/sessions.mdx
@@ -434,25 +434,24 @@ The most basic API exposed is on the hub level and lets you start and stop sessi
 We track the health of each release of projects in Sentry by sending session payloads from SDKs.
 The session payload provide data such as session duration and the presence or absence of errors/crashes.
 
-SDKs may track sessions in one of two modes:
-- Individual Sessions
+SDKs track sessions in one of two modes:
+- Single Session
 - Session Aggregates
 
-Individual sessions are the general case, and are a good fit for (relatively short-lived) applications that are typically involving only a single user. Examples:
+Single session is the general case, and is a good fit for (relatively short-lived) applications that are typically involving only a single user. Examples:
+- command-line utility like [`craft`](https://github.com/getsentry/craft); <dfn title="Note: this is just an example, assuming craft was instrumented with a Sentry SDK supporting sessions">every execution of a `craft` subcommand</dfn> report a single session to Sentry
+- user interacting with a mobile app
+- user loading a web site with their favorite browser
 
-- a command-line utility like `craft`; every execution of a `craft` subcommand would report a single session to Sentry
-- a user interacting with a mobile app
-- a user loading a web site with their favorite browser
-
-Session aggregates are a special case for batching up individual sessions for efficiency of reporting them. This mode is used when sending individual sessions would be undesirable or unpractical. That is typically the choice for programs that run for an arbitrarily long time and handle larger throughputs for different users, such as web servers, background job workers, etc. Note that for those types of application, a better definition of session matches the execution of a single HTTP request or task, instead of a single execution of the whole application process.
+Session aggregates are used when sending individual sessions would be undesirable or unpractical. To constrain resource usage (namely memory and network), SDKs keep track of summary information about a batch of sessions that occured in the recent past, never actually having to deal with session objects representing the individual sessions that make up the aggregate. This mode is the choice for applications that run for an arbitrarily long time and handle larger throughputs for potentially multiple users, such as web servers, background job workers, etc. Note that for those types of application, a better definition of session matches the execution of a single HTTP request or task, instead of a single execution of the whole application process.
 
 In either case, SDKs should create and report sessions by default, choosing to report them individually or as aggregates depending on the type of application.
 
-If an SDK can detect that an application is better served by aggregated sessions, then it must not report an application-wide session (it may be created during SDK initialization but must be aborted and never sent to Sentry). As an example, in the Node.js SDK, we can detect an application is probably a web server if it uses the `requestHandler` integration that is provided.
+If an SDK can detect that an application is better served by session aggregates, then it must not report an application-wide session. The application-wide session may still be created during SDK initialization but must be aborted and never sent to Sentry. As an example, in the Node.js SDK, we can detect an application is probably a web server if it uses the `requestHandler` integration that is provided.
 
 ### Individual Session Functionality
 
-## Configuration
+#### Configuration
 
 - On by default for global/static API;
 User should be able to disable sessions if they don't want to track them.
@@ -461,7 +460,7 @@ Pre-requisites for reporting sessions and determining Release Health of projects
 
 (Maybe, needs discussion) if I pre-requisite cannot be detect (for example, no good way to determine release version), then we set some default value so that we can always report sessions by default (depending on discussion, this might not be a change in SDK code, but in Relay, basically removing hard-requirements in the session payload).
 
-## Lifetime of a Session
+#### Lifetime of a Session
 
 Sessions should be enabled by default only for the global hub/client that is initialized by Sentry.init, and disabled by default for any other manually created client.
 A session is started when the SDK is initialized (ideally when the default client is bound to the global hub) and ended when one of these conditions happen:
@@ -472,14 +471,20 @@ The program terminates with an unhandled promise rejection.
 
 Care must be taken to never attempt to send new session payloads to Sentry for a session that is already ended. For example, if the user manually ends the session with Hub.endSession(), there should not be any new updates to the session when the program terminates.
 
-## When Session Goes to Sentry
+#### Session Atrributes and Mutability
+
+#### Sending Session to Sentry
 
 Session is sent initially after a certain (initially hard-coded, less config is better) delay (something between 1s to 30s TBD), and then updated with the duration and final status and error count when the program terminates. Note that, as an optimization, short lived programs will not send 2 session requests to Relay, but only the final one with status and duration.
 
-
 ### Session Aggregates Functionality
 
+#### Configuration
+
 Sessions should be enabled by default, a session is started as soon as a request is received by a web server and ends as soon as the response is fully sent back.
+
+#### Lifetime of a Session
+
 Sessions are never tracked nor sent individually, instead they are aggregated and the aggregates are sent every 30s and a final time when the web server is terminating.
 As an implementation hint to the point above, when a "Client" is closed or flushed the associated  "Session Flusher" shall also be flushed and submit the current aggregates the the transport, before the transport is flushed/closed.
 Make sure this works reasonably for Serverless — there we shall not use "request mode" and SessionFlusher because we cannot have any work that happens outside of the request-response flow.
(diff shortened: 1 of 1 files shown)


src/docs/sdk/sessions.mdx
```diff
@@ -427,3 +427,65 @@ The most basic API exposed is on the hub level and lets you start and stop sessi
...
+(Maybe, needs discussion) if I pre-requisite cannot be detect (for example, no good way to determine release version), then we set some default value so that we can always report sessions by default (depending on discussion, this might not be a change in SDK code, but in Relay, basically removing hard-requirements in the session payload).
+
+#### Lifetime of a Session
+
+Sessions should be enabled by default only for the global hub/client that is initialized by Sentry.init, and disabled by default for any other manually created client.
+A session is started when the SDK is initialized (ideally when the default client is bound to the global hub) and ended when one of these conditions happen:
+The Hub.endSession() method is explicitly called; or
+The program terminates without errors; or
+The program terminates with an unhandled exception; or
+The program terminates with an unhandled promise rejection.
```
ahmedetefy, line 470: I think this is Node specific.... unhandled promise rejections do not exist in Python for example...

src/docs/sdk/sessions.mdx
```diff
@@ -427,3 +427,65 @@ The most basic API exposed is on the hub level and lets you start and stop sessi
...
+
+### Session Aggregates Functionality
+
+#### Configuration
+
+Sessions should be enabled by default, a session is started as soon as a request is received by a web server and ends as soon as the response is fully sent back.
+
+#### Lifetime of a Session
+
+Sessions are never tracked nor sent individually, instead they are aggregated and the aggregates are sent every 30s and a final time when the web server is terminating.
```
ahmedetefy, line 488: every 60 minutes*

src/docs/sdk/sessions.mdx
```diff
@@ -427,3 +427,65 @@ The most basic API exposed is on the hub level and lets you start and stop sessi
...
+As an implementation hint to the point above, when a "Client" is closed or flushed the associated  "Session Flusher" shall also be flushed and submit the current aggregates the the transport, before the transport is flushed/closed.
```
ahmedetefy, line 489: I am not sure if SessionFlusher exists in sdks other than Python and JS, do they?

src/docs/sdk/sessions.mdx
```diff
@@ -427,3 +429,65 @@ The most basic API exposed is on the hub level and lets you start and stop sessi
 `auto_session_tracking`
 
 > This enables / disables automatic session tracking through integrations.
+
+
+## SDK Implementation Guideline
+
+We track the health of each release of projects in Sentry by sending session payloads from SDKs.
+The session payload provide data such as session duration and the presence or absence of errors/crashes.
```
ahmedetefy, line 437: Not sure this is entirely a correct statement because we do not send duration in SessionAggregates

src/docs/sdk/sessions.mdx
```diff
@@ -427,3 +429,65 @@ The most basic API exposed is on the hub level and lets you start and stop sessi
...
+- command-line utility like [`craft`](https://github.com/getsentry/craft); <dfn title="Note: this is just an example, assuming craft was instrumented with a Sentry SDK supporting sessions">every execution of a `craft` subcommand</dfn> report a single session to Sentry
+- user interacting with a mobile app
+- user loading a web site with their favorite browser
+
+Session aggregates are used when sending individual sessions would be undesirable or unpractical. To constrain resource usage (namely memory and network), SDKs keep track of summary information about a batch of sessions that occured in the recent past, never actually having to deal with session objects representing the individual sessions that make up the aggregate. This mode is the choice for applications that run for an arbitrarily long time and handle larger throughputs for potentially multiple users, such as web servers, background job workers, etc. Note that for those types of application, a better definition of session matches the execution of a single HTTP request or task, instead of a single execution of the whole application process.
```
ahmedetefy, line 448: > never actually having to deal with session objects representing the individual sessions that make up the aggregate

This is only valid for Node afaik
Python does use the Session object up until before it is added to the Transport
Not sure that is what your intention was with this comment but I guess it could be true if you are referring to how its finally sent to the server

src/docs/sdk/sessions.mdx
```diff
@@ -427,3 +429,65 @@ The most basic API exposed is on the hub level and lets you start and stop sessi
...
+
+In either case, SDKs should create and report sessions by default, choosing to report them individually or as aggregates depending on the type of application.
+
+If an SDK can detect that an application is better served by session aggregates, then it must not report an application-wide session. The application-wide session may still be created during SDK initialization but must be aborted and never sent to Sentry. As an example, in the Node.js SDK, we can detect an application is probably a web server if it uses the `requestHandler` integration that is provided.
```
ahmedetefy, line 452: > application-wide session
Would rename this to Single session to have consistent naming across the doc

ahmedetefy, line 452: >in the Node.js SDK, we can detect an application is probably a web server if it uses the `requestHandler` integration that is provided

I would also add
> And in python we detect that an application is probably a web server through the WSGI integration

Just for better clarification

src/docs/sdk/sessions.mdx
```diff
@@ -427,3 +429,65 @@ The most basic API exposed is on the hub level and lets you start and stop sessi
...
+
+Pre-requisites for reporting sessions and determining Release Health of projects in Sentry, such as release should be automatically detected by the SDK such as by looking up env vars.
+
+(Maybe, needs discussion) if I pre-requisite cannot be detect (for example, no good way to determine release version), then we set some default value so that we can always report sessions by default (depending on discussion, this might not be a change in SDK code, but in Relay, basically removing hard-requirements in the session payload).
+
+#### Lifetime of a Session
+
+Sessions should be enabled by default only for the global hub/client that is initialized by Sentry.init, and disabled by default for any other manually created client.
+A session is started when the SDK is initialized (ideally when the default client is bound to the global hub) and ended when one of these conditions happen:
+The Hub.endSession() method is explicitly called; or
```
ahmedetefy, line 469: Also it would be ended if a user starts a session
which is probably why we need to document that if a user wants to explicitly start and end their own sessions they should disable autoSessionTracking?

src/docs/sdk/sessions.mdx
```diff
@@ -427,3 +429,65 @@ The most basic API exposed is on the hub level and lets you start and stop sessi
...
+The program terminates without errors; or
+The program terminates with an unhandled exception; or
+The program terminates with an unhandled promise rejection.
+
+Care must be taken to never attempt to send new session payloads to Sentry for a session that is already ended. For example, if the user manually ends the session with Hub.endSession(), there should not be any new updates to the session when the program terminates.
```
ahmedetefy, line 474: I would clarify this a bit better by saying that we should not send sessions for sessions that are in terminal states 
Ref: https://develop.sentry.dev/sdk/sessions/

>String, optional, default is ok. The current status of the session. A session can only be in two states effectively: ok which means the session is alive or one of the terminal states. When a session is moved away from ok it must not be updated anymore.

src/docs/sdk/sessions.mdx
```diff
@@ -427,3 +429,65 @@ The most basic API exposed is on the hub level and lets you start and stop sessi
...
+
+#### Session Atrributes and Mutability
+
+#### Sending Session to Sentry
+
+Session is sent initially after a certain (initially hard-coded, less config is better) delay (something between 1s to 30s TBD), and then updated with the duration and final status and error count when the program terminates. Note that, as an optimization, short lived programs will not send 2 session requests to Relay, but only the final one with status and duration.
```
ahmedetefy, line 480: It is not so much a delay as an interval every 1 min

ahmedetefy, line 480: >Note that, as an optimization, short lived programs will not send 2 session requests to Relay, but only the final one with status and duration.

I am not sure what that means?

src/docs/sdk/sessions.mdx
```diff
@@ -427,3 +429,65 @@ The most basic API exposed is on the hub level and lets you start and stop sessi
...
+
+#### Configuration
+
+Sessions should be enabled by default, a session is started as soon as a request is received by a web server and ends as soon as the response is fully sent back.
+
+#### Lifetime of a Session
+
+Sessions are never tracked nor sent individually, instead they are aggregated and the aggregates are sent every 30s and a final time when the web server is terminating.
+As an implementation hint to the point above, when a "Client" is closed or flushed the associated  "Session Flusher" shall also be flushed and submit the current aggregates the the transport, before the transport is flushed/closed.
+Make sure this works reasonably for Serverless — there we shall not use "request mode" and SessionFlusher because we cannot have any work that happens outside of the request-response flow.
```
ahmedetefy, line 492: SessionFlusher exists for both Single Sessions and SessionAggregates in Python 
SessionFlusher that only serves SessionAggregates is just an implementation detail in Node

src/docs/sdk/sessions.mdx
```diff
@@ -427,3 +429,65 @@ The most basic API exposed is on the hub level and lets you start and stop sessi
...
+Provide an easy way to integrate with existing Node frameworks (Express, Next.js, Koa).
```
ahmedetefy, line 493: I would also add explanation on how the `requestHandler` is used for that in Node frameworks and `WSGI` is used for Python frameworks

ahmedetefy (changes requested):
Another pass at this

HazAT:
This is good enough to be merged and we can update this over time. Given this is only a guideline and not defacto the truth. Not merging this will do more harm than merging it, let's keep iterating if necessary.

robblovell:
Hi, I was directed to ask questions about session duration to this pull request:

Situation: I have a CLI command that I wish to track duration between commands, one command opens something, the other closes something. In this case, the longer period between session is better. Unlike tracking durations of web request fulfillment, I need to track how long the user has a session open, where "open" means they are using the product. Short durations mean that they aren't really using the product.

What I have tried and would like validation of is the following:

```
if (params.clean) { // this is the end of a session, log the duration to sentry
        const startTimestamp: string = await session.get('timestamp')
        const duration = (((timestamp as any) - (new Date(startTimestamp) as any))) // in ms

        if (duration) {
            Sentry.getCurrentHub().startSession({ did: session.signatureHash, duration, ignoreDuration: false })
            Sentry.getCurrentHub().endSession()
        }
    } else { // This is the start of the session, set the session timestamp.
        Sentry.getCurrentHub().startSession({ did: session.signatureHash, ignoreDuration: true })
        session.set('timestamp', timestamp.toISOString())
        Sentry.getCurrentHub().endSession()
    }
```
Here, 'clean' means the user is closing a session, otherwise, the session is being opened. We track the start timestamp in our own session object and then calculate the duration from this when the 'clean' is given. 

I have also tried:
```
            Sentry.captureMessage(util.format('%O', { duration: durationSeconds, params, session }))
```
But I am not sure how this would be useful.

Will this work to record the duration for our open/clean cycles of usage so that we can visualize the results in a dashboard or in the performance area of the Sentry website?

How should this be done?

PR #328, opened 2021-05-06
feat(flagr): Remove the need for allowed features

wmak:
- We no longer need allowed_features

### DIFF
--- a/src/docs/feature-flags.mdx
+++ b/src/docs/feature-flags.mdx
@@ -10,7 +10,7 @@ You can find a list of features available by looking at
 `sentry/features/__init__.py`. They're declared on the `FeatureManager` like so:
 
 ```python
-default_manager.add("organizations:onboarding", OrganizationFeature)  # NOQA
+default_manager.add("organizations:onboarding", OrganizationFeature, True)  # NOQA
 ```
 
 The feature can be enabled with the following in your `sentry.conf.py`, usually located at `~/.sentry/`:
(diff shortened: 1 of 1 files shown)


src/docs/feature-flags.mdx
```diff
@@ -10,7 +10,7 @@ You can find a list of features available by looking at
 `sentry/features/__init__.py`. They're declared on the `FeatureManager` like so:
 
 ```python
-default_manager.add("organizations:onboarding", OrganizationFeature)  # NOQA
+default_manager.add("organizations:onboarding", OrganizationFeature, True)  # NOQA
 ```
```
Zylphrex, line 14: Can we add something here to explain what `True` is doing? Or link them to the explanation below? Unless the intent is for all feature flagging to be done through flagr, this might confuse some people when they read this.

markstory (approved):

src/docs/feature-flags.mdx
k-fish: The true could also be a kwarg and implicitly explain what it is everywhere you see it.

k-fish (approved):

src/docs/feature-flags.mdx
wmak: went for both
```python
# Don't set entity_feature, or set it to False if you don't plan to use Flagr
default_manager.add("organizations:onboarding", OrganizationFeature, entity_feature=True)
```

PR #329, opened 2021-05-06
Proposal: Culture Context

ueman:
This PR is a proposal to add a `current culture` context to Contexts. 
It contains various kinds of information around the culture which is in use.

Who is using this (at least these are the ones I'm aware of):
- https://github.com/getsentry/sentry-dotnet
- This Draft PR is also adapting to it: https://github.com/getsentry/sentry-dart/pull/452

Where this also might be useful:
- Android
- iOS
- Desktop Linux
- Windows
- macOS
- Browser (?)

Additional relevant properties not yet included:
- decimal seperator
- units (weight, length, money)

Why do I want to have this information?
When my software supports multiple cultures/locales and I'm
- Parsing strings into numbers
- Calculating with units
- Displaying dates

### DIFF
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -375,3 +375,26 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
   }
 }
 ```
+
+## Current Culture Context
+
+Current Culture Context describes certain properties of the culture in which 
+the software is used.
+
+The `type` and default key is `"current culture"`.
+
+`calendar`
+
+: _Optional_. For example `GregorianCalendar`. Free form string. 
+
+`displayName`
+
+: _Optional_. Human readable name of the current culture. For example `English (United States)`
+
+`locale`
+
+: _Optional_. Two letter code of the current culture. For example `en-US`
+
+`24hourFormat`
+
+: _Optional_. boolean, either `true` or `false`.
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -395,6 +395,6 @@ The `type` and default key is `"current culture"`.
 
 : _Optional_. Two letter code of the current culture. For example `en-US`
 
-`24hourFormat`
+`is_24_hour_format`
 
 : _Optional_. boolean, either `true` or `false`.
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -376,7 +376,7 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
 }
 ```
 
-## Current Culture Context
+## Culture Context
 
 Current Culture Context describes certain properties of the culture in which 
 the software is used.
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -387,7 +387,7 @@ The `type` and default key is `"current culture"`.
 
 : _Optional_. For example `GregorianCalendar`. Free form string. 
 
-`displayName`
+`display_name`
 
 : _Optional_. Human readable name of the current culture. For example `English (United States)`
 
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -381,7 +381,7 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
 Current Culture Context describes certain properties of the culture in which 
 the software is used.
 
-The `type` and default key is `"current_culture"`.
+The `type` and default key is `"culture"`.
 
 `calendar`
 
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -389,7 +389,7 @@ The `type` and default key is `"culture"`.
 
 `display_name`
 
-: _Optional_. Human readable name of the current culture. For example `English (United States)`
+: _Optional_. Human readable name of the culture. For example `English (United States)`
 
 `locale`
 
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -378,7 +378,7 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
 
 ## Culture Context
 
-Current Culture Context describes certain properties of the culture in which 
+Culture Context describes certain properties of the culture in which 
 the software is used.
 
 The `type` and default key is `"culture"`.
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -381,7 +381,7 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
 Current Culture Context describes certain properties of the culture in which 
 the software is used.
 
-The `type` and default key is `"current culture"`.
+The `type` and default key is `"current_culture"`.
 
 `calendar`
 
@@ -393,8 +393,12 @@ The `type` and default key is `"current culture"`.
 
 `locale`
 
-: _Optional_. Two letter code of the current culture. For example `en-US`
+: _Optional_. The name identifier, usually following the RFC 4646. For example `en-US` or `pt-BR`.
 
 `is_24_hour_format`
 
 : _Optional_. boolean, either `true` or `false`.
+
+`timezone`
+
+: _Optional_. The timezone of the locale. For example, `Europe/Vienna`.


bruno-garcia:
For context: https://github.com/getsentry/sentry-dart/pull/452#discussion_r626895094

untitaker:
Those props all read like device settings to me. On device context we already have a timezone prop which would also fit into "culture"

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -375,3 +375,26 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
...
+
+: _Optional_. For example `GregorianCalendar`. Free form string. 
+
+`displayName`
+
+: _Optional_. Human readable name of the current culture. For example `English (United States)`
+
+`locale`
+
+: _Optional_. Two letter code of the current culture. For example `en-US`
```
bruno-garcia, line 396: In .NET it's:

> The name is a combination of an ISO 639 two-letter lowercase culture code associated with a language and an ISO 3166 two-letter uppercase subculture code associated with a country or region.
```suggestion

So maybe we change to:
: _Optional_. The name identifier, usually following the RFC 4646. For example `en-US` or `pt-BR`.
```

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -375,3 +375,26 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
   }
 }
 ```
+
+## Current Culture Context
```
bruno-garcia, line 379: This "current culture" is a .NET thing because you can have a different culture set on a thread. So the device culture can be one, and the thread another. We could simplify here (and/or reword it) if that makes sense.

bruno-garcia (commented):
Thanks for taking the initiative! This is great.
Lets discuss some details before landing this one.

src/docs/sdk/event-payloads/contexts.mdx
marandaneto: @bruno-garcia would be possible then that you are sending 2+ current_culture values? per thread for example?

marandaneto:
> Those props all read like device settings to me. On device context we already have a timezone prop which would also fit into "culture"

indeed, but there's a difference, this configuration could be per App or per OS.
so for example, I chose my App to be en-US but my phone is pt-BR, and I'd like to know both.
we've to allow multiple configurations, having a `current_culture` would make that possible as we can have multiple keys with the same `type`.

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -375,3 +375,26 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
...
+
+Current Culture Context describes certain properties of the culture in which 
+the software is used.
+
+The `type` and default key is `"current culture"`.
```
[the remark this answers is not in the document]
marandaneto: yep, thats the standard already, lets stick to it

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -375,3 +375,26 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
...
+
+`displayName`
+
+: _Optional_. Human readable name of the current culture. For example `English (United States)`
+
+`locale`
+
+: _Optional_. Two letter code of the current culture. For example `en-US`
+
+`24hourFormat`
```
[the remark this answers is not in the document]
marandaneto: wondering if in another language, that are different formats other than 12-24, guess not ^^

src/docs/sdk/event-payloads/contexts.mdx
marandaneto, line 379: maybe it'd make sense to move some stuff out from the device context to this one then, like `timezone`

src/docs/sdk/event-payloads/contexts.mdx
marandaneto: already exists under `device.language`, we'd have to move it

src/docs/sdk/event-payloads/contexts.mdx
bruno-garcia: .NET currently sends two (if they differ). `Culture` and `Culture Culture`. With this format as key :hides:

But if this lands in the spec we'd be adjusting accordingly

ueman:
Open questions:

1. Is `Current Culture` the best name for this?
2. Should some related properties of the device moved to `Current Culture`? 
    1. And if so, should `Current Culture` have an `origin` property which describes if the info comes from a thread or from device settings.

src/docs/sdk/event-payloads/contexts.mdx
jan-auer: Contexts per thread is a concept we don't really support, but you can send several contexts with different keys.

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -375,3 +375,30 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
   }
 }
 ```
+
+## Current Culture Context
+
+Current Culture Context describes certain properties of the culture in which 
+the software is used.
+
+The `type` and default key is `"current_culture"`.
```
jan-auer, line 384: Please use `type: culture` here, since that is the true type of the context. You can still have a `current_culture` or other key for SDK-specific conventions.

jan-auer (commented):
Overall LGTM. I'd just ask you to make it a "Culture" context (type) and then you can have a "current culture" custom key for .NET.

ueman:
> @ueman which properties would that be? 

- `timezone`
- `language` (doesn't seem to be documented [here](https://github.com/getsentry/develop/pull/329#issuecomment-834106286) but it's included in the [source code](https://github.com/getsentry/sentry/blob/9707fd3f98fa51d5d363cbe34b59aa51588db246/static/app/components/events/contexts/device/getDeviceKnownDataDetails.tsx#L182) already)

> Since a number of SDKs, particularly the Cocoa SDK, has been sending that context for a while now, there's probably little value in changing that.

I agree that there is litte value in changing that, also because of 

> so for example, I chose my App to be en-US but my phone is pt-BR, and I'd like to know both.

from this [comment](https://github.com/getsentry/develop/pull/329#issuecomment-834106286)

jan-auer:
We would have to clarify that, since the device context was likely filled with app-local information. That said, I'm +1 on keeping both.

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -375,3 +375,30 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
   }
 }
 ```
+
+## Culture Context
+
+Current Culture Context describes certain properties of the culture in which
```
ueman, line 381:
```suggestion
Culture Context describes certain properties of the culture in which 
```

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -375,3 +375,30 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
...
+
+The `type` and default key is `"culture"`.
+
+`calendar`
+
+: _Optional_. For example `GregorianCalendar`. Free form string. 
+
+`display_name`
+
+: _Optional_. Human readable name of the current culture. For example `English (United States)`
```
ueman, line 392: This could be inferred from the local identifier.

dashed:
For browsers, information could be extracted from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl

bruno-garcia (approved):
LGTM, maybe @mitsuhiko wants to 👀  this before we merge.
If he doesn't reply in a few days it means it's OK. :D

src/docs/sdk/event-payloads/contexts.mdx
marandaneto: @ueman what's about deprecating `device.language` and moving to culture? I guess it'd make sense

src/docs/sdk/event-payloads/contexts.mdx
marandaneto: as timezone now belongs to culture as well, lets deprecate the one in the device context?

ueman: See 
- https://github.com/getsentry/develop/pull/329#issuecomment-836457964
- https://github.com/getsentry/develop/pull/329#issuecomment-836485242

src/docs/sdk/event-payloads/contexts.mdx
ueman: See 
- https://github.com/getsentry/develop/pull/329#issuecomment-836457964
- https://github.com/getsentry/develop/pull/329#issuecomment-836485242

src/docs/sdk/event-payloads/contexts.mdx
marandaneto: deprecating is more like, adding a note that this is also available under the new culture type, not really removing.

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -375,3 +375,30 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
...
+
+`locale`
```
marandaneto, line 394: aka device.language

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -375,3 +375,30 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
...
+
+: _Optional_. The name identifier, usually following the RFC 4646. For example `en-US` or `pt-BR`.
+
+`is_24_hour_format`
+
+: _Optional_. boolean, either `true` or `false`.
+
+`timezone`
```
marandaneto, line 402: aka device.timezone

src/docs/sdk/event-payloads/contexts.mdx
marandaneto, line 392:
```suggestion
: _Optional_. Human readable name of the culture. For example `English (United States)`
```

marandaneto:
Screenshot 2021-05-17 at 10 44 03

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -375,3 +375,30 @@ The following example illustrates the contexts part of the <Link to="/sdk/event-
...
+
+The `type` and default key is `"culture"`.
+
+`calendar`
+
+: _Optional_. For example `GregorianCalendar`. Free form string. 
+
+`display_name`
+
+: _Optional_. Human readable name of the culture. For example `English (United States)`
```
ueman, line 392: `display_name` should probably be inferred from `local`. So if `local` is `en-US`, `display_language` should be `English (United States)`.

marandaneto:
@ueman I'm discussing this change internally

PR #330, opened 2021-05-06
Proposal: Darkmode field for OS context

ueman:
This PR proposed to add a `brightness` field to the OS context.

All major OSs (Windows, macOS, Linux (Ubuntu), Android, iOS) have implemented dark mode. 
Darkmode is a system wide option, thus I added this to the OS context.

I could imagine that it also makes sense as a field of the app.

### DIFF
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -168,7 +168,7 @@ The `type` and default key is `"os"`.
 
 `brightness`
 
-: _Optional_. Either `light` or `dark`. Describes wether the OS runs in dark mode or not.
+: _Optional_. Either `light` or `dark`. Describes whether the OS runs in dark mode or not.
 
 `raw_description`
 
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -166,6 +166,10 @@ The `type` and default key is `"os"`.
 
 : _Optional_. A flag indicating whether the OS has been jailbroken or rooted.
 
+`brightness`
+
+: _Optional_. Either `light` or `dark`. Describes wether the OS runs in dark mode or not.
+
 `raw_description`
 
 : _Optional_. An unprocessed description string obtained by the operating
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -166,7 +166,7 @@ The `type` and default key is `"os"`.
 
 : _Optional_. A flag indicating whether the OS has been jailbroken or rooted.
 
-`brightness`
+`theme`
 
 : _Optional_. Either `light` or `dark`. Describes whether the OS runs in dark mode or not.
 


src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -166,6 +166,10 @@ The `type` and default key is `"os"`.
 
 : _Optional_. A flag indicating whether the OS has been jailbroken or rooted.
 
+`brightness`
+
+: _Optional_. Either `light` or `dark`. Describes wether the OS runs in dark mode or not.
```
bruno-garcia, line 171:
```suggestion
: _Optional_. Either `light` or `dark`. Describes whether the OS runs in dark mode or not.
```

bruno-garcia (commented):
Not too familiar with details to suggest anything related to name/possible values/etc but the general idea makes sense!

@matejminar Do we need to add it to the `contexts.tsx` or something to get the proper title?

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -166,6 +166,10 @@ The `type` and default key is `"os"`.
 
 : _Optional_. A flag indicating whether the OS has been jailbroken or rooted.
 
+`brightness`
+
+: _Optional_. Either `light` or `dark`. Describes whether the OS runs in dark mode or not.
```
marandaneto, line 171: wondering if this should be OS only, can I have an OS with light mode but my App is Dark mode?

ueman: That's why I said that this field also makes sense in the App context.
Though the at the app level the SDK can't really detect it and the user would have to set it himself.

A lot of apps have options for forcing it to be light/dark or to have it detect the OS mode. Like Twitter for example.

marandaneto: oh just noticed its in the description, my bad :P

matejminar:
> Do we need to add it to the contexts.tsx or something to get the proper title?

@bruno-garcia It should show up by default at the very bottom of the list with label used as the title

If we want to change the order we need to add it here:
https://github.com/getsentry/sentry/blob/9707fd3f98fa51d5d363cbe34b59aa51588db246/static/app/components/events/contexts/operatingSystem/operatingSystem.tsx#L18-L23

and if we want to format the title then we can do that here:
https://github.com/getsentry/sentry/blob/6268868306ac4d6372100945e4fb9b6d8a00651e/static/app/components/events/contexts/operatingSystem/getOperatingSystemKnownDataDetails.tsx#L11-L42

_nit:_ should we call it `brightness`? for some reason I immediately think about the number/percentage, would something like `theme` be more fitting?

ueman:
> _nit:_ is `brightness` the right name for this? for some reason I immediately think about the number/percentage, would something like `theme` be more fitting?

Yeah, you are probably right.

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -166,6 +166,10 @@ The `type` and default key is `"os"`.
 
 : _Optional_. A flag indicating whether the OS has been jailbroken or rooted.
 
+`brightness`
```
ueman, line 169:
```suggestion
`theme`
```

marandaneto (approved):

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -166,6 +166,10 @@ The `type` and default key is `"os"`.
 
 : _Optional_. A flag indicating whether the OS has been jailbroken or rooted.
 
+`theme`
+
+: _Optional_. Either `light` or `dark`. Describes whether the OS runs in dark mode or not.
```
marandaneto, line 171: `theme` is under the `device` context, but the description is "whether the OS runs", we also have the `os` context, but its more related to the versioning, so it makes sense under the device I guess

PR #335, opened 2021-05-10
Document missing properties

ueman:
This documents some missing properties which already have made it into the spec.

cc @marandaneto and @bruno-garcia

### DIFF
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -73,6 +73,14 @@ The `type` and default key is `"device"`.
 
 : _Optional_. The screen resolution. (e.g.: 800x600, 3040x1444).
 
+`screen_height_pixels`
+
+: _Optional_. The height of the screen.
+
+`screen_width_pixels`
+
+: _Optional_. The width of the screen.
+
 `screen_density`
 
 : _Optional_. A floating point denoting the screen density.
@@ -136,6 +144,10 @@ The `type` and default key is `"device"`.
 `timezone`
 
 : _Optional_. The timezone of the device. For example, `Europe/Vienna`.
+
+`language`
+
+: _Optional_. The language of the device. For example, `en-US`.
 
 ## OS Context
 


src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -137,6 +145,10 @@ The `type` and default key is `"device"`.
 
 : _Optional_. The timezone of the device. For example, `Europe/Vienna`.
 
+`language`
```
ueman, line 148: See https://github.com/getsentry/sentry/blob/9707fd3f98fa51d5d363cbe34b59aa51588db246/static/app/components/events/contexts/device/getDeviceKnownDataDetails.tsx#L182

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -73,6 +73,14 @@ The `type` and default key is `"device"`.
 
 : _Optional_. The screen resolution. (e.g.: 800x600, 3040x1444).
 
+`screen_height_pixels`
```
ueman, line 76: See https://github.com/getsentry/sentry/blob/9707fd3f98fa51d5d363cbe34b59aa51588db246/static/app/components/events/contexts/device/getDeviceKnownDataDetails.tsx#L227

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -73,6 +73,14 @@ The `type` and default key is `"device"`.
...
+
+: _Optional_. The height of the screen.
+
+`screen_width_pixels`
```
ueman, line 80: See https://github.com/getsentry/sentry/blob/9707fd3f98fa51d5d363cbe34b59aa51588db246/static/app/components/events/contexts/device/getDeviceKnownDataDetails.tsx#L237

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -73,6 +73,14 @@ The `type` and default key is `"device"`.
 
 : _Optional_. The screen resolution. (e.g.: 800x600, 3040x1444).
 
+`screen_height_pixels`
+
+: _Optional_. The height of the screen.
```
bruno-garcia, line 78: Ideally this would be explicit about the unit, and/or an example.And clarify how does it relate to `screen_resolution`?

If `screen_resolution` is 800x600 this would be `800` and `screen_width_pixels` would be `600`?
I wonder if any SDK sends this already, couldn't find this key in the Java/Android or Cocoa SDKs

src/docs/sdk/event-payloads/contexts.mdx
bruno-garcia: Nice! Do we have examples of SDKs that send it in this format or another already?

ueman: Android does it. The format is for example en_US.

src/docs/sdk/event-payloads/contexts.mdx
marandaneto: Android SDK was previously sending `screen_resolution` but we added screen_width_pixels and screen_height_pixels, because we also send screen_density and screen_dpi, I don't remember the reason why we've split up, maybe because they were read from separate fields but required a ValueXValue format in the protocol so we needed to string format to do calculations

ueman: I'm also never sure if it's width times height or the other way around so I really appreciate two properties which state exactly what they're for.

marandaneto: maybe that was the reason :D

src/docs/sdk/event-payloads/contexts.mdx
bruno-garcia: Feel free to raise an issue on getsentry/sentry requesting that. 

I requested one feature once and it landed super quickly: https://github.com/getsentry/sentry/issues/24209

bruno-garcia (approved):
LGTM

marandaneto:
thanks @ueman

Issue #307, opened 2021-04-15
default span status if there was an error or http response code does not match any span status code


marandaneto:
### Steps to Reproduce

Trigger an unknown HTTP response code (that does not match any span status code) or an exception so span either won't have a status or it'll diverge across SDKs

### Expected Result

All SDKs match the default status in such cases:

an exception was thrown, maybe `INTERNAL_ERROR`
HTTP response code does not match any span status code, set maybe `UNKNOWN`? we could also make this based on HTTP code range.

eg HTTP response code is 502, we don't have such span status code, what's the default for 5xx range?
eg HTTP response code is 402, same as above, what's the default for 4xx range?

This should be aligned across SDKs for a better experience.

### Actual Result

Android OkHttp sets `INTERNAL_ERROR` as status if there was an error or HTTP response code does not match span status code;
Java Spring sets no status at all;
.NET sets `Unknown`;

marandaneto:
@bruno-garcia @Tyrrrz @rhcarvalho @maciejwalkowiak @brustolin

marandaneto:
for Java, we are changing to:

If there's an exception, we set the span status to `INTERNAL_ERROR` and rethrow.
If the request was successful, match the HTTP response code with span status, if it does not match, leave it without, so no misleading status.

This only applies to spans created out of HTTP requests

marandaneto:
for non HTTP requests, the span should be finished with `OK` if everything went well or `INTERNAL_ERROR` in case there's an error.
otherwise, the UI/relay marks the status as `unknown`.

PR #341, opened 2021-05-19
Document Expected Features for the HTTP client integrations

### LINKED ISSUE #254: Document in SDK `Expected Features` the HTTP integrations
In [SDK Expected Features](https://develop.sentry.dev/sdk/features/) we need some guideline of what we expect from integrating with HTTP clients in SDKs, when that's possible.

For example:
* Add a breadcrumb to the Scope (so error events show that request)
  * Mention the breadcrumb categories and types that make sense here. Include HTTP status. 
  * `category` could be the name of the HTTP integration. Type is `http`. [Docs](https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/#breadcrumb-types)
* Get some metrics (that for now can go as breadcrumb.data Example: bytes In, Bytes Out).
* Should the SDK raise an event if a 5xx comes? If so what contexts.. If not, how does it propagate context upstream?

If Performance Monitoring is supported by the SDK:
* Create a Span, child to the current Span bound to the Hub
* Finish the span as success or failure depending on response HTTP status (500 is a failed Span)
  * We need to align/discuss 4xx and other status code range. 
* Include `sentry-trace` id to propagate a transaction downstream. [Link to specific docs](https://develop.sentry.dev/sdk/performance/#header-sentry-trace).

Link to some examples:

[.NET](https://github.com/getsentry/sentry-dotnet/blob/69126250838daf1f4489bfeae9c538f5f4dd1a98/src/Sentry/SentryHttpMessageHandler.cs#L45-L89)
[Java](https://github.com/getsentry/sentry-java/blob/35c37a5ce7d245fdde819bb142e2f6c8f79e9fd0/sentry-spring/src/main/java/io/sentry/spring/tracing/SentrySpanClientHttpRequestInterceptor.java#L35-L47)
[Python](https://github.com/getsentry/sentry-python/blob/7f65e718d62300afcc880a01e98fa3806722ceb1/sentry_sdk/integrations/stdlib.py#L57)

maciejwalkowiak:
Fixes #254.

### DIFF
--- a/src/docs/sdk/features.mdx
+++ b/src/docs/sdk/features.mdx
@@ -163,7 +163,7 @@ Add a breadcrumb for each outgoing HTTP request after the request finishes:
 - category: `http`
 - data: `url`, `method` (uppercase HTTP method), optionally `requestBodySize` and `responseBodySize` (in bytes)
 
-If Performance Monitoring is both supported by the SDK and enabled in the client application when the transaction is active a new `Span` must be created around the HTTP request.
+If Performance Monitoring is both supported by the SDK and enabled in the client application when the transaction is active a new `Span` must be created around the HTTP request; The `Span#operation` should be `http.client` and the `Span#description` should follow the convention `$METHOD $url` eg `GET https://sentry.io`.
 
 - HTTP request must be enhanced with a [`sentry-trace` HTTP header](/sdk/performance/#header-sentry-trace)
 - span status must match HTTP response status code ([see Span status to HTTP status code mapping](/sdk/event-payloads/span/))
--- a/src/docs/sdk/features.mdx
+++ b/src/docs/sdk/features.mdx
@@ -154,3 +154,17 @@ This is mostly useful on mobile clients where connectivity is often not availabl
 ## HTTP Proxy
 
 Ability to use an HTTP proxy. Often easy to implement using the existing HTTP client.  This should be picked up from the system config if possible or explicit config in the client options.
+
+## HTTP Client Integrations
+
+Add breadcrumb for each outgoing HTTP request after the request finishes:
+
+- type: `http`
+- category: `http`
+- data: `url`, `method` (uppercase HTTP method), optionally `requestBodySize` and `responseBodySize` (in bytes)
+
+If the Performance Monitoring is supported by the SDK and enabled in the client application, when the transaction is active, a new `Span` must be created around the HTTP request.
+
+- HTTP request must be enhanced with a [`sentry-trace` HTTP header](/sdk/performance/#header-sentry-trace)
+- span status must match HTTP response status code ([see Span status to HTTP status code mapping](/sdk/event-payloads/span/))
+- when network error occurs, span status must be set to `INTERNAL_SERVER_ERROR`
--- a/src/docs/sdk/features.mdx
+++ b/src/docs/sdk/features.mdx
@@ -163,7 +163,10 @@ Add a breadcrumb for each outgoing HTTP request after the request finishes:
 - category: `http`
 - data: `url`, `method` (uppercase HTTP method), optionally `requestBodySize` and `responseBodySize` (in bytes)
 
-If Performance Monitoring is both supported by the SDK and enabled in the client application when the transaction is active a new `Span` must be created around the HTTP request; The `Span#operation` should be `http.client` and the `Span#description` should follow the convention `$METHOD $url` eg `GET https://sentry.io`.
+If Performance Monitoring is both supported by the SDK and enabled in the client application when the transaction is active a new `Span` must be created around the HTTP request:
+
+- operation: `http.client`
+- description: `$METHOD $url` (uppercase HTTP method), e.g. `GET https://sentry.io`
 
 - HTTP request must be enhanced with a [`sentry-trace` HTTP header](/sdk/performance/#header-sentry-trace)
 - span status must match HTTP response status code ([see Span status to HTTP status code mapping](/sdk/event-payloads/span/))
--- a/src/docs/sdk/features.mdx
+++ b/src/docs/sdk/features.mdx
@@ -167,7 +167,6 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
 
 - operation: `http.client`
 - description: `$METHOD $url` (uppercase HTTP method), e.g. `GET https://sentry.io`
-
 - HTTP request must be enhanced with a [`sentry-trace` HTTP header](/sdk/performance/#header-sentry-trace)
 - span status must match HTTP response status code ([see Span status to HTTP status code mapping](/sdk/event-payloads/span/))
 - when network error occurs, span status must be set to `internal_error`
--- a/src/docs/sdk/features.mdx
+++ b/src/docs/sdk/features.mdx
@@ -157,7 +157,7 @@ Ability to use an HTTP proxy. Often easy to implement using the existing HTTP cl
 
 ## HTTP Client Integrations
 
-Add breadcrumb for each outgoing HTTP request after the request finishes:
+Add a breadcrumb for each outgoing HTTP request after the request finishes:
 
 - type: `http`
 - category: `http`
--- a/src/docs/sdk/features.mdx
+++ b/src/docs/sdk/features.mdx
@@ -163,7 +163,7 @@ Add breadcrumb for each outgoing HTTP request after the request finishes:
 - category: `http`
 - data: `url`, `method` (uppercase HTTP method), optionally `requestBodySize` and `responseBodySize` (in bytes)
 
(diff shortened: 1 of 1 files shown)


maciejwalkowiak:
cc @marandaneto @bruno-garcia

marandaneto:
cc @PeloWriter for some reason can't add you as a reviewer

src/docs/sdk/features.mdx
```diff
@@ -154,3 +154,17 @@ This is mostly useful on mobile clients where connectivity is often not availabl
 ## HTTP Proxy
 
 Ability to use an HTTP proxy. Often easy to implement using the existing HTTP client.  This should be picked up from the system config if possible or explicit config in the client options.
+
+## HTTP Client Integrations
+
+Add breadcrumb for each outgoing HTTP request after the request finishes:
```
PeloWriter, line 160:
```suggestion
Add a breadcrumb for each outgoing HTTP request after the request finishes:
```

src/docs/sdk/features.mdx
```diff
@@ -154,3 +154,17 @@ This is mostly useful on mobile clients where connectivity is often not availabl
...
+
+- type: `http`
+- category: `http`
+- data: `url`, `method` (uppercase HTTP method), optionally `requestBodySize` and `responseBodySize` (in bytes)
+
+If the Performance Monitoring is supported by the SDK and enabled in the client application, when the transaction is active, a new `Span` must be created around the HTTP request.
```
PeloWriter, line 166:
```suggestion
If Performance Monitoring is both supported by the SDK and enabled in the client application when the transaction is active a new `Span` must be created around the HTTP request.
```

PeloWriter (approved):

PeloWriter:
I'm not a reviewer on the developer site, but happy to add comments if you ping me (as you did for this).

src/docs/sdk/features.mdx
```diff
@@ -154,3 +154,17 @@ This is mostly useful on mobile clients where connectivity is often not availabl
...
+
+- type: `http`
+- category: `http`
+- data: `url`, `method` (uppercase HTTP method), optionally `requestBodySize` and `responseBodySize` (in bytes)
+
+If Performance Monitoring is both supported by the SDK and enabled in the client application when the transaction is active a new `Span` must be created around the HTTP request.
+
+- HTTP request must be enhanced with a [`sentry-trace` HTTP header](/sdk/performance/#header-sentry-trace)
+- span status must match HTTP response status code ([see Span status to HTTP status code mapping](/sdk/event-payloads/span/))
+- when network error occurs, span status must be set to `INTERNAL_SERVER_ERROR`
```
rhcarvalho, line 170: What is meant by "network error" here? There are many things that can happen like failed to resolve DNS, timeout, etc.

Note that `INTERNAL_SERVER_ERROR` is not a valid status (`internal_error` is).

Lacking more context, I guess `unknown_error` is more appropriate as a generic catch-all status, but I look forward to us simplifying the statuses in the near future, the current list is confusing and not well fitting the product.

marandaneto, line 170:
```suggestion
- when network error occurs, span status must be set to `internal_error`
```

marandaneto: @rhcarvalho about the `unknown_error `, See -> https://github.com/getsentry/develop/issues/307#issuecomment-822240763
we've changed it because the description of `unknown_error` is actually:

> An unknown error raised by APIs that don't return enough error information

and in our case, network error is eg you've no connection, something happened that you could not really talk with the endpoint, that's why internal rather unknown.

marandaneto: I agree that status are a bit confusing, but that's what we've right now, we can revisit this in the future

rhcarvalho: I see. Reusing `internal_error` unfortunately confuses with a 500 response.

marandaneto: indeed, but `unknown_error` is also 500 anyway, so it's not any better IMO

rhcarvalho: To be sure we understand the same, while the status values (first column, `States`) is sort of fixed ([known by Relay](https://github.com/getsentry/relay/blob/55127c75d4eeebf787848a05a12150ee5c59acd9/relay-common/src/constants.rs#L283-L306)), the `Description` and `HTTP status code equivalent` are under our control; it is whatever we want to document and it is relatively easy to change. AFAIK nothing in the product depends on those descriptions.

bruno-garcia (commented):
This lgtm but I believe we should circle other folks for thoughts before merging.

@jan-auer @HazAT @rhcarvalho @kamilogorek @Swatinem

src/docs/sdk/features.mdx
```diff
@@ -154,3 +154,17 @@ This is mostly useful on mobile clients where connectivity is often not availabl
...
+
+## HTTP Client Integrations
+
+Add a breadcrumb for each outgoing HTTP request after the request finishes:
+
+- type: `http`
+- category: `http`
+- data: `url`, `method` (uppercase HTTP method), optionally `requestBodySize` and `responseBodySize` (in bytes)
+
+If Performance Monitoring is both supported by the SDK and enabled in the client application when the transaction is active a new `Span` must be created around the HTTP request.
```
marandaneto, line 166: describe span `op` and `desc`
eg `op=http.client` and `desc=$method $url`

also the OkHttp integration now has a `BeforeSpanCallback` which allows customizing the `span`

bruno-garcia: > describe span op and desc
> eg op=http.client and desc=$method $url

You mean just adding these examples? What description do you have in mind?

> also the OkHttp integration now has a BeforeSpanCallback which allows customizing the span

What's the suggestion? To add a note that integrations can take a callback to modify the span directly in a way that the HTTP Context is provided to the callback?

bruno-garcia: the PR looks good to me and got 👍 from @kamilogorek and other folks tagged had 1 month to react.
Happy to merge as-is or feel free to add the suggestions, commit them and merge.

🚀 🚀

marandaneto: my suggestion was just to explain that the `op` should be `http.client` and the description of the span should follow the convention `$method $url` eg `GET https://sentry.io`, so we don't diverge on that.

the `BeforeSpanCallback` can be added later since we're waiting feedback about it yet

marandaneto:
```suggestion
If Performance Monitoring is both supported by the SDK and enabled in the client application when the transaction is active a new `Span` must be created around the HTTP request; The `Span#operation` should be `http.client` and the `Span#description` should follow the convention `$METHOD $url` eg `GET https://sentry.io`.
```

src/docs/sdk/features.mdx
marandaneto: yep, but right now that's how its implemented in some of the SDKs (using inetrnal_error) and its the one that fits the most in the description https://develop.sentry.dev/sdk/event-payloads/transaction/
so unless we fix the SDKs and the status descriptions, let's keep it like this, I'll merge as it is.

src/docs/sdk/features.mdx
```diff
@@ -154,3 +154,17 @@ This is mostly useful on mobile clients where connectivity is often not availabl
...
+
+## HTTP Client Integrations
+
+Add a breadcrumb for each outgoing HTTP request after the request finishes:
+
+- type: `http`
+- category: `http`
+- data: `url`, `method` (uppercase HTTP method), optionally `requestBodySize` and `responseBodySize` (in bytes)
+
+If Performance Monitoring is both supported by the SDK and enabled in the client application when the transaction is active a new `Span` must be created around the HTTP request; The `Span#operation` should be `http.client` and the `Span#description` should follow the convention `$METHOD $url` eg `GET https://sentry.io`.
```
marandaneto, line 166:
```suggestion
If Performance Monitoring is both supported by the SDK and enabled in the client application when the transaction is active a new `Span` must be created around the HTTP request:

- operation: `http.client`
- description: `$METHOD $url` (uppercase HTTP method), e.g. `GET https://sentry.io`
```

src/docs/sdk/features.mdx
```diff
@@ -154,3 +154,20 @@ This is mostly useful on mobile clients where connectivity is often not availabl
...
+
+- type: `http`
+- category: `http`
+- data: `url`, `method` (uppercase HTTP method), optionally `requestBodySize` and `responseBodySize` (in bytes)
+
+If Performance Monitoring is both supported by the SDK and enabled in the client application when the transaction is active a new `Span` must be created around the HTTP request:
+
+- operation: `http.client`
+- description: `$METHOD $url` (uppercase HTTP method), e.g. `GET https://sentry.io`
+
```
marandaneto, line 170:
```suggestion
```

PR #351, opened 2021-06-09
Fix pyenv setup for fish

relaxolotl:
I bumped into a few issues while setting up the environment while using fish as my shell, so this fixes those up and performs a bit of additional cleanup.

Setup script updates: https://github.com/getsentry/sentry/pull/26511

### DIFF
--- a/src/docs/environment/index.mdx
+++ b/src/docs/environment/index.mdx
@@ -110,11 +110,11 @@ exec "$SHELL"
 After this, if you type `which python`, you should see something like `$HOME/.pyenv/shims/python`
 rather than `/usr/bin/python`. This is because the following has been added to your start up script:
 
-```shell {filename: ~/.bash_profile} {tabTitle: Bash}
+```shell {tabTitle: Bash}
 Given that the bash instructions vary greatly based on the user's configuration, it is recommended to visit https://github.com/pyenv/pyenv#installation for instructions on how to set up Bash.
 ````
 
-```shell {filename: ~/.zshrc} {tabTitle: Zsh}
+```shell {filename: ~/.zprofile} {tabTitle: Zsh}
 # It is assumed that pyenv is installed via Brew, so this is all we need to do.
 eval "$(pyenv init --path)"
 ````
--- a/src/docs/environment/index.mdx
+++ b/src/docs/environment/index.mdx
@@ -213,7 +213,7 @@ And after doing that, reload your shell:
 exec "$SHELL"
 ````
 
-Any time the `.envrc` configuration changes (including the first time) you will be prompted to run `direnv allow` before any
+Any time the `.envrc` configuration changes (including the first load) you will be prompted to run `direnv allow` before any
 of the configuration will run. This is for security purposes and you are encouraged to inspect the changes before running this command.
 
 ### Customize your development environment variables
@@ -308,9 +308,9 @@ You might also be interested in <Link to="/continuous-integration/#troubleshooti
 
 ---
 
-**Problem:** something like `pkg_resources.DistributionNotFound: The 'some_dependency<0.6.0,>=0.5.5' distribution was not found and is required by sentry`
+**Problem:** You see an error that mentions something like `pkg_resources.DistributionNotFound: The 'some_dependency<0.6.0,>=0.5.5' distribution was not found and is required by sentry`
 
-**Solution:** Your virtualenv needs to be updated. run `make install-py-dev`.
+**Solution:** Your virtualenv needs to be updated. Run `make install-py-dev`.
 
 ---
 
@@ -330,9 +330,9 @@ make build-platform-assets
 
 ---
 
-**Problem:** `SessionNotCreatedException: Message: session not created: This version of ChromeDriver only supports Chrome version 76`
+**Problem:** You see `SessionNotCreatedException: Message: session not created: This version of ChromeDriver only supports Chrome version 76`
 
-Also, another symptom indicating you need to upgrade chromedriver:
+Or:
 
 ```shell
 Traceback (most recent call last):
@@ -346,6 +346,8 @@ RetryException: Could not successfully execute <functools.partial object at 0x10
 ```
 
 **Solution:**
+
+ChromeDriver needs to be updated.
 
 ```shell
 brew upgrade --cask chromedriver
@@ -383,9 +385,9 @@ pip install --no-cache-dir uwsgi
 
 ---
 
-**Problem:** `DoesNotExist: Subscription matching query does not exist`
+**Problem:** You see `DoesNotExist: Subscription matching query does not exist`
 
-**Solution:** In getsentry run the following to mock a subscription:
+**Solution:** In getsentry, run the following to mock a subscription:
 
 ```shell
 ./bin/mock-subscription <org_slug>
@@ -393,7 +395,7 @@ pip install --no-cache-dir uwsgi
 
 ---
 
-**Problem:** Something like `Error: No such container: sentry_postgres`
+**Problem:** You see something like `Error: No such container: sentry_postgres`
 
 **Solution:** Review the [bootstrap services](/environment/#bootstrap-services)
 section or spin up Docker services with:
--- a/src/docs/environment/index.mdx
+++ b/src/docs/environment/index.mdx
@@ -89,6 +89,17 @@ computer is actually compiling Python!
 make setup-pyenv
 ```
 
+fish users will need to manually set some of the environment variables. This only needs to be done
+once.
+
+```shell {tabTitle: Fish}
+set -Ux PYENV_ROOT $HOME/.pyenv
+# fish >=3.2.0
+fish_add_path $PYENV_ROOT/bin
+# fish <3.2.0
+set -U fish_user_paths $PYENV_ROOT/bin $fish_user_paths
+```
+
 Once that's done, your shell needs to be reloaded. You can either reload it in-place, or close your terminal and start it
 again and cd into sentry. To reload it, run:
 
@@ -105,15 +116,28 @@ exec $SHELL
 ```
 
 After this, if you type `which python`, you should see something like `$HOME/.pyenv/shims/python`
-rather than `/usr/bin/python` this is because the following has been added to your start up script:
+rather than `/usr/bin/python`. This is because the following has been added to your start up script:
 
-```shell
+
+```shell {filename: ~/.bash_profile} {tabTitle: Bash}
+eval "$(pyenv init --path)"
 if command -v pyenv 1>/dev/null 2>&1; then
   eval "$(pyenv init -)"
 fi
-```
+````
 
-Note: If you use Fish, add `source (pyenv init - | psub)` to `~/.config/fish/config.fish`.
+```shell {filename: ~/.zshrc} {tabTitle: Zsh}
+eval "$(pyenv init --path)"
+if command -v pyenv 1>/dev/null 2>&1; then
+  eval "$(pyenv init -)"
+fi
+````
+
+```shell {filename: ~/.config/fish/config.fish} {tabTitle: Fish}
+# pyenv init
+status is-login; and pyenv init --path | source
+pyenv init - | source
+````
 
 ### Virtual Environment
 
--- a/src/docs/environment/index.mdx
+++ b/src/docs/environment/index.mdx
@@ -168,17 +168,9 @@ curl https://get.volta.sh | bash
 The volta installer will tell you to "open a new terminal to start using Volta", but you don't have to! You can just
 reload your shell:
 
-```shell {tabTitle: Bash}
+```shell
 exec "$SHELL"
-```
-
-```shell {tabTitle: Zsh}
-exec "$SHELL"
-```
-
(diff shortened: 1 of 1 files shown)


src/docs/environment/index.mdx
```diff
@@ -213,16 +257,25 @@ And after doing that, reload your shell:
...
+
+- `SENTRY_SILENCE_DIRENV_NOTICE`: Make `direnv` output less noisy.
+- `SENTRY_DEVENV_NO_REPORT`: Do not report development environment errors to Sentry.io.
+- `SENTRY_DIRENV_DEBUG`: Print debug messages to stdout.
+
+A good starter `.env`:
+```shell {filename: .env}
+SENTRY_SILENCE_DIRENV_NOTICE=1
+SENTRY_DEVENV_NO_REPORT=1
+````
```
relaxolotl, line 278: Not all of the values supported in .env are documented anywhere, so this should serve as a
starting point in documenting these settings. An example .env has also been added to
improve ease of use. Just create the file, copy-paste the example, and you're set!

src/docs/environment/index.mdx
```diff
@@ -89,23 +89,55 @@ computer is actually compiling Python!
...
+exec "$SHELL"
+```
+
+```shell {tabTitle: Zsh}
 exec "$SHELL"
 ```
 
+```shell {tabTitle: Fish}
+exec $SHELL
+```
```
relaxolotl, line 116: This change was made in a few other additional places in the PR.

It's less code overall to just mention fish's differences in text, but it's also additional
required reading for the end user. Stuffing these into code blocks provides a more streamlined
experience since their selected languages sync across an entire page, meaning that once a user
decides to view the fish instructions in one code block, the others will switch to displaying
fish as well.

src/docs/environment/index.mdx
```diff
@@ -89,23 +89,55 @@ computer is actually compiling Python!
...
+````
+
+```shell {filename: ~/.zshrc} {tabTitle: Zsh}
+eval "$(pyenv init --path)"
+if command -v pyenv 1>/dev/null 2>&1; then
+  eval "$(pyenv init -)"
+fi
+````
 
-Note: If you use Fish, add `source (pyenv init - | psub)` to `~/.config/fish/config.fish`.
```
relaxolotl, line 108: See: pyenv's instructions in their [README](https://github.com/pyenv/pyenv#basic-github-checkout)

src/docs/environment/index.mdx
```diff
@@ -89,23 +89,55 @@ computer is actually compiling Python!
...
+once.
+
+```shell {tabTitle: Fish}
+set -Ux PYENV_ROOT $HOME/.pyenv
+# fish >=3.2.0
+fish_add_path $PYENV_ROOT/bin
+# fish <3.2.0
+set -U fish_user_paths $PYENV_ROOT/bin $fish_user_paths
+```
+
```
relaxolotl, line 102: It's possible to insert lines equivalent to the interactively run commands into the 
startup script instead, but it's difficult to get this right. A common fish mistake is to append the 
pyenv directory without any regard for whether it's already present in `PATH` or not, causing 
it to explode in length based on how often the user spins up new shells. 

In comparison, the fish commands proposed in this PR only need to be run once, will properly append to the `PATH`, and therefore should not cause too much developer pain.

src/docs/environment/index.mdx
```diff
@@ -213,16 +257,25 @@ And after doing that, reload your shell:
...
 ### Customize your development environment variables
 
 If you want to personalize your environment variables, you can do so by creating a `.env` file. This file is
 ignored by `git`, thus, you will not be able to leak it into one of your PRs.
 
-Two examples of variables that you can place there are `SENTRY_SILENCE_DIRENV_NOTICE` to make `direnv` output less noisy
-and `SENTRY_DEVENV_NO_REPORT=1` to prevent the development environment to report errors to Sentry.
+Some useful variables to set in `.env`:
+
+- `SENTRY_SILENCE_DIRENV_NOTICE`: Make `direnv` output less noisy.
```
relaxolotl, line 270: I actually can't seem to find any evidence of this variable's existence in sentry nor getsentry. Does this still exist?

src/docs/environment/index.mdx
```diff
@@ -89,23 +89,55 @@ computer is actually compiling Python!
...
+```
+
 After this, if you type `which python`, you should see something like `$HOME/.pyenv/shims/python`
-rather than `/usr/bin/python` this is because the following has been added to your start up script:
+rather than `/usr/bin/python`. This is because the following has been added to your start up script:
 
-```shell
+
+```shell {filename: ~/.bash_profile} {tabTitle: Bash}
+eval "$(pyenv init --path)"
```
armenzg, line 123: I read the bash instructions and I decided to leave it to the user to follow the instructions (besides most engineers don't use bash).
```suggestion
Visit https://github.com/pyenv/pyenv#installation for instructions on how to set up Bash.
```

src/docs/environment/index.mdx
```diff
@@ -89,23 +89,55 @@ computer is actually compiling Python!
...
 if command -v pyenv 1>/dev/null 2>&1; then
   eval "$(pyenv init -)"
 fi
-```
+````
+
+```shell {filename: ~/.zshrc} {tabTitle: Zsh}
+eval "$(pyenv init --path)"
```
armenzg, line 130:
```suggestion
# Since we have installed via Brew we only need to do this
echo 'eval "$(pyenv init --path)"' >> ~/.zprofile
```

src/docs/environment/index.mdx
```diff
@@ -89,23 +89,55 @@ computer is actually compiling Python!
...
+if command -v pyenv 1>/dev/null 2>&1; then
```
armenzg, line 131:
```suggestion
```

src/docs/environment/index.mdx
```diff
@@ -89,23 +89,55 @@ computer is actually compiling Python!
...
+  eval "$(pyenv init -)"
```
armenzg, line 132:
```suggestion
```

src/docs/environment/index.mdx
```diff
@@ -89,23 +89,55 @@ computer is actually compiling Python!
...
+fi
```
armenzg, line 133:
```suggestion
```

src/docs/environment/index.mdx
```diff
@@ -135,17 +175,24 @@ curl https://get.volta.sh | bash
...
+exec "$SHELL"
+```
+
+```shell {tabTitle: Zsh}
 exec "$SHELL"
 ```
 
+```shell {tabTitle: Fish}
+exec $SHELL
+```
```
armenzg, line 188: Are you sure this is better since the three are exactly the same?

I can see *some* value if they ever were to differ.

src/docs/environment/index.mdx
```diff
@@ -213,16 +257,25 @@ And after doing that, reload your shell:
...
 of the configuration will run. This is for security purposes and you are encouraged to inspect the changes before running this command.
 
 ### Customize your development environment variables
 
 If you want to personalize your environment variables, you can do so by creating a `.env` file. This file is
 ignored by `git`, thus, you will not be able to leak it into one of your PRs.
 
-Two examples of variables that you can place there are `SENTRY_SILENCE_DIRENV_NOTICE` to make `direnv` output less noisy
-and `SENTRY_DEVENV_NO_REPORT=1` to prevent the development environment to report errors to Sentry.
+Some useful variables to set in `.env`:
```
armenzg, line 268: We now have a make target that gives us the latest doc. Would you mind deleting the other lines and just keeping this one I suggest?

```suggestion
If you execute `make direnv-help` you will see the latest env variables to configure direnv to your taste.
```

src/docs/environment/index.mdx
armenzg: Maybe we can create a `.env.example` in Sentry and we slowly add all variables with their *default* values.
Or we even `cp .env.example .env` as part of the bootstrap script if the file does not exist.

What do you think?

src/docs/environment/index.mdx
armenzg: You're right. It's fallen out of date.

armenzg (changes requested):
Thank you for this PR and getting fish into a better state!

Few small changes, otherwise, it looks good!

src/docs/environment/index.mdx
relaxolotl: I like that idea! To build onto that, I'm thinking we could also omit the explanations for the variables by including descriptions for them in the `.env.example` itself, assuming it's possible to include comments there. If not, then I can keep this list.

Our suggested defaults appear to include all of the supported variables right now, so once a non-default variable is introduced we could begin exploring how we would like to document those in this repository. Thoughts?

src/docs/environment/index.mdx
relaxolotl: I'll remove this from the diff. 👍

src/docs/environment/index.mdx
armenzg: For optional aspects of the dev env, I would like to keep it directly via the CLI in Sentry. There's no easy way of keeping the develop docs in sync with what is happening in Sentry.

src/docs/environment/index.mdx
relaxolotl: I've opted to keep the second line as is (`eval "$(pyenv init --path)"`) since this is meant to represent what gets inserted into the user's `.zprofile`. It's subtle, but the filename portion of the codeblock should also provide context on where exactly this gets added.

src/docs/environment/index.mdx
```diff
@@ -213,16 +234,17 @@ And after doing that, reload your shell:
...
 
 ### Customize your development environment variables
 
 If you want to personalize your environment variables, you can do so by creating a `.env` file. This file is
-ignored by `git`, thus, you will not be able to leak it into one of your PRs.
+ignored by `git`, thus, you will not be able to leak it into one of your PRs. 
 
-Two examples of variables that you can place there are `SENTRY_SILENCE_DIRENV_NOTICE` to make `direnv` output less noisy
-and `SENTRY_DEVENV_NO_REPORT=1` to prevent the development environment to report errors to Sentry.
+One of `bootstrap`'s steps [creates a starter `.env`](https://github.com/getsentry/sentry/blob/master/.env.example) which should
```
relaxolotl, line 245: This link points to a nonexistent destination until https://github.com/getsentry/sentry/pull/26530 is merged in.

src/docs/environment/index.mdx
relaxolotl: I've taken slight liberties in updating this; Since we're [playing with the idea of having a template `.env` file](https://github.com/getsentry/sentry/pull/26530) I've made note of that in this PR. `make direnv-help` does get brought up, though.

relaxolotl:
Thank you again @armenzg for also reviewing this! I believe I've addressed all comments, and any slight changes are mentioned in replies to your comments. This should be good for a re-review.

src/docs/environment/index.mdx
```diff
@@ -89,23 +89,40 @@ computer is actually compiling Python!
...
+set -U fish_user_paths $PYENV_ROOT/bin $fish_user_paths
+```
+
 Once that's done, your shell needs to be reloaded. You can either reload it in-place, or close your terminal and start it
 again and cd into sentry. To reload it, run:
 
 ```shell
 exec "$SHELL"
-```
+````
```
armenzg, line 108: Do not make this change (3 single quotes is the right way).

src/docs/environment/index.mdx
```diff
@@ -89,23 +89,40 @@ computer is actually compiling Python!
...
+rather than `/usr/bin/python`. This is because the following has been added to your start up script:
 
-```shell
-if command -v pyenv 1>/dev/null 2>&1; then
-  eval "$(pyenv init -)"
-fi
-```
+```shell {tabTitle: Bash}
+Given that the bash instructions vary greatly based on the user's configuration, it is recommended to visit https://github.com/pyenv/pyenv#installation for instructions on how to set up Bash.
+````
```
armenzg, line 115: Fix 4 single quotes.

src/docs/environment/index.mdx
```diff
@@ -89,23 +89,40 @@ computer is actually compiling Python!
...
+
+```shell {filename: ~/.zprofile} {tabTitle: Zsh}
+# It is assumed that pyenv is installed via Brew, so this is all we need to do.
+eval "$(pyenv init --path)"
+````
```
armenzg, line 120: 4 single quotes.

src/docs/environment/index.mdx
```diff
@@ -137,15 +162,14 @@ reload your shell:
 
 ```shell
 exec "$SHELL"
-```
+````
```
armenzg, line 165: 4 single quotes.

armenzg (changes requested):
Some more minor changes.

src/docs/environment/index.mdx
relaxolotl: Thanks for catching this, looks like I copy-pasted a bad change to a couple of places 🙃

PR #352, opened 2021-06-10
Added "Redirect on update"

PetzJohannes:
`Redirect on update` has to be selected or the redirect back to sentry won't work.


leeandher:
hey @PetzJohannes thanks for the PR, but I didn't encounter any issues with Github Installation without that option selected when i was trying to reproduce what you mentioned. The installation redirect should work without that update option. If you are encountering a bug, please feel free to file an issue for it though 👍

PR #355, opened 2021-06-15
ref(self-hosted): Add upgrade paths w/ hard-stops

BYK:
This is in relation with getsentry/sentry#26567 and getsentry/onpremise#983


markstory (commented):
Looks good to me. Thanks for putting this together.

markstory (approved):

PR #364, opened 2021-06-29
Fix default session interval for mobile

marandaneto:


bruno-garcia:
Sounds like I applied my bad memory to some docs.
Thanks for fixing this

PR #366, opened 2021-06-29
feat: Document projectconfig versioning patterns for Relay

untitaker:


src/docs/services/relay/projectconfig-versioning.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+Relay fetches project configs from an "upstream". That upstream can either be Sentry or another Relay, both have their own implementation of the same endpoint.
+
+Your configuration will flow from Sentry to either:
+
+- Relay (freight deployment = `relay`)
+
+- External Relays (customers run those, we practically have no deprecation policy for old versions)
+
+  - At some point we want to serve those requests from Relay, such that only internal Relays can hit Sentry. This would limit complexity.
```
jan-auer, line 15: I think it wouldn't change much, since you still have to think about most if not all of these constraints. I'd remove this note.

src/docs/services/relay/projectconfig-versioning.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+Likewise *either Sentry or (internal) Relay* can forward your configuration to:
```
jan-auer, line 17: This is wrong. PoPs only talk to Relay. The way this is designed and soon implemented the chain is always:
Customer -> PoP -> Processing Relay -> Sentry

src/docs/services/relay/projectconfig-versioning.mdx
```diff
@@ -0,0 +1,138 @@
...
+## Adding Top-Level Fields
+
+In case you're developing a new feature, you likely want to add completely new data to the project config and not touch existing structs. Here's a small checklist that applies in addition to everything else:
+
+- Verify that the field is in `camelCase`.
+- Ensure the field name has not been used in a past version of Relay. `git log -G field_name` is a simple way to search git history for strings that occur in diffs.
+- Make sure your field is either optional or has a default
+- Do not serialize when the field is `None` or default via `serde(skip_serializing_if)`.
+- Use `ErrorBoundary` around the field type for as long as your new feature is pre-GA and can change its definition. This makes sure that if the field definition is malformed, Relay contains the error and continues to parse the rest of the project config.
+- Do not use `usize`.
```
jan-auer, line 84:
```suggestion
- Do not use types with platform specific behavior, such as `usize`.
```

Not sure if there are any other types we'd ever consider adding there, but it's better to show clear intent.

src/docs/services/relay/projectconfig-versioning.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+- As demonstrated above, changing types is *not* safe. See below for instructions on changing types.
+- Ensure the field name has never been used before. It could have been implemented with different types, or trigger unwanted behavior in prior versions of Relay.
+- Note that adding enum variants requires a different approach.
+
+## Renaming Fields
+
+Do not rename a field for features that are generally available for external Relays. Renaming a field makes the field invisible to older Relays. This makes it infeasible to rename fields for aesthetic reasons (it *may* be fine if your feature is pre-EA), but for the same reason [it is a useful mechanism for versioning](#breaking-changes-and-actual-versioning)
+
+## Changing Field Types
```
jan-auer, line 111: I think we need to revise this section:

1. Let's lead with external Relays, because this is the critical case. You can then lower the constraints for internal Relays. 
2. Type changes are only permitted when they have exactly identical serialization behavior. For example, changing `String` to `Cow<str>` would be allowed. There are no other exceptions, everything else is forbidden.

Types cannot be widened, not even with untagged enums. This is because Relay would then forward the new value to a downstream that doesn't support it yet. 

Types cannot be made stricter, because then Relay might reject project configs from an older onpremise version.

src/docs/services/relay/projectconfig-versioning.mdx
```diff
@@ -0,0 +1,138 @@
...
+- [Use an untagged enum](https://serde.rs/enum-representations.html#untagged) (this is like `Union[str, int]` in Python)
+- Write a custom `Deserialize` impl for your type to be able to deserialize from older values
+
+## Deployment & Monitoring
+
+Talk to the ingest team for deploying and monitoring.
+
+## Breaking changes, and actual versioning
+
+> `myfield_v2_final_final.docx`
```
jan-auer, line 132: What does this mean?

src/docs/services/relay/projectconfig-versioning.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+We've mostly gone through things that we don't want you to do, as they will cause incidents.
+
+In case you do need to make larger changes to the schema, or you want to do a large amount of smaller changes and feel like guaranteeing backwards compat for all of them is too hard, consider renaming the top-level field `x` to `x_v2`.
```
jan-auer, line 136: There is another approach to version the project config format. See https://github.com/getsentry/relay/blob/3933cd6ccd59c3270f56853ffbc8678b219e44e6/relay-server/src/actors/project_upstream.rs#L25-L34

jan-auer (changes requested):
Thanks, looks good. A few remarks, mostly on renaming, then this is G2G.

src/docs/services/relay/projectconfig-versioning.mdx
untitaker: stupid joke, I removed

src/docs/services/relay/projectconfig-versioning.mdx
untitaker: >There are no other exceptions, everything else is forbidden.

I think enlarging integer types is probably safe most of the time but I just added a general disclaimer that this needs to be discussed with the team

Applied the rest of the feedback

src/docs/services/relay/projectconfig-versioning.mdx
untitaker: I now mention it as sidenote for *us* to investigate but I haven't thought enough about this to know when to use it. So I'd not write anything to steer people towards it

jan-auer (approved):

PR #367, opened 2021-06-29
docs(envelope): Add note regarding envelope header authentication

kamilogorek:
@jan-auer can you verify please?

Ref https://github.com/getsentry/sentry-javascript/issues/3762

### DIFF
--- a/src/docs/sdk/envelopes.mdx
+++ b/src/docs/sdk/envelopes.mdx
@@ -512,6 +512,8 @@ Envelope header authentication requires **Relay v21.6.0**.
 Earlier versions of Relay do not support Envelope header authentication and respond with HTTP error
 _401 Unauthorized ("missing authorization information")_ to envelope uploads.
 
+SDKs should not rely on Envelope header authentication to retain backward compatibility with older versions of Sentry on-premise unless absolutely required. Instead, stick to HTTP headers or query parameters wherever possible.
+
 </Alert>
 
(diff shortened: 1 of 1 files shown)


jan-auer (approved):
Thanks, I'd suggest adding some more advice:

> SDKs should not rely on Envelope header authentication to retain backwards compatibility with older versions of Sentry onpremise, unless absolutely required. Instead, stick to HTTP headers or query parameters wherever possible.

src/docs/sdk/envelopes.mdx
```diff
@@ -506,6 +506,14 @@ If multiple forms of authentication are given, the endpoint validates that the
 information matches and otherwise rejects the request. If both are missing, the
 Envelope is rejected with status code `403 Forbidden`.
 
+<Alert title="Backward Compatibility" level="info">
+
+Envelope header authentication requires **Relay v21.6.0**.
+Earlier versions of Relay do not support Envelope header authentication and respond with HTTP error
+_401 Unauthorized ("missing authorization information")_ to envelope uploads.
```
kamilogorek, line 513:
```suggestion
_401 Unauthorized ("missing authorization information")_ to envelope uploads.

SDKs should not rely on Envelope header authentication to retain backward compatibility with older versions of Sentry on-premise unless absolutely required. Instead, stick to HTTP headers or query parameters wherever possible.
```

PR #370, opened 2021-07-03
Super Nuclear Option for Zookeeper

felix-engelmann:
Adding a super nuclear option to delete Zookeeper volumes too for recovery.


BYK:
I think #372 superseded this?

felix-engelmann:
Sure, #372 is way better.

PR #356, opened 2021-06-21
Document tracing SDK API evolution

rhcarvalho:
Creating a space to summarize the evolution of the tracing SDK APIs and where we go next.

This is the first part. Will continue on follow up PRs for ease of reviewing.

PR #381, opened 2021-07-19
Document tracing issues related to scope propagation

rhcarvalho:
Follows up on #356.

### DIFF
--- a/src/docs/sdk/research/performance/index.mdx
+++ b/src/docs/sdk/research/performance/index.mdx
@@ -138,6 +138,14 @@ t2
 
 As a side-effect of not being able to correctly determine the current span, the present implementation of the `fetch` integration (and others) in [the JavaScript Browser SDK chooses to create flat transactions](https://github.com/getsentry/sentry-javascript/blob/61eda62ed5df5654f93e34a4848fc9ae3fcac0f7/packages/tracing/src/browser/request.ts#L169-L178), where all child spans are direct children of the transaction (instead of having a proper multi-level tree structure).
 
+Note that other tracing libraries have the same kind of challenge. There are several (at the time open) issues in OpenTelemetry for JavaScript related to determining the parent span and proper context propagation (including async code):
+
+- [Context leak if several TracerProvider instances are used #1932](https://github.com/open-telemetry/opentelemetry-js/issues/1932)
+- [How to created nested spans without passing parents around #1963](https://github.com/open-telemetry/opentelemetry-js/issues/1963)
+- [Nested Child spans are not getting parented correctly #1940](https://github.com/open-telemetry/opentelemetry-js/issues/1940)
+- [OpenTracing shim doesn't change context #2016](https://github.com/open-telemetry/opentelemetry-js/issues/2016)
+- [Http Spans are not linked / does not set parent span #2333](https://github.com/open-telemetry/opentelemetry-js/issues/2333)
+
 ### 2. Conflicting Data Propagation Expectations
 
 There is a conflict of expectations that appear whenever we add a `trace` function as discussed earlier, or simply try to address scope propagation with Zones.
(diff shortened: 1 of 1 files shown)


HazAT (approved):

src/docs/sdk/research/performance/index.mdx
```diff
@@ -24,4 +24,124 @@ In the next section, we’ll discuss some of the shortcomings with the current m
...
+
+JavaScript, for example, is single-threaded with an event loop and async code execution. There is no standard way to carry contextual data that works across async calls. So for JavaScript browser applications, there is only one global `hub` shared for sync and async code.
+
+A similar situation appears on Mobile SDKs. There is an user expectation that contextual data like tags, what the current user is, breadcrumbs, and other information stored on the `scope` to be available and settable from any thread. Therefore, in those SDKs there is only one global `hub`.
+
+In both cases, everything was relatively fine when the SDK had to deal with reporting errors. With the added responsibility to track transactions and spans, the `scope` became a poor fit to store the current `span`, because it limits the existence of concurrent spans.
+
+For JavaScript, a possible solution is the use of [Zone.js](https://github.com/angular/angular/blob/master/packages/zone.js/README.md), part of the Angular framework. The main challenge is that it increases bundle size and may inadvertendly impact end user apps as it monkey-patches key parts of the JavaScript runtime engine.
```
AbhiPrasad, line 47:
```suggestion
For Browser JavaScript, a possible solution is the use of [Zone.js](https://github.com/angular/angular/blob/master/packages/zone.js/README.md), part of the Angular framework. The main challenge is that it increases bundle size and may inadvertendly impact end user apps as it monkey-patches key parts of the JavaScript runtime engine.
```

src/docs/sdk/research/performance/index.mdx
```diff
@@ -24,4 +24,124 @@ In the next section, we’ll discuss some of the shortcomings with the current m
...
+await Sentry.trace(
+  {
+    op: 'dom',
+    description: 'Walk DOM Tree',
+  },
+  async () => await walkDomTree()
+);
+```
+
+Users wouldn’t have to worry about keeping the reference to the correct transaction or span when adding timing data. Users are free to create child spans within the `walkDomTree` function and spans will be ordered in the correct hierarchy.
```
AbhiPrasad, line 63:
```suggestion
With the `Sentry.trace` function, users wouldn’t have to worry about keeping the reference to the correct transaction or span when adding timing data. Users are free to create child spans within the `walkDomTree` function and spans would be ordered in the correct hierarchy.
```

src/docs/sdk/research/performance/index.mdx
```diff
@@ -24,4 +24,124 @@ In the next section, we’ll discuss some of the shortcomings with the current m
...
+
+The implementation of the actual `trace` function is relatively simple (see [a PR which has an example implementation](https://github.com/getsentry/sentry-javascript/pull/3697/files#diff-f5bf6e0cdf7709e5675fcdc3b4ff254dd68f3c9d1a399c8751e0fa1846fa85dbR158)), however, knowing what is the current span in async code and global integrations is a challenge yet to be overcome.
+
+Here are two examples that summarize the problems above:
```
AbhiPrasad, line 67: Think we can `rm -rf` this because we already introduced the two classes of problems in the introduction.

```suggestion
```

rhcarvalho: There is a double two here... this sentence refer to two examples of scope propagation problems, the introduction talked about two higher level classes of problems: scope propagation and the span ingestion model.

AbhiPrasad: Ah I failed to capture that, that's probably on me.

rhcarvalho: I've reworded this sentence so the reader doesn't need to keep track of what "the problems above" refer to.

HazAT (approved):

PR #390, opened 2021-07-27
Specify 21.6.1 as stopper instead of 21.6.3

mawalu:
I encountered this problem while upgrading. As fare as I understand the upgrade path `21.6.1` is the last version that still can migrate off `South`. See [21.6.1](https://github.com/getsentry/sentry/blob/21.6.1/src/sentry/runner/commands/upgrade.py) vs [21.6.2](https://github.com/getsentry/sentry/blob/21.6.2/src/sentry/runner/commands/upgrade.py). [e062636](https://github.com/getsentry/sentry/commit/e06263660e6ddf458a3c8c4d2014af5f14552a17) also mentions `21.6.1` as stopper in the changelog.


markstory:
Thank you 👍

PR #394, opened 2021-07-30
Unity context added

semuserable:


semuserable:
cc @bruno-garcia @bitsandfoxes

bruno-garcia:
To be honest, since this specific to one platform, I'm not sure we should add it here

AbhiPrasad:
If we treat it differently in the product we should add it (there is django or relay logic around it), otherwise I'm fine with leaving it out for now (and just making sure it's noted somewhere in the unity git repo). As an example, the React SDK sets context under the `react` key, but we didn't document in develop/docs.

bruno-garcia:
There's no Sentry special support for it. It's just a dynamic context like a user-defined one.
Lets close for now

PR #398, opened 2021-08-05
fix: Update the docs to reflect that 21.6 is required

markstory:
Because the squashed migrations were removed in 21.7 folks need to stop at 21.6 before going forward.


markstory:
Replaced by #405

PR #404, opened 2021-08-16
feat(api): Add API checklist and API concepts

manuzope:
This PR does the following:
## Move API documentation instructions to it's own section in the sidebar
It is currently nested under "Documentation Guide" and it's not easily findable.

## Add a public API checklist
This has been moved from our internal wiki to the develop site. This was discussed and agreed upon during the last backend TSC. Since what was presented in backend TSC:
1. I've removed which status codes we respond with cause it was simply listing out the HTTP spec
2. Have not added guidelines on error objects - this is still something we should do but I do not want to block these guidelines while we figure that out
3. Made the list numbered

## Add API concepts
This has also been moved from our internal wiki. I've added `collapse` to the guidelines.

## Notes to the reviewer
Please click on the Vercel link to see what this will look like on production.

### DIFF
--- a/src/docs/api/concepts.mdx
+++ b/src/docs/api/concepts.mdx
@@ -1,5 +1,5 @@
 ---
-title: "API concepts"
+title: "API Concepts"
 ---
 In this document, we will be looking at API concepts that exist and should be followed by endpoints. We also describe why these concepts exist so that developers can use them at their own discretion.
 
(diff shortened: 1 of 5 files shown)


src/docs/api/concepts.mdx
```diff
@@ -0,0 +1,96 @@
+---
+title: "API concepts"
```
ceorourke, line 2: Can you capitalize "concepts" here?

ceorourke (approved):
👍

PR #407, opened 2021-08-20
feat(sdk): Describe trace contexts

jan-auer:
Rendered: https://develop-git-feat-trace-context.sentry.dev/sdk/trace-context/

### DIFF
--- a/src/components/sidebar.tsx
+++ b/src/components/sidebar.tsx
@@ -203,7 +203,9 @@ export default () => {
           <SidebarLink to="/sdk/store/">Store Endpoint</SidebarLink>
           <SidebarLink to="/sdk/envelopes/">Envelopes</SidebarLink>
           <SidebarLink to="/sdk/rate-limiting/">Rate Limiting</SidebarLink>
-          <SidebarLink to="/sdk/performance/">Performance</SidebarLink>
+          <SidebarLink to="/sdk/performance/" title="Performance">
+            <SidebarLink to="/sdk/trace-context/">Trace Contexts</SidebarLink>
+          </SidebarLink>
           <SidebarLink to="/sdk/event-payloads/" title="Event Payloads">
             <SidebarLink to="/sdk/event-payloads/transaction/">
               Transaction Type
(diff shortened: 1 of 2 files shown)


src/components/sidebar.tsx
```diff
@@ -203,7 +203,9 @@ export default () => {
           <SidebarLink to="/sdk/store/">Store Endpoint</SidebarLink>
           <SidebarLink to="/sdk/envelopes/">Envelopes</SidebarLink>
           <SidebarLink to="/sdk/rate-limiting/">Rate Limiting</SidebarLink>
-          <SidebarLink to="/sdk/performance/">Performance</SidebarLink>
+          <SidebarLink to="/sdk/performance/" title="Performance">
+            <SidebarLink to="/sdk/trace-context/">Trace Contexts</SidebarLink>
```
rhcarvalho, line 207: Any reason to have  the URL path `/sdk/trace-context/` instead of nesting under `/sdk/performance/trace-context/`? The navigation is not working as expected, once you click the link the new page Trace Contexts has no corresponding item in the sidebar visible.

src/docs/sdk/trace-context.mdx
```diff
@@ -0,0 +1,228 @@
+---
+title: "Trace Contexts (Experimental)"
```
rhcarvalho, line 2: This name is very unfortunate because SDKs already have a "Trace Context", an object that is sent in events (both errors and transactions) as `event.contexts.trace`.

src/components/sidebar.tsx
jan-auer: I didn't care much about the URL and didn't want to create folders, to be honest. We can move it if you like.

src/docs/sdk/trace-context.mdx
jan-auer: Also for @RaduW -- suggestions on how we name the new trace context, instead? How about "trace state" borrowing from W3C?

RaduW: Trace state works for me

PR #411, opened 2021-08-25
add 21.5.0 to upgrade hard stops

Sechmann:
it's required to get the south -> django migration


markstory (approved):

chadwhitacre (changes requested):
This is insufficient imo, I think we need a third bullet in between the existing two. No?

PR #412, opened 2021-08-25
ref(selfhosted) Add missing hard stop

markstory:
Users coming from sentry 9.x need to stop at 21.5.0 as it is the last release that contains south upgrade shims. They will also need to stop at 21.6.3 to run the last of the django migrations that were squashed.


markstory:
Duplicate of #411

PR #420, opened 2021-08-30
fix(tracing): Small fixes to tracestate docs about third-party data

lobsterkatie:
This fixes two problems with the docs on how to handle third-party `tracestate` data:

- In the "joined by comma" example in the Tracestate Headers section, there shouldn't be a space between the entries. See https://www.w3.org/TR/trace-context/#combined-header-value.

- In the last snippet in that same section, with an example header containing both sentry and third-party data, the order needs to be reversed, since whatever key-value pair you're adding or modifying should end up the leftmost one. See https://www.w3.org/TR/trace-context/#mutating-the-tracestate-field.

### DIFF
--- a/src/docs/sdk/performance/trace-context.mdx
+++ b/src/docs/sdk/performance/trace-context.mdx
@@ -72,7 +72,7 @@ Tracestate headers contain several vendor-specific opaque data. As per HTTP spec
 
 - Joined by comma:
   ```
-  tracestate: sentry=<data>, other=<data>
+  tracestate: sentry=<data>,other=<data>
   ```
 - Repetition:
   ```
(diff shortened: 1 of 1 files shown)


src/docs/sdk/performance/trace-context.mdx
```diff
@@ -119,7 +119,7 @@ ewogICJ0cmFjZV9pZCI6ICI3NzFhNDNhNDE5MjY0MmYwYjEzNmQ1MTU5YTUwMTcwMCIsCiAgInB1Ymxp
 Resulting in the header:
 
 ```
-tracestate: other=[omitted],sentry=ewogICJ0cmFjZV9pZCI6ICI3NzFhNDNhNDE5MjY0MmYwYjEzNmQ1MTU5YTUwMTcwMCIsCiAgInB1YmxpY19rZXkiOiAiNDlkMGY3Mzg2YWQ2NDU4NThhZTg1MDIwZTM5M2JlZjMiLAogICJyZWxlYXNlIjogIjEuMS4yMiIsCiAgImVudmlyb25tZW50IjogImRldiIsCiAgInVzZXIiOiB7CiAgICAic2VnbWVudCI6ICJ2aXAiLAogICAgImlkIjogIjdlZmE0OTc4ZGExNzc3MTNkZjA4OGY4NDZmOGM0ODRkIgogIH0KfQ
+tracestate: sentry=ewogICJ0cmFjZV9pZCI6ICI3NzFhNDNhNDE5MjY0MmYwYjEzNmQ1MTU5YTUwMTcwMCIsCiAgInB1YmxpY19rZXkiOiAiNDlkMGY3Mzg2YWQ2NDU4NThhZTg1MDIwZTM5M2JlZjMiLAogICJyZWxlYXNlIjogIjEuMS4yMiIsCiAgImVudmlyb25tZW50IjogImRldiIsCiAgInVzZXIiOiB7CiAgICAic2VnbWVudCI6ICJ2aXAiLAogICAgImlkIjogIjdlZmE0OTc4ZGExNzc3MTNkZjA4OGY4NDZmOGM0ODRkIgogIH0KfQ,other=[omitted]
```
jan-auer, line 122: I put other in front so that it's easy to spot. It's good having this in the right order though. 

What if we add a note about this above, something like:

> Resulting in the following header. Note how other the header contains other trace state at the end:

jan-auer (approved):

PR #423, opened 2021-09-01
Add status to span payload and reorder fields logically

marandaneto:
Add status to the example as it is one of the possible fields.
Reorder fields in some logical way for readability, most important and required fields first.


rhcarvalho (approved):
Thanks @marandaneto. I took the opportunity to check all other fields -- now the first span in the example has all possible span fields, the second span has only some (as many of the fields are optional).

PR #427, opened 2021-09-02
ref(review): Add Jira link to pull request names

MikeRumplerSentry:
To track changes made back ti issues in Jira (User Stories, Customer Issues, ...) add the Jira ID to the commit message. This shows braches, commits, pull-requests and their status in Jira. Example here: https://getsentry.atlassian.net/browse/INGEST-293

### DIFF
--- a/src/docs/code-review.mdx
+++ b/src/docs/code-review.mdx
@@ -74,7 +74,7 @@ The header has a special format that includes a type, a scope and a subject:
 <footer>
 ```
 
-The **header** is mandatory and the **scope** of the header is optional. If you have a Jira issue to link to, add the **jira-id** the commit belongs to. This helps to track changes back to Jira tickets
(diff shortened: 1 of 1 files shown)


jan-auer (approved):
Works for me. Only concern is that we're putting this in a rather central place and further eating from the space granted to the subject.

To confirm: There's no other way to get this done from Jira's end?

MikeRumplerSentry:
> Works for me. Only concern is that we're putting this in a rather central place and further eating from the space granted to the subject.
> 
> To confirm: There's no other way to get this done from Jira's end?

For the time being there is no easier option. We could check if the [<jira id>] helps as GitHub may render this also in the subject....

jan-auer:
Oh, one more thing. Together with "autolink references", GitHub can even make those short-IDs a link directly into JIRA. We've enabled them in a bunch of repository settings like this:

image

PR #428, opened 2021-09-03
ref: Clarify Trace Context SDK Guidelines

philipphofmann:
The implementation guidelines for the trace context were a bit
misleading. On Java, for example, the first implementation only added the envelope
header for transactions bound to the scope.
This is fixed now by changing the wording a bit.


jan-auer (approved):

marandaneto:
the commit has a comment:

> On Java, for example, the first implementation only added the envelope
> header for transactions bound to the scope.

I don't think this is the case though, can you elaborate @philipphofmann ?
cc @maciejwalkowiak

bruno-garcia (approved):

PR #433, opened 2021-09-15
docs(envelopes): Update envelope header requirement

kamilogorek:
* add a note regarding envelope header being required
* remove an example without envelope header
* fix title of one of the examples

### DIFF
--- a/src/docs/sdk/envelopes.mdx
+++ b/src/docs/sdk/envelopes.mdx
@@ -161,7 +161,7 @@ about the decision.
 
 Notes for implementors:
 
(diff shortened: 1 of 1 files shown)


src/docs/sdk/envelopes.mdx
```diff
@@ -161,6 +161,7 @@ about the decision.
 
 Notes for implementors:
 
+- Envelope header is **required**, but it can be empty
```
kamilogorek, line 164:
```suggestion
- Envelope header is **required**, but it can be empty.
```

mitsuhiko (approved):

PR #448, opened 2021-10-05
docs/unified-api: Fix wording issue

Turbo87:
I'm no native speaker, but "it also setups all default integrations" felt wrong when I just read it in the docs. This PR replaces it with:

> it also sets up all default integrations


dcramer (approved):
Thanks!

Issue #392, opened 2021-07-28
Document spec around span operations


AbhiPrasad:
`op` is part of our [Span spec](https://develop.sentry.dev/sdk/event-payloads/span/#attributes), but we don't have a list of used ops anywhere.

In the section where we detail [HTTP Client features](https://develop.sentry.dev/sdk/features/#http-client-integrations) we mention the operation `http.client`, but we don't further elaborate.

In addition with operations breakdown, the product now depends on op values. We should make that clear in the develop docs. We default breakdowns to use [`http`, `db`, `resource`, and `browser` operations](https://github.com/getsentry/sentry/blob/809b7fe54c6f06cc1e4c503cf83ded896472a011/src/sentry/projectoptions/defaults.py#L74), and relay has [logic for reading/using the operation name](https://github.com/getsentry/relay/blob/9fd1a4310ea7f18a09df9f3a772339a546d6e422/relay-general/src/store/normalize/breakdowns.rs#L148)

This issue serves two purposes. First to start formalizing a list of ops that should be used by the sdks and second to track the inclusion of that list in the develop docs somewhere.

marandaneto:
@AbhiPrasad maybe this can be closed now?

AbhiPrasad:
I was going to close it after I sent a #shipped email - but yes it’s in a finished state.

AbhiPrasad:
Closed with https://vanguard.getsentry.net/p/cl8swagig15050ls6tedzqax7 (internal to Sentry employees)

PR #452, opened 2021-10-25
feat(perf): Document span operations spec

AbhiPrasad:
Edit: Please note, the scope of this PR has changed since it's creation. After reading the PR description, please see: https://github.com/getsentry/develop/pull/452#issuecomment-983716185

This PR adds documentation around the span operations spec that we use.

# Background

Before reviewing, I would recommend reading through the `Introduction` of this document: [https://develop.sentry.dev/sdk/research/performance#introduction](https://develop.sentry.dev/sdk/research/performance#introduction)

`OpenTracing`, which we based our initial performance product off of, had a concept of operation name. We can see it being leveraged in this [early JS SDK PR](https://github.com/getsentry/sentry-javascript/pull/1918/files#diff-b7d385e8e3c69cd4ade2edd41436b21255dde1d851717dc5c22cc7623a911a37R61).

According to OpenTracing (which became OpenTelemetry btw): [https://opentracing.io/specification/](https://opentracing.io/specification/)

As we moved forward, we introduced [both the operation and description fields to spans](https://github.com/getsentry/sentry-python/pull/372/files#diff-771cb012ea65fcb826b77e2abb36b504fdf3407fbe8054857b05fe5a41ed5164R80). Now, these seem redundant (and in fact OpenTelemetry just has [span name](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span)) - but if we use them correctly, they can be an advantage in our spec.

image

From the current usage of these attributes, span operation seems to typically be a low cardinality value, a short code that provides instant understanding to the user when they see a span. For example, `db` or `http`. Descriptions tend to be way more high cardinality, so a `db` operation would have the SQL query as it's description, while a `http` operation would have the resource URL as it's description.

Although these conventions were never codified or formally specced out, they seem to be providing a high degree of customer value. As such, when the span operations spec was created, it was heavily based on the existing conventions.

# What was done

Please read through `src/docs/sdk/performance/span-operations.mdx`.

In short, we introduce the ideas of categories and actions when defining span operations. Operations are expected to generally take the form: `CATEGORY.ACTION`, where the categories and actions can have optional identifiers.

# Next Steps

For now, this PR heavily focuses on defining the span operations. As a next step, we may start taking a look at [span descriptions](#463) or other metadata. Eventually, we want to establish a set of semantic conventions around performance monitoring, similar to OpenTelemetry's. In this case, we probably want to rely on OpenTelemetry's semantic conventions for backend focused use cases. For areas like desktop, mobile and browser (where OpenTelemetry lacks details), we should be more liberal in setting standards.

---

For #392.

### DIFF
--- a/src/docs/sdk/event-payloads/properties/op.mdx
+++ b/src/docs/sdk/event-payloads/properties/op.mdx
@@ -6,6 +6,6 @@ For more details, see <Link to="/sdk/performance/span-operations/">Sentry's conv
 
   ```json
   {
-    "op": "sql.query"
+    "op": "db.query"
   }
   ```
--- a/src/docs/sdk/event-payloads/properties/op.mdx
+++ b/src/docs/sdk/event-payloads/properties/op.mdx
@@ -2,6 +2,8 @@
 
 : _Recommended_. Short code identifying the type of operation the span is measuring.
 
+For more details, see <Link to="/sdk/performance/span-operations/">Sentry's conventions around span operations</Link>.
+
   ```json
   {
     "op": "sql.query"
--- a/src/docs/sdk/performance/span-operations.mdx
+++ b/src/docs/sdk/performance/span-operations.mdx
@@ -18,38 +18,40 @@ Having both a category and action is recommended. Operations should be lowercase
 
 ## Browser
 
-| Op         | Usage           | Description                                                                                                         |
-| ---------- | --------------- | ------------------------------------------------------------------------------------------------------------------- |
-| pageload   |                 | A full page load of a web application                                                                               |
-| navigation |                 | Client-side browser history change in a web application                                                             |
-| resource   |                 | Resource as per [Performance Resource Timing](https://w3c.github.io/resource-timing/#sec-performanceresourcetiming) |
-|            | resource.script |                                                                                                                     |
-|            | resource.link   |                                                                                                                     |
-|            | resource.css    |                                                                                                                     |
-|            | resource.img    |                                                                                                                     |
-| browser    |                 | Usage of browser APIs or functionality                                                                              |
-|            | browser.paint   |                                                                                                                     |
-| mark       |                 | Usage of [performance.mark() API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark)                |
-| measure    |                 | Usage of [performance.measure() API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure)          |
-| user       |                 |                                                                                                                     |
-| animate    |                 |                                                                                                                     |
+| Category    | Usage           | Description                                                                                                         |
+| ----------- | --------------- | ------------------------------------------------------------------------------------------------------------------- |
+| pageload    |                 | A full page load of a web application                                                                               |
+| navigation  |                 | Client-side browser history change in a web application                                                             |
+| resource    |                 | Resource as per [Performance Resource Timing](https://w3c.github.io/resource-timing/#sec-performanceresourcetiming) |
+|             | resource.script |                                                                                                                     |
+|             | resource.link   |                                                                                                                     |
+|             | resource.css    |                                                                                                                     |
+|             | resource.img    |                                                                                                                     |
+| browser     |                 | Usage of browser APIs or functionality                                                                              |
+|             | browser.paint   |                                                                                                                     |
+| mark        |                 | Usage of [performance.mark() API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark)                |
+| measure     |                 | Usage of [performance.measure() API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure)          |
+| interaction |                 | A user interaction                                                                                                  |
+| animate     |                 | An animation                                                                                                        |
+| render      |                 | Time it takes to render an element                                                                                  |
+| element     |                 | Operations related to [Web Components](https://developer.mozilla.org/en-US/docs/Web/Web_Components)                 |
 
 ### JS Frameworks
 
-| Op      | Usage        | Description                                    |
-| ------- | ------------ | ---------------------------------------------- |
-| react   |              | Spans related to [React](https://reactjs.org/) |
-|         | react.mount  |                                                |
-|         | react.render |                                                |
-| vue     |              |                                                |
-|         | vue.mount    |                                                |
-|         | vue.update   |                                                |
-| angular |              |                                                |
-| ember   |              |                                                |
+| Category | Usage        | Description                                      |
+| -------- | ------------ | ------------------------------------------------ |
+| react    |              | Spans related to [React](https://reactjs.org/)   |
+|          | react.mount  |                                                  |
+|          | react.render |                                                  |
+| vue      |              | Spans related to [Vue.js](https://vuejs.org/)    |
+|          | vue.mount    |                                                  |
+|          | vue.update   |                                                  |
+| angular  |              | Spans related to [Angular](https://angular.io/)  |
+| ember    |              | Spans related to [EmberJS](https://emberjs.com/) |
 
 ## Web Server
 
-| Op        | Usage           | Description |
+| Category  | Usage           | Description |
 | --------- | --------------- | ----------- |
 | http      |                 |             |
 |           | http.client     |             |
@@ -60,66 +62,69 @@ Having both a category and action is recommended. Operations should be lowercase
 |           | template.init   |             |
 |           | template.parse  |             |
 |           | template.render |             |
-| view      |                 |             |
+| render    |                 |             |
 |           | view.render     |             |
 | serialize |                 |             |
+| console   |                 |             |
 
 ## Web Frameworks
 
-| Op      | Usage             | Description |
-| ------- | ----------------- | ----------- |
-| django  |                   |             |
-|         | django.middleware |             |
-|         | django.view       |             |
-| express |                   |             |
-| rails   |                   |             |
-| rack    |                   |             |
-| console |                   |             |
+| Category | Usage             | Description |
+| -------- | ----------------- | ----------- |
+| django   |                   |             |
+|          | django.middleware |             |
+|          | django.view       |             |
+| express  |                   |             |
+| rails    |                   |             |
+| rack     |                   |             |
 
 ## Database
 
-| Op  | Usage            | Description                |
-| --- | ---------------- | -------------------------- |
-| db  |                  | An operation on a database |
-|     | db.connection    |                            |
-|     | db.transaction   |                            |
-|     | db.sql.query     |                            |
-|     | db.query         |                            |
-|     | db.query.compile |                            |
+| Category | Usage            | Description                |
+| -------- | ---------------- | -------------------------- |
+| db       |                  | An operation on a database |
+|          | db.connection    |                            |
+|          | db.transaction   |                            |
+|          | db.sql.query     |                            |
+|          | db.query         |                            |
+|          | db.query.compile |                            |
 
 ## Serverless (FAAS)
 
-| Op  | Usage       | Description |
-| --- | ----------- | ----------- |
-| aws |             |             |
-|     | aws.lambda  |             |
-|     | aws.request |             |
-| gcp |             |             |
+| Category | Usage            | Description |
+| -------- | ---------------- | ----------- |
+| faas     |                  |             |
+|          | faas.aws         |             |
+|          | faas.aws.lambda  |             |
+|          | faas.aws.request |             |
+|          | faas.gcp         |             |
 
 ## Mobile
 
-| Op         | Usage      | Description                         |
-| ---------- | ---------- | ----------------------------------- |
-| app        |            | Data about the mobile app           |
-|            | app.start  |                                     |
-| ui         |            | An operation on a mobile/desktop UI |
-|            | ui.load    |                                     |
-|            | ui.action  |                                     |
-| navigation |            | Navigating to another screen        |
-| file       |            | Operations on the file system       |
-|            | file.read  |                                     |
-|            | file.write |                                     |
+| Category   | Usage          | Description                         |
+| ---------- | -------------- | ----------------------------------- |
+| app        |                | Data about the mobile app           |
+|            | app.start      |                                     |
+|            | app.start.warm |                                     |
+|            | app.start.cold |                                     |
+| ui         |                | An operation on a mobile/desktop UI |
+|            | ui.load        |                                     |
+|            | ui.action      |                                     |
+| navigation |                | Navigating to another screen        |
+| file       |                | Operations on the file system       |
+|            | file.read      |                                     |
+|            | file.write     |                                     |
 
 ## Messages/Queues
 
-| Op     | Usage         | Description |
-| ------ | ------------- | ----------- |
-| topic  |               |             |
-|        | topic.send    |             |
-|        | topic.recieve |             |
-|        | topic.process |             |
-| queue  |               |             |
-|        | queue.process |             |
-| job    |               |             |
-|        | job.exec      |             |
-| celery |               |             |
+| Category | Usage         | Description |
+| -------- | ------------- | ----------- |
+| topic    |               |             |
+|          | topic.send    |             |
+|          | topic.recieve |             |
+|          | topic.process |             |
+| queue    |               |             |
+|          | queue.process |             |
+| job      |               |             |
+|          | job.exec      |             |
+| celery   |               |             |
--- a/src/docs/sdk/performance/span-operations.mdx
+++ b/src/docs/sdk/performance/span-operations.mdx
@@ -6,13 +6,7 @@ Span operations are a short code identifying the type of operation the span is m
 
 Operations are expected to follow [OpenTelemetry's semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/24de67b3827a4e3ab2515cd8ab62d5bcf837c586/specification/trace/semantic_conventions/README.md) as much as possible.
 
-Below is a table containing a list of span operations that are recommended to be used.
-
-Operations are expected to generally take the form: `CATEGORY.ACTION`, where the categories and actions can have optional identifiers. For example, a redis database operation takes the form `db.query`, where `db` is the category and `query` is the action. Another example is a mobile app rendering it's screen. It would have the operation `ui.load.hot`, where `ui` is the category, `load` is the action, and `hot` is the action identifier.
-
 It's important to keep categories consistent between SDKs and integrations as they are used by Sentry in the [Operations Breakdown](https://docs.sentry.io/product/sentry-basics/tracing/event-detail/#operations-breakdown) feature. For example, both `db.init` and `db.query` will be categorized as database operations (`db`). The default operations breakdown config can be seen [here](https://github.com/getsentry/sentry/blob/809b7fe54c6f06cc1e4c503cf83ded896472a011/src/sentry/projectoptions/defaults.py#L74).
-
-Having both a category and action is recommended. Operations should be lowercased.
 
 # List of Operations
 
--- a/src/docs/sdk/performance/span-operations.mdx
+++ b/src/docs/sdk/performance/span-operations.mdx
@@ -10,11 +10,15 @@ Below is a table containing a list of span operations that are recommended to be
 
 Operations are expected to generally take the form: `CATEGORY.ACTION`, where the categories and actions can have optional identifiers. For example, a redis database operation takes the form `db.sql.query`, where `db` is the category, `sql` is the category identifier, and `query` is the action. Another example is a mobile app rendering it's screen. It would have the operation `ui.load.hot`, where `ui` is the category, `load` is the action, and `hot` is the action identifier.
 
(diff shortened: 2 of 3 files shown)


src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+## Database
+
+| Op  | Usage            | Description                |
+| --- | ---------------- | -------------------------- |
+| db  |                  | An operation on a database |
+|     | db.connection    |                            |
+|     | db.transaction   |                            |
+|     | db.sql.query     |                            |
+|     | db.query         |                            |
```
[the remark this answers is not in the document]
maciejwalkowiak: Yes.

maciejwalkowiak:
@marandaneto with GraphQL on the server side, the transaction name is likely already fixed, so all queries/mutations go to the same transaction. Operation name is the way to distinguish what actually gets executed, which matters even more when there is more than one query sent with single request. I believe we should keep them dynamic.

marandaneto:
> @marandaneto with GraphQL on the server side, the transaction name is likely already fixed, so all queries/mutations go to the same transaction. Operation name is the way to distinguish what actually gets executed, which matters even more when there is more than one query sent with single request. I believe we should keep them dynamic.

the problem is that we have a chart with an operation breakdown, if its dynamic, its not gonna be readable, maybe the dynamic part should go under the description, but we can talk about that after agreeing on a specification here.

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+It's important to keep categories consistent between sdks and integrations as they are used by Sentry in the [Operations Breakdown](https://docs.sentry.io/product/sentry-basics/tracing/event-detail/#operations-breakdown) feature. For example, both `db.init` and `db.query` will be categorized as database operations. The default operations breakdown config can be seen [here](https://github.com/getsentry/sentry/blob/809b7fe54c6f06cc1e4c503cf83ded896472a011/src/sentry/projectoptions/defaults.py#L74).
+
+Having both a category and action is recommended. Operations should be lowercased.
+
+# List of Operations
+
+## Browser
+
+| Op         | Usage           | Description                                                                                                         |
```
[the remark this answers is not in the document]
AbhiPrasad: Yeah this is a function of this being WIP. Will rename `Op` -> `Category` to make my intentions here a little more clear. 

The `Usage` column was just a collection of operations I've seen being used throughout the SDKs

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,125 @@
...
+|            | resource.img    |                                                                                                                     |
+| browser    |                 | Usage of browser APIs or functionality                                                                              |
+|            | browser.paint   |                                                                                                                     |
+| mark       |                 | Usage of [performance.mark() API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark)                |
+| measure    |                 | Usage of [performance.measure() API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure)          |
+| user       |                 |                                                                                                                     |
+| animate    |                 |                                                                                                                     |
+
+### JS Frameworks
```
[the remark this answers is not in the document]
AbhiPrasad: Yes I think the natural next step here is to build out a set of semantic conventions for ourselves on how you would build different integrations and what type of info (error -> context, tags, capture calls, performance -> transactions, spans, tags) you collect.

I think for now we keep this here as a starting point.

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,125 @@
...
+| http      |                 |             |
+|           | http.client     |             |
+|           | http.server     |             |
+| rpc       |                 |             |
+| grpc      |                 |             |
+| template  |                 |             |
+|           | template.init   |             |
+|           | template.parse  |             |
+|           | template.render |             |
+| view      |                 |             |
```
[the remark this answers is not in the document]
AbhiPrasad: I think I might use `render` instead then.

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,125 @@
...
+| browser    |                 | Usage of browser APIs or functionality                                                                              |
+|            | browser.paint   |                                                                                                                     |
+| mark       |                 | Usage of [performance.mark() API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark)                |
+| measure    |                 | Usage of [performance.measure() API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure)          |
+| user       |                 |                                                                                                                     |
+| animate    |                 |                                                                                                                     |
+
+### JS Frameworks
+
+| Op      | Usage        | Description                                    |
```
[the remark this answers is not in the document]
AbhiPrasad: Added as part of Browser operations

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,125 @@
...
+|     | db.query.compile |                            |
+
+## Serverless (FAAS)
+
+| Op  | Usage       | Description |
+| --- | ----------- | ----------- |
+| aws |             |             |
+|     | aws.lambda  |             |
+|     | aws.request |             |
+| gcp |             |             |
```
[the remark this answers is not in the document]
AbhiPrasad: You think they should follow `http.server` possibly? We could also special case `serverless`.

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+| Op      | Usage             | Description |
+| ------- | ----------------- | ----------- |
+| django  |                   |             |
+|         | django.middleware |             |
+|         | django.view       |             |
+| express |                   |             |
+| rails   |                   |             |
+| rack    |                   |             |
+| console |                   |             |
```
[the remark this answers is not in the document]
AbhiPrasad: grabbed from https://github.com/getsentry/sentry-symfony/blob/5129f962a9b921f08fa2100fc5593cd6d8f06566/src/EventListener/TracingConsoleListener.php#L62, but this seems to be more related to running stuff on the command line. I'll move into web server (since using rails, django cmd line is pretty common operations).

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,125 @@
...
+| resource   |                 | Resource as per [Performance Resource Timing](https://w3c.github.io/resource-timing/#sec-performanceresourcetiming) |
+|            | resource.script |                                                                                                                     |
+|            | resource.link   |                                                                                                                     |
+|            | resource.css    |                                                                                                                     |
+|            | resource.img    |                                                                                                                     |
+| browser    |                 | Usage of browser APIs or functionality                                                                              |
+|            | browser.paint   |                                                                                                                     |
```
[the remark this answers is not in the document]
AbhiPrasad: For sure. We've been using plain categories (with no actions or identifiers) for now. We can change this, but I think that's something we can come back to.

image

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,125 @@
...
+title: "Span Operations"
+---
+
+Span operations are a short code identifying the type of operation the span is measuring. Span operations are low cardinality attributes - they should be as general as possible while still being human readable and useful.
+
+Operations are expected to follow [OpenTelemetry's semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/README.md) as much as possible.
+
+Below is a table containing a list of span operations that are recommended to be used.
+
+Operations are expected to generally take the form: `CATEGORY.ACTION`, where the categories and actions can have optional identifiers. For example, a redis database operation takes the form `db.sql.query`, where `db` is the category, `sql` is the category identifier, and `query` is the action. Another example is a mobile app rendering it's screen. It would have the operation `ui.load.hot`, where `ui` is the category, `load` is the action, and `hot` is the action identifier.
```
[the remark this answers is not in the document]
AbhiPrasad: Yeah we can decide on how to make individual platforms more consistent after we come to a general decision on the spec here.

AbhiPrasad (commented):
Thanks everyone for your feedback/comments. I'm going to work to polish this up and add more examples (including snippets from different SDKs) so we can do another round of reviews.

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,130 @@
...
+| ---------- | -------------- | ----------------------------------- |
+| app        |                | Data about the mobile app           |
+|            | app.start      |                                     |
+|            | app.start.warm |                                     |
+|            | app.start.cold |                                     |
+| ui         |                | An operation on a mobile/desktop UI |
+|            | ui.load        |                                     |
+|            | ui.action      |                                     |
+| navigation |                | Navigating to another screen        |
+| file       |                | Operations on the file system       |
```
philipphofmann, line 114: In a call, we discussed what to put in the description of the file span. We thought that filename + bytes count `file.txt (203 KB)` should go in the description and the full file path should go into data of the span as `data["filepath"] = "/absolute/example/file.txt"`. Should the bytes count be KB, MB, GB, ... ?

Can we document this here?

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,130 @@
...
+
+Having both a category and action is recommended. Operations should be lowercased.
+
+# List of Operations
+
+## Browser
+
+| Category    | Usage           | Description                                                                                                         |
+| ----------- | --------------- | ------------------------------------------------------------------------------------------------------------------- |
+| pageload    |                 | A full page load of a web application                                                                               |
```
philipphofmann, line 23: Should we specify what should go into the description of the spans as well? I think it would be awesome to specify them precisely, so these are consistent as well.

philipphofmann (commented):
Good job, @AbhiPrasad. I like the direction where this is going.

src/docs/sdk/performance/span-operations.mdx
AbhiPrasad: Yes I think this is a great idea, but out of the scope of this PR. I want to focus on getting the operations right, we can then take a look at descriptions next.

src/docs/sdk/performance/span-operations.mdx
AbhiPrasad: > Can we document this here?

We should - but I'm not going to include it in this PR.

Instead we should be trying to develop our own set of semantic conventions (especially for desktop/mobile/browser - areas where open telemetry doesn't have anything). I plan to take this on after this PR gets merged in.

philipphofmann: Okay, I created an issue so this doesn't get lost https://github.com/getsentry/develop/issues/463

src/docs/sdk/performance/span-operations.mdx
philipphofmann: I created an issue for that https://github.com/getsentry/develop/issues/463

AbhiPrasad:
I've updated the PR description with more details, and this is now ready for review. Any feedback is highly appreciated!

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,141 @@
...
+## Database
+
+Databased related spans are expected to follow OpenTelemetry's [Database](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md) semantic conventions when possible.
+
+| Category | Usage            | Description                |
+| -------- | ---------------- | -------------------------- |
+| db       |                  | An operation on a database |
+|          | db.connection    |                            |
+|          | db.transaction   |                            |
+|          | db.sql.query     |                            |
```
romtsn, line 95: Still a bit unclear what's the diff between `db.sql.query` and `db.query`? Should we just get rid of `db.sql.query`?

AbhiPrasad: According to the outlined spec `sql` is an identifier for the `db` category. I'm of the opinion that both are fine to exist - `db.query` is just the more general one. 

The key takeaway here is that since both use the same category, `db`, they will be classified the same way in the product. Everything after the `db` prefix just provides additional context, SDK developers should feel free to choose whatever they think fits the best (as long as the category stays consistent).

Open to discussing further if you feel that we should take a more opinionated approach.

src/docs/sdk/performance/span-operations.mdx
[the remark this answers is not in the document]
rhcarvalho: > I think Redis was an unfortunate example, it doesn't use SQL.

The Redis example also doesn't follow the `CATEGORY.ACTION` format we're just starting to introduce in the prose, so I'd suggest starting with something simpler, and following up with a more complex case.

```suggestion
Operations are expected to generally take the form: `CATEGORY.ACTION`, where the categories and actions can have optional identifiers. For example, an instrumented HTTP server creates transaction spans named `http.server`, where `http` is the category, and `server` is the category identifier. Another example is a mobile app rendering a screen with a span operation `ui.load.hot`, where `ui` is the category, `load` is the action, and `hot` is the action identifier.
```

Hmm not happy with `http.server`... I'd rather have something that follows the `CATEGORY.ACTION` format there.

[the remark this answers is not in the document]
rhcarvalho: 🤔 💭 

There's both "category identifier" and "action identifier" on top of "category" and "action", but in the current shape of the spec we only see 1, 2 or 3 dot-separated items making up the operation name.

I wonder if there's a simpler way to refer to the parts. `CATEGORY[.CAT_IDENT].ACTION[.ACTION_IDENT]` is ambiguous to parse (for computers and humans), and so the 4 names for the parts feel like unnecessary complexity that we can't really make good use of.

For example, we cannot split up the `span.op` and render something descriptive in Sentry like "Operation: DB; Action: Insert".

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+Below is a table containing a list of span operations that are recommended to be used.
+
+Operations are expected to generally take the form: `CATEGORY.ACTION`, where the categories and actions can have optional identifiers. For example, a redis database operation takes the form `db.sql.query`, where `db` is the category, `sql` is the category identifier, and `query` is the action. Another example is a mobile app rendering it's screen. It would have the operation `ui.load.hot`, where `ui` is the category, `load` is the action, and `hot` is the action identifier.
+
+It's important to keep categories consistent between sdks and integrations as they are used by Sentry in the [Operations Breakdown](https://docs.sentry.io/product/sentry-basics/tracing/event-detail/#operations-breakdown) feature. For example, both `db.init` and `db.query` will be categorized as database operations (`db`). The default operations breakdown config can be seen [here](https://github.com/getsentry/sentry/blob/809b7fe54c6f06cc1e4c503cf83ded896472a011/src/sentry/projectoptions/defaults.py#L74).
```
rhcarvalho, line 13:
```suggestion
It's important to keep categories consistent between SDKs and integrations as they are used by Sentry in the [Operations Breakdown](https://docs.sentry.io/product/sentry-basics/tracing/event-detail/#operations-breakdown) feature. For example, both `db.init` and `db.query` will be categorized as database operations (`db`). The default operations breakdown config can be seen [here](https://github.com/getsentry/sentry/blob/809b7fe54c6f06cc1e4c503cf83ded896472a011/src/sentry/projectoptions/defaults.py#L74).
```

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+Due to use of categories in the Sentry product, span categories are expected to be low cardinality data. **Categories should not be auto-generated**. SDK developers should feel free to automatically generate the actions and identifiers for span operations.
```
rhcarvalho, line 15: Needs verification

- [ ] We start out saying "Span operations are low cardinality attributes", and this sentence here relaxes that expectation, essentially accepting high-cardinality span operations, which sounds contradictory at a minimum, and might have negative performance consequences.

rhcarvalho (commented):
A pass on the spec, ignored the examples for now.

src/docs/sdk/event-payloads/properties/op.mdx
```diff
@@ -2,6 +2,8 @@
 
 : _Recommended_. Short code identifying the type of operation the span is measuring.
 
+For more details, see <Link to="/sdk/performance/span-operations/">Sentry's conventions around span operations</Link>.
+
   ```json
   {
     "op": "sql.query"
```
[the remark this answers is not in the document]
rhcarvalho: `http.server` is probably a case of `CATEGORY.CATEGORY_IDENTIFIER`, a `CATEGORY.ACTION` example would fit better.

rhcarvalho:
- [ ] Add a link from [/sdk/performance/](https://develop-git-abhi-span-operations.sentry.dev/sdk/performance/) to the [new Span Operations subpage](https://develop-git-abhi-span-operations.sentry.dev/sdk/performance/span-operations/)

src/docs/sdk/performance/span-operations.mdx
romtsn: Ah alright, I thought it will affect the operation breakdown rendering in the UI, but if it's just there for context, I'm good with it, thanks

src/docs/sdk/performance/span-operations.mdx
k-fish: The important point here isn't that it's generated, but that the span categories stay low cardinality. I think just recommending that if a span category **is** generated, that it should only use low cardinality inputs (a static array like how we create the ops breakdown measurements) and not something like an `id`, `url`, etc. 

Since `generated` is a fairly conflated term, I'd perhaps rewrite this to state that:
- It should be low cardinality
- If you **are** generating this somehow, take care to not use high cardinality inputs

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,141 @@
+---
+title: "Span Operations"
```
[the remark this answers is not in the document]
AbhiPrasad: Opened https://github.com/getsentry/develop/issues/464 to track that

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -0,0 +1,141 @@
...
+---
+
+Span operations are a short code identifying the type of operation the span is measuring. Span operations are low cardinality attributes - they should be as general as possible while still being human readable and useful. They should avoid including high cardinality data like ids and URLs.
```
rhcarvalho, line 5:
```suggestion
Span operations are a short code identifying the type of operation the span is measuring. Span operations are low cardinality attributes - they should be as general as possible while still being human readable and useful. They should avoid including high cardinality data like IDs and URLs.
```

nit: We use `ID`/`IDs` in prose here and in docs.sentry.io.

AbhiPrasad:
In the spirit of going step by step, I slightly changed the scope of the PR.

This PR now contains just a brief overview of operations and the table of the values we want to be used. This hopefully will help unblock the SDKs who just want the list of operations + encourage others to start adding to/refining this list.

I removed the note about operations structure (the stuff about `CATEGORY` and `ACTION` ) in https://github.com/getsentry/develop/pull/452/commits/c3c8428e05c2930e3a9b1f2711bfa0ec9620c0ce as that still needs to be discussed further. We can move those convos to another PR.

This is ready for another review

rhcarvalho (approved):
This is a good pass at documenting the current state.

rhcarvalho:
> In the spirit of going step by step, I slightly changed the scope of the PR.

👍 

Reminder to check if you want `Fixes #392` in the PR description or not, as the issue itself is twofold.

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -27,8 +27,8 @@ If a span operation is not provided, the value of `default` is used.
 |            | resource.img    |                                                                                                                     |
 | browser    |                 | Usage of browser APIs or functionality                                                                              |
 |            | browser.paint   |                                                                                                                     |
-|            | browser.mark    | Usage of [performance.mark() API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark)                |
-|            | browser.measure | Usage of [performance.measure() API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure)          |
+| mark       |                 | Usage of [performance.mark() API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark)                |
+| measure    |                 | Usage of [performance.measure() API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure)          |
```
rhcarvalho, line 31: Hmm I feel like I could live without this commit; but you adding it to this PR suggests the distinction between `Category` and `Usage` columns is not really necessary.

All we have as actual operations in SDKs are strings, and since we've dropped the intention to document how categories work we could as well have a single column:

| Operation | Description |
| --- | --- |
| ... | ... |

k-fish (approved):
This looks good to me, glad we added `ui.*` as I think it's something we need to be consistent in adding coverage for 👍

src/docs/sdk/performance/span-operations.mdx
AbhiPrasad: Will address this in the next PR.

PR #458, opened 2021-11-03
feat: Start react testing library guide

matejminar:
A follow-up to the discussion of the October 7th frontend TSC.

I kicked it off with tips around querying+assertions and pitfalls around that.

### DIFF
--- a/src/docs/frontend/using-rtl.mdx
+++ b/src/docs/frontend/using-rtl.mdx
@@ -1,0 +1,148 @@
+---
+title: Using React Testing Library
+---
+
+We are in the process of converting our tests from Enzyme to React Testing Library.
+In this guide, you'll find tips to follow best practices and anticipate common pitfalls.
+
+We have two ESLint rules in place to help with this:
+
+- [eslint-plugin-jest-dom](https://github.com/testing-library/eslint-plugin-jest-dom)
+- [eslint-plugin-testing-library](https://github.com/testing-library/eslint-plugin-testing-library)
+
+We strive to write tests in such a way that closely resembles how our application is used.
+
+So rather than dealing with instances of rendered components, we query the DOM in the same way the user would.
+We find form elements by their label text (just like a user would), we find links and buttons from their text (like a user would).
+
+As a part of this goal, we avoid testing implementation details so refactors (changes to implementation but not functionality) don't break the tests.
+
+We are generally in favor of Use Case Coverage over Code Coverage.
+
+## Querying
+
+- use `getBy...` as much as possible
+- use `queryBy...` only when checking for non-existence
+- use `await findBy...` only when expecting an element to appear after but the change to the DOM might not happen immediately
+
+To ensure that the tests resemble how users interact with our code we recommend the following priority for querying:
+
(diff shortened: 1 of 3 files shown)


src/docs/frontend/using-rtl.mdx
```diff
@@ -0,0 +1,148 @@
+---
+title: Using React Testing Library
+---
+
+We are in the process of converting our tests from Enzyme to React Testing Library.
+In this guide, you'll find tips to follow best practices and anticipate common pitfalls.
```
markstory, line 6:
```suggestion
In this guide, you'll find tips to follow best practices and avoid common pitfalls.
```

src/docs/frontend/using-rtl.mdx
```diff
@@ -0,0 +1,148 @@
...
+
+We have two ESLint rules in place to help with this:
+
+- [eslint-plugin-jest-dom](https://github.com/testing-library/eslint-plugin-jest-dom)
+- [eslint-plugin-testing-library](https://github.com/testing-library/eslint-plugin-testing-library)
+
+We strive to write tests in such a way that closely resembles how our application is used.
```
markstory, line 13:
```suggestion
We strive to write tests in a way that closely resembles how our application is used.
```

src/docs/frontend/using-rtl.mdx
```diff
@@ -0,0 +1,148 @@
...
+
+So rather than dealing with instances of rendered components, we query the DOM in the same way the user would.
```
markstory, line 15:
```suggestion
Rather than dealing with instances of rendered components, we query the DOM in the same way the user would.
```

src/docs/frontend/using-rtl.mdx
```diff
@@ -0,0 +1,148 @@
...
+
+As a part of this goal, we avoid testing implementation details so refactors (changes to implementation but not functionality) don't break the tests.
+
+We are generally in favor of Use Case Coverage over Code Coverage.
+
+## Querying
+
+- use `getBy...` as much as possible
+- use `queryBy...` only when checking for non-existence
+- use `await findBy...` only when expecting an element to appear after but the change to the DOM might not happen immediately
```
markstory, line 26:
```suggestion
- use `await findBy...` only when expecting an element to appear after a change to the DOM that might not happen immediately
```

src/docs/frontend/using-rtl.mdx
```diff
@@ -0,0 +1,148 @@
...
+  expect(screen.getByRole("alert")).toBeInTheDocument();
+});
+
+// ✅
+mountWithTheme(<Example />);
+expect(await screen.findByRole("alert")).toBeInTheDocument();
+```
+
+Avoid using `waitFor` for waiting on disappearance, use `waitForElementToBeRemoved` instead ([examples](https://github.com/getsentry/sentry/pull/29547)).
+The latter is using MutationObserver which is more efficient than polling the DOM at regular intervals with waitFor.
```
markstory, line 96:
```suggestion
The latter is uses `MutationObserver` which is more efficient than polling the DOM at regular intervals with waitFor.
```

markstory (approved):

k-fish (approved):
Awesome! My only comment is that I think `findX` is preferred over waitFor since find has wait under the hood already.

PR #461, opened 2021-11-11
feat(docs): Add code snippet for safely using defaults in new columns

leeandher:
This PR will add a code snippet to explain how to modify migrations to safely add defaults in new columns.

image


wedamija (approved):
Looks good, appreciate it!

PR #465, opened 2021-11-18
Update email.mdx

thibodelanghe:
Add SENTRY_MAIL_HOST configuration documentation fixing issue https://github.com/getsentry/onpremise/issues/1158

### DIFF
--- a/src/docs/self-hosted/email.mdx
+++ b/src/docs/self-hosted/email.mdx
@@ -8,7 +8,7 @@ title: 'Self-Hosted Email'
 
 ## Outbound Email
 
-Self-hosted Sentry ships with a built-in outgoing SMTP server powered by [exim4](https://hub.docker.com/r/tianon/exim4). The default configuration is set to use this server. All you need to do is to set a valid address for `mail.from` setting in `config.yml` and the FQDN of your Sentry instance for `SENTRY_MAIL_HOST` in `.env`. Keep in mind that if you start sending too many emails to public addresses, your new server may get marked as a spammer and banned.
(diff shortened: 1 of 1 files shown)


src/docs/self-hosted/email.mdx
```diff
@@ -8,7 +8,7 @@ title: 'Self-Hosted Email'
 
 ## Outbound Email
 
-Self-hosted Sentry ships with a built-in outgoing SMTP server powered by [exim4](https://hub.docker.com/r/tianon/exim4). The default configuration is set to use this server. All you need to do is to set a valid address for `mail.from` setting in `config.yml` . Keep in mind that if you start sending too many emails to public addresses, your new server may get marked as a spammer and banned.
+Self-hosted Sentry ships with a built-in outgoing SMTP server powered by [exim4](https://hub.docker.com/r/tianon/exim4). The default configuration is set to use this server. All you need to do is to set a valid address for `mail.from` setting in `config.yml` and the FQDN of your Sentry instance for `SENTRY_MAIL_HOST` in `.env`. Keep in mind that if you start sending too many emails to public addresses, your new server may get marked as a spammer and banned.
```
chadwhitacre, line 11: How about linking FQDN to save a few clicks for folks unaware of the term?

```suggestion
Self-hosted Sentry ships with a built-in outgoing SMTP server powered by [exim4](https://hub.docker.com/r/tianon/exim4). The default configuration is set to use this server. All you need to do is to set a valid address for `mail.from` setting in `config.yml` and the [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of your Sentry instance for `SENTRY_MAIL_HOST` in `.env`. Keep in mind that if you start sending too many emails to public addresses, your new server may get marked as a spammer and banned.
```

chadwhitacre:
Thanks @thibodelanghe! :)

PR #470, opened 2021-11-26
ref(RTL): Add tip  with recommended library for testing user interactions

priscilawebdev:
Add a new tip to the React Test Library doc, recommending using `userEvent` instead of `fireEvent` whenever possible.

### DIFF
--- a/src/docs/frontend/using-rtl.mdx
+++ b/src/docs/frontend/using-rtl.mdx
@@ -146,3 +146,28 @@ expect(screen.getByText("Hello World")).toBeInTheDocument();
 mountWithTheme(<Example />);
 expect(screen.getByText(/hello world/i)).toBeInTheDocument();
 ```
+
+Use `@testing-library/user-event` over `fireEvent` where possible.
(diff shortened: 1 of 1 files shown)


src/docs/frontend/using-rtl.mdx
```diff
@@ -146,3 +146,28 @@ expect(screen.getByText("Hello World")).toBeInTheDocument();
 mountWithTheme(<Example />);
 expect(screen.getByText(/hello world/i)).toBeInTheDocument();
 ```
+
+Use `userEvent` over `fireEvent` where possible.
+`userEvent` comes from the package`@testing-library/user-event` which is built on top of `fireEvent`, but it provides several methods that resemble the user interactions more closely.
```
matejminar, line 151:
```suggestion
`userEvent` comes from the package `@testing-library/user-event` which is built on top of `fireEvent`, but it provides several methods that resemble the user interactions more closely.
```

matejminar (approved):

PR #491, opened 2022-01-13
Document alternative format for Exception/Breadcrumb/Thread payload item

rhcarvalho:
Sentry's ingestion server can deal with flat lists, and some SDKs have historically used that. To avoid confusion, document that possibility.

For implementation details, see
https://github.com/getsentry/relay/blob/4639455273ed8caccf5c11b929ef42d3a7ac6d61/relay-general/src/protocol/types.rs#L29
https://github.com/getsentry/relay/blob/4639455273ed8caccf5c11b929ef42d3a7ac6d61/relay-general/src/protocol/event.rs#L406-L430

Related: https://github.com/getsentry/sentry-go/pull/408


untitaker (approved):

untitaker:
FTR I believe we should not generally document all normalization behavior, or the full set of accepted values by Relay, only the things that we think are reasonable for an SDK to send. Relay supports a ton of legacy SDKs that send all kinds of malformed data that we need to support until the end of time, but that doesn't mean we should write documentation about it or build SDKs that rely on this behavior.

I consider `{"values": ...}` to be the more canonical way of sending things, but it is "reasonable" to want to send a list there.

rhcarvalho:
> I believe we should not generally document all normalization behavior, or the full set of accepted values by Relay, (...)

I agree 👍

> I consider {"values": ...} to be the more canonical way of sending things (...)

Also in agreement here.

Thanks for the notes, @untitaker. It's good to have our reasoning written down.

PR #492, opened 2022-01-18
ref(rate-limits): Promote 60s default rate limit

jan-auer:
The description of a default rate limit is easy to miss since it's not declared
at response handling but rather the interpretation section. I'm moving this up
to the earlier list to make it more apparent.


jan-auer:
Tagging @bruno-garcia since he originally introduced it.

bruno-garcia (approved):

bruno-garcia:
Makes sense, thanks

PR #494, opened 2022-01-20
ref: Improve Code review docs

markstory:
Add a few more points to consider during code review.

### DIFF
--- a/src/docs/code-review.mdx
+++ b/src/docs/code-review.mdx
@@ -20,7 +20,7 @@ Because Sentry is an open source project maintained via GitHub we want to ensure
 
 ### GitHub Teams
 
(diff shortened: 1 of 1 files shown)


src/docs/code-review.mdx
```diff
@@ -26,17 +36,13 @@ Additionally, language specific teams exist, primarily for SDKs:
 - [@getsentry/php](https://github.com/orgs/getsentry/teams/php)
 - [@getsentry/python](https://github.com/orgs/getsentry/teams/python)
```
AbhiPrasad, line 37: While we are here, we can probably get rid of the language specific teams (which are mostly now defunct). Instead I would prefer to link to https://open.sentry.io/structure/ which list all the GH teams as appropriate.

AbhiPrasad (approved):

PR #497, opened 2022-01-27
ref(integrations): Add Jira Server docs

ceorourke:
Add documentation for how on prem users can install Jira Server. I opted to link to the main docs page since it's 99% the same to avoid maintaining 2 versions.


src/docs/integrations/index.mdx
```diff
@@ -2,13 +2,12 @@
 title: Integrations
 ---
 
-TODO: Describe the difference between integration/plugin generations.
```
ceorourke, line 5: If anyone feels strongly about leaving this in, I can, but it's bothered me for a while (this is visible, it is not a comment).

NisanthanNanthakumar (approved):

PR #504, opened 2022-02-16
feat(docs): update dev api doc instructions

JoshFerge:
- Update small mistakes / broken commands
- Flesh out examples more with annotated concrete examples
- Add more helpful links

### DIFF
--- a/src/docs/api/public.mdx
+++ b/src/docs/api/public.mdx
@@ -157,10 +157,10 @@ Write a test [here](https://github.com/getsentry/sentry/tree/master/tests/apidoc
 
 **Tips**:
 
-- `make build-api-docs` will simply attempt to build the OpenAPI JSON. The build will fail if there are any warnings.
-- `make test-api-docs` runs all API docs tests. This command also builds the derefed JSON. This runs in CI but can be useful for local development to make sure everything passes.
-- `yarn run build-derefed-docs` builds the derefed JSON. This is useful in the case where you're only running one test locally, making a change to the schema, and then re-running the test. Since the tests run against the derefed version, you'll need to rebuild it.
-  **To see your changes in the docs locally**:
+- `make test-api-docs` builds the OpenAPI JSON and runs all API docs tests.
+- To only build the docs, run `make build-api-docs`. The build will fail if there are any warnings.
+
+**To see your changes in the docs locally**:
 
 In `sentry`:
 
--- a/src/docs/api/public.mdx
+++ b/src/docs/api/public.mdx
@@ -91,7 +91,7 @@ from sentry.apidocs.decorators import declare_public
 class OrganizationSCIMMemberDetails(...):
 ```
 
-Tag the endpoint view if needed (A parent class may already be tagged), this is how the endpoint makes it into a particular [sidebar endpoint grouping](https://docs.sentry.io/api/). You can see the current list of tags or add tags [here](https://github.com/getsentry/sentry/blob/master/src/sentry/apidocs/build.py).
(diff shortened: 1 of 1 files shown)


src/docs/api/public.mdx
```diff
@@ -80,31 +80,50 @@ API Documentation consists of a few items:
...
 ```python
 from sentry.apidocs.decorators import declare_public
 
-@declare_public(methods={"GET"})
+@public(methods={"GET"})
 class OrganizationSCIMMemberDetails(...):
 ```
 
-Tag the endpoint view if needed (you can also tag parent classes), this is how the endpoint makes it into a particular sidebar grouping.
+Tag the endpoint view if needed (A parent class may already be tagged), this is how the endpoint makes it into a particular [sidebar endpoint grouping](https://docs.sentry.io/api/). You can see the current list of tags or add tags [here](https://github.com/getsentry/sentry/blob/master/src/sentry/apidocs/build.py).
```
RyanSkonnord, line 94:
```suggestion
Tag the endpoint view if needed (a parent class may already be tagged), this is how the endpoint makes it into a particular [sidebar endpoint grouping](https://docs.sentry.io/api/). You can see the current list of tags or add tags [here](https://github.com/getsentry/sentry/blob/master/src/sentry/apidocs/build.py).
```
Capitalization typo?

src/docs/api/public.mdx
```diff
@@ -125,20 +157,22 @@ Write a test [here](https://github.com/getsentry/sentry/tree/master/tests/apidoc
 
 **Tips**:
 
+- `make build-api-docs` will simply attempt to build the OpenAPI JSON. The build will fail if there are any warnings.
 - `make test-api-docs` runs all API docs tests. This command also builds the derefed JSON. This runs in CI but can be useful for local development to make sure everything passes.
```
RyanSkonnord, line 161: Maybe we could combine the bullet points for `make build-api-docs` and `make test-api-docs` to make clear that `test-api-docs` invokes `build-api-docs` as a prerequisite. (That's what "This command also builds the derefed JSON" is telling us, but the detail seems easy to misunderstand.)

src/docs/api/public.mdx
```diff
@@ -125,20 +157,22 @@ Write a test [here](https://github.com/getsentry/sentry/tree/master/tests/apidoc
...
 - `yarn run build-derefed-docs` builds the derefed JSON. This is useful in the case where you're only running one test locally, making a change to the schema, and then re-running the test. Since the tests run against the derefed version, you'll need to rebuild it.
-
```
RyanSkonnord, line 130: Need to restore this line break; this messes up the Markdown syntax.

src/docs/api/public.mdx
```diff
@@ -113,8 +132,21 @@ Specify the parameters using a list, and a serializer if needed.
 Specify the request and response serializers.
 
 - Request serializers, since they use DRF serializers, can generally just be the serializer itself. There is plenty of customization and if needed can be overriden with `inline_serializer`. These will be passed to the `request` parameter in `extend_schema`.
-- For our response serializers, you must type the `serialize` function's return with a `TypedDict` and pass the serializer as a parameter in `extend_schema`.
+- For our response serializers, you must type the `serialize` function's return with a `TypedDict` and pass the serializer as a parameter in `extend_schema`. See [here](https://github.com/getsentry/sentry/blob/master/src/sentry/api/serializers/models/organization_member.py) for an example of this.
```
RyanSkonnord, line 135: There probably isn't much risk that the example will move out from under this link, but in case you want to hedge against it, we could link to the particular hash that's in the master branch right now.

RyanSkonnord (approved):
Great stuff!

src/docs/api/public.mdx
JoshFerge: I'd like to keep it live for now, in case we add any more improvements to it, but will try and watch out for breaking this link.

PR #515, opened 2022-03-03
feat(measurements): Add new measurements section to transaction

sl0thentr0py:


src/docs/sdk/event-payloads/properties/measurements.mdx
```diff
@@ -0,0 +1,18 @@
+`measurements`
+
+: _Optional_. An object containing standard/custom measurements with keys signifying the name of the measurement.
+
+Standard measurement keys currently supported are from the set  `"fp" | "lcp" | "fid" | "ttfb" | "ttfb.requesttime"`.
```
AbhiPrasad, line 5: Can we clarify what "supported" means here?

Also, what do you think about linking to the relay/sentry allow list code here? 

This is all browser web vitals, we also need to include the mobile vitals here as well.

sl0thentr0py, line 5: I took these from [here](https://github.com/getsentry/sentry/blob/6f31e2fc489a823c9395bc23fcd5389b3dff509c/src/sentry/discover/arithmetic.py#L156-L161) and also from the various setters in [`sentry-javascript/metrics`](https://github.com/getsentry/sentry-javascript/blob/f715e8754c11797120989e52e2342c89b98b5ea1/packages/tracing/src/browser/metrics.ts#L235). Feel free to add/update this list.

src/docs/sdk/event-payloads/properties/measurements.mdx
```diff
@@ -0,0 +1,18 @@
...
+
+Custom measurements need units to be specified. Units currently supported are durations from the set  `"ns" | "ms" | "s"`.
```
sl0thentr0py, line 7: and these I took from [`relay/MetricUnit`](https://github.com/getsentry/relay/blob/1e45a8817e45408ecbdbd3f0a679a5f5e4885290/relay-metrics/src/protocol.rs#L47-L52).

AbhiPrasad (commented):
Pinging Kevan to also take a look here.

src/docs/sdk/event-payloads/properties/measurements.mdx
AbhiPrasad: Hmm interesting that the mobile vitals are not in the Sentry allow list. So you can't do discover arithmetic on them?

sl0thentr0py: I think 'supported' is open to discussion based on either
* what is currently implemented end-end in browser/mobile
* what gets assigned units automatically if we say we _need_ units for custom measurements from the user (?)

sl0thentr0py: there is another list [here](https://github.com/getsentry/sentry/blob/6f31e2fc489a823c9395bc23fcd5389b3dff509c/src/sentry/search/events/constants.py#L124-L134) which also has `cls`.

AbhiPrasad: Hmm perhaps @wmak can help figure this out here

wmak: Hmm not being able to do arithmetic on vitals is definitely an oversight on my part (ie. we didn't update this allowlist when working on it)

But i believe the standard measurement list should be everything we prefix measurements on in the UI? ie. we should include the ones [here](https://github.com/getsentry/sentry/blob/a8c960a933d2ded5225841573d8fc426a482ca9c/static/app/utils/discover/fields.tsx#L654-L676) that aren't rates or percentages (those are computed)

src/docs/sdk/event-payloads/properties/measurements.mdx
```diff
@@ -0,0 +1,43 @@
...
+  "ttfb",
+  "ttfb.requesttime",
+  // mobile
+  "app_start_cold",
+  "app_start_warm",
+  "frames_total",
+  "frames_slow",
+  "frames_frozen",
+  "frames_slow_rate",
+  "frames_frozen_rate",
```
marandaneto, line 24: I believe we don't use them, @philipphofmann what's about on iOS?

src/docs/sdk/event-payloads/properties/measurements.mdx
```diff
@@ -0,0 +1,43 @@
...
+  "stall_count",
+  "stall_total_time",
+  "stall_longest_time",
+  "stall_percentage"
```
marandaneto, line 28: I didn't find the usage of this one on RN.

bruno-garcia: It's referenced by Sentry: https://github.com/getsentry/sentry/blob/7447331d8e02031d585bd0c0aec76ed1334b9bde/src/sentry/search/events/constants.py#L32

Couldn't find it in Relay, not sure how this works. Maybe @untitaker can help.

bruno-garcia (commented):
Are we including measurements that are not sent by the client but are generated on the backend?

src/docs/sdk/event-payloads/properties/measurements.mdx
sl0thentr0py: One of the open questions is if relay 'knows' the difference between standard/custom measurements or if this is only reflected in the sentry codebase.

jan-auer: Regarding `stall_*` measurements: I searched through all of getsentry, and it looks like they were _only_ defined in Sentry. Later on, we pulled them into Relay and the ingest load tester. If they are not emitted by any SDK, we should remove them.

Regarding units: Relay will have to know about units of the past built-in measurements for backwards compatibility. For new measurements, and also going forward for all SDKs, it would be better to declare all units on the client and send them in to avoid mismatches and outdated lists.

untitaker: if I understood @jan-auer correctly, relay will have to know existing/standard measurements to infer their units.

currently only sentry has such a list, whether that list moves to Relay is tbd (we currently send it from sentry to relay via projectconfig)

marandaneto: stall measurements are sent by the [RN SDK](https://github.com/getsentry/sentry-react-native/blob/main/src/js/tracing/stalltracking.ts)

jan-auer: Thanks, and the rates / percentages are defined in Sentry so they don't have to be added to the public interface:
https://github.com/getsentry/sentry/blob/1c4ee6be215dadbf9bec19b3b428464a1dfe05b3/src/sentry/search/events/fields.py#L380-L418

jan-auer: Actually, I'm wondering whether we should mark them as *reserved* and then never use them. They'll be defined as derived metrics in Sentry and can never be ingested.

src/docs/sdk/event-payloads/properties/measurements.mdx
jan-auer: Could you check which units would actually make sense and how we would like to specify units? What we have in Relay was only a stub, and now is a good time to rethink this. Generally, we have two options:
- Define a strict allow list of units that we define in Relay, and then propagate down.
- Allow arbitrary strings as unit. Relay still lists well-known units that it can handle, eg. for conversion.

src/docs/sdk/event-payloads/properties/measurements.mdx
jan-auer: Let's not do any special casing for the three computed "measurements", but document them as reserved. That means, we should not reject the names in SDKs or during ingestion, but we should let people know that the name is used in Sentry.

There can be multiple metrics with the same name, and they are disambiguated in the API by type, so in the product we'll be fine. I'm still working out what that would mean for Discover though.

@wmak sharing a thought here that we can follow up on somewhere else: We should evaluate renaming those three specific measurements:
- Two are called `*_rate` like everything else in the product, but one is called `_percentage`. As opposed to functions like `failure_rate()` they _can_ be computed within a single transaction though, if I'm not mistaken?
- They share the same namespace with ingested measurements but they are not actually ingested. I wonder if they should actually be materialized in Relay.

src/docs/sdk/event-payloads/properties/measurements.mdx
sl0thentr0py: ok so making a list of potentially useful categories, maybe product people have to add here.


| type | units | remarks |
| ---- | ---- | ---- |
| time | ns, microseconds, ms, s, m | longer than minute not useful, formatting for microseconds |
| memory | bytes, KiB, MiB, GiB, TiB | KB vs KiB |
| unitless / absolute counts | |
| unitless / percentages | % |

Formatting consistently for some cases is problematic if we go with short forms, for instance `microseconds` vs `ns` and I don't know if we want to get into unicode mess here.

sl0thentr0py: I'm not sure of the value provided by arbitrary user strings.
There's no semantic associated with a random `foobar` unit, so I assume it will just be a passthrough all the way to wherever the metric will be displayed on the UI. I think it is more valuable to actually look at valid measurement use cases and we can always extend our supported list with things that make sense.

sl0thentr0py: ok looking around a bit, should have done this first
* [Datadog](https://docs.datadoghq.com/metrics/units/)
* [The Unified Code for Units of Measure](https://ucum.org/ucum.html)
* [OpenTelemetry unmerged PR on units spec](https://github.com/open-telemetry/opentelemetry-specification/pull/1177)

src/docs/sdk/event-payloads/properties/measurements.mdx
wmak: They can be computed within a single transaction :+1:
Is it possible to materialize these in relay if they're going to be a metric as well? Would simplify the product a bit

sl0thentr0py:
Summarizing the discussion so far.

1. I've removed `frames_slow_rate`, `frames_frozen_rate` and `stall_percentage` since they are computed/derived from existing meaurements ([ref: jira](https://getsentry.atlassian.net/browse/INGEST-1057)).

2. The units topic is a larger discussion that will be continued in a separate PR but for now:
  * We will not go with arbitrary user defined unit strings but try to identify use cases which make sense.
    * The downside to this is SDKs will need validation on the unit so users don't send random stuff. If someone has opinions  on this, let us know. 
  * Currently we will just go with time/duration units `[ns, ms, s]` that are already defined in `MetricUnit` in relay.

src/docs/sdk/event-payloads/properties/measurements.mdx
jan-auer: Yes, in that case we can do the division in Relay, write them back into the measurements array, before extracting metrics for all measurements including the computed ones. I'd have to think a bit about migrating this forward, for a while we should probably keep the computation in discover around until we've reached 90d of history.

PR #520, opened 2022-03-10
feat: Disable relay by default in devservices

dcramer:
Requires getsentry/sentry#32497

### DIFF
--- a/src/docs/environment/index.mdx
+++ b/src/docs/environment/index.mdx
@@ -267,6 +267,12 @@ If you would like to be able to run `devserver` outside of your root checkout, y
   displayed correctly.
 </Alert>
 
+### A Note on Relay
+
+Some services are not run in all situations, and one of those is <Link to="/services/relay/">Relay</Link>. If you need a more production-like environment
+in development, you can set `SENTRY_USE_RELAY=True` in `~/.sentry/sentry.conf.py`. This will launch relay as part of the
+`devserver` workflow.
+
 ## Setting up Getsentry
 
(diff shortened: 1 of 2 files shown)


src/docs/environment/index.mdx
```diff
@@ -267,6 +267,12 @@ If you would like to be able to run `devserver` outside of your root checkout, y
   displayed correctly.
 </Alert>
 
+### A Note on Relay
+
+Some services are not run in all situations, and one of those is <Link to="/services/relay/">Relay</Link>. If you need a more production-like environment
+in development, you can set `SENTRY_USE_RELAY=True` in `~/.sentry/sentry.conf.py`. This will launch relay as part of the
+`devserver` workflow.
```
jan-auer, line 274: Suggestion as per https://github.com/getsentry/sentry/pull/32497#issuecomment-1064419810:

```suggestion
### Ingestion Pipeline (Relay)

Some services are not run in all situations, among those are <Link to="/services/relay/">Relay</Link> and the ingest workers. If you need a more production-like environment
in development, you can set `SENTRY_USE_RELAY=True` in `~/.sentry/sentry.conf.py`. This will launch Relay as part of the
`devserver` workflow.
```

jan-auer (approved):

dcramer:
Aside you'll note I moved Env to the top of our dev docs, because generally it is the first and largest complexity a contributor needs to tackle. There are some areas that could use improvement elsewhere, like the 'Workflow' page which feels very out of place now and misaligned with other pages.

PR #388, opened 2021-07-27
Add troubleshooting section on Postgres disk space

chendo:
This PR adds a section to the Troubleshooting guide on dealing with Postgres disk space usage due to `nodestore` and what happens if Postgres fails to release disk space from deleted rows.

From https://github.com/getsentry/onpremise/issues/783#issuecomment-880448830

PR #523, opened 2022-03-12
Fix gp_repack docker-compose command

yegorLitvinov:
Command with `run` command will fail with `Can't connect to socket..` error for `pg_repack`


yegorLitvinov:
@chendo @BYK Could you please review this pr. 
Related PR is https://github.com/getsentry/develop/pull/388

chendo:
@yegorLitvinov looks good. `exec` is what I have in my own scripts, not sure how that became `run` here, my bad

BYK (changes requested):
This one doesn't make much sense to me. `exec` would execute this command in a live container, requiring the system to be online. I would advice against this. If the system is taken down, then using `run` should not cause any issues. Do you know why you get the error?

chendo:
Presumably because if you `run`, it doesn't run Postgres and thus it can't connect to the postgres socket. We run ours in a cronjob while Sentry is running. `pg_repack` is designed to be run online

BYK:
>Presumably because if you run, it doesn't run Postgres and thus it can't connect to the postgres socket. 

Ahh, right as we are overriding the default `CMD` for Postgres. I think all we need is a `postgres` or `postgres &` call in that `run` line and we don't need `exec` or the system being up. Would you mind trying that?

yegorLitvinov:
@BYK the `run` command you advised is working well, but I can't check the result as there is not enough space on my disk for `pg_repack`. @chendo Could you test it?

src/docs/self-hosted/troubleshooting.mdx
```diff
@@ -109,9 +109,9 @@ An example script below:
 
 ```shell
 # Only keep the last 7 days of nodestore data. We heavily use performance monitoring.
-docker-compose run -T web cleanup --days 7 -m nodestore -l debug
+docker-compose run --rm -T web cleanup --days 7 -m nodestore -l debug
 # This ensures pg-repack exists before running as the container gets recreated on upgrades
-docker-compose run -T postgres bash -c "apt update && apt install -y --no-install-recommends postgresql-9.6-repack && su postgres -c 'pg_repack -E info -t nodestore_node'"
+docker-compose run --rm -T postgres bash -c "/usr/local/bin/docker-entrypoint.sh postgres & apt update && apt install -y --no-install-recommends postgresql-9.6-repack && su postgres -c 'pg_repack -E info -t nodestore_node'"
```
BYK, line 114: I think the main issue here is `postgresql-9.6-repack` part as we are now on [v14.11](https://github.com/getsentry/self-hosted/blob/419d6ccf4e4d0dde2db301e43370d019f9a4c69b/docker-compose.yml#L135)?

PR #529, opened 2022-03-24
fix: Data categories for client reports

philipphofmann:
The data categories used for client reports are not the same as for
the envelope item types.

### DIFF
--- a/src/docs/sdk/envelopes.mdx
+++ b/src/docs/sdk/envelopes.mdx
@@ -139,7 +139,7 @@ There are two generic headers for every Item:
 `type`
 
 : **String, required.** Specifies the type of this Item and its contents. Based
-  on the Item type, more headers may be required. See Data Integrity for a list
+  on the Item type, more headers may be required. See [Data Model](#data-model) for a list
   of all Item types.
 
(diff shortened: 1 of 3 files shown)


src/docs/sdk/envelopes.mdx
```diff
@@ -255,7 +255,9 @@ EOF:**
 Each Envelope consists of headers and a potentially empty list of Items, each
 with their own headers. Which Headers are required depends on the Items in an
 Envelope. This section describes all Item types and their respective required
-headers.
+headers. It is worth noting that the list of Item types doesn't match the data
```
marandaneto, line 258:
```suggestion
headers. It is worth noting that the list of Item types don't match the data
```

`don't` since list of item types is plural, I guess?!

marandaneto (approved):
Thanks for keeping the specs up to date.

src/docs/sdk/envelopes.mdx
philipphofmann: Hmm, I'm referring to the list here, which is singular and not plural. I think singular should be fine.

PR #531, opened 2022-03-25
feat: Clarify expectations for client reports

philipphofmann:
This PR clarifies the expectations for client reports.
It is not required that the numbers for these are 100%
accurate and therefore the implemenation can loose some
data. Communicating this clearly will stop developers
implementing a more complex solution which would ensure
to never loose any client reports.


src/docs/sdk/client-reports.mdx
```diff
@@ -98,14 +98,23 @@ need to match the reason codes that relay would emit directly to Sentry.
...
+the SDKs should ensure not dropping too many reports. It is not required, for example: 
+ - to persist the data when an application crashes.
+ - to move an envelope item with a client report to the next envelope when the cache for envelopes is full. 
+
+SDKs are encouraged to reduce the total amount of needless communication, especially
+on mobile. A recommended approach is to keep track of the discard events in the
+transport directly and attach it to an already scheduled envelope or periodically flush
+them out as separate envelope items.
+For mobile SDKs, the recommendation is to attach the client report to the next envelope
+enqueued for submission instead of periodically flushing them out. We accept the tradeoff
```
marandaneto, line 113: This part looks a bit weird, we mention that SDKs are encouraged to reduce the total of communication, but then we say: `or periodically flush
them out as separate envelope items.`

philipphofmann: I didn't write that initially. I guess the intention of @mitsuhiko was that you shouldn't send an envelope item each time an event is dropped. Especially on the backend you could have several items dropped per second and doing an HTTP request for each of them would be crazy. I will try to rephrase that a bit.

philipphofmann: Is it better now @marandaneto?

mitsuhiko (approved):

src/docs/sdk/client-reports.mdx
marandaneto: yes, thanks.

PR #532, opened 2022-03-25
feat: Don't double count client reports

philipphofmann:


src/docs/sdk/client-reports.mdx
```diff
@@ -109,3 +109,8 @@ instead or attach it to pending session envelopes.
 There is no expectation that such bookkeeping can work transparently for custom
 transports which is why it's generally acceptable if client reports are optional
 for custom transports.
+
+The one who drops an envelope item must record and report it. If the server drops
+an envelope item, for example, with the response 429, the client SDK must not record
+this as the server already does. Still, the SDK must record lost envelope items
+when dropping them itself caused by an active rate limit.
```
marandaneto, line 116: > the SDK must record lost envelope items
> when dropping them itself caused by an active rate limit.

Just to confirm that I understood it right if the server drops an item, we don't record it, but if we drop an item on the client because it's within the rate limit (e.g. 30s), then we do, right?

bruno-garcia (approved):
Lgtm

src/docs/sdk/client-reports.mdx
philipphofmann: Yes, exactly.

PR #535, opened 2022-03-28
Add client report item type to envelopes spec

mattjohnsonpint:
The [Envelopes Data Model](https://develop.sentry.dev/sdk/envelopes/#data-model) lists each possible envelope item type.  Client Reports was documented on its own page, but was missing from this list.

### DIFF
--- a/src/docs/sdk/envelopes.mdx
+++ b/src/docs/sdk/envelopes.mdx
@@ -465,6 +465,26 @@ The item contains a user feedback / user report JSON payload:
 
 *None*
 
+### Client Report
+
+Item type `"client_report"` contains a client report payload encoded in JSON.
+
+See the <Link to="/sdk/client-reports/">client reports</Link> documentation for the payload
+details.
+
+**Constraints:**
+
+- This Item may occur at most once per Envelope.
+- This Item can either be included in an Envelope with other Items, or it may be sent by itself.
+
+**Envelope Headers:**
+
+*None*
+
+**Additional Item Headers:**
+
+*None*
+
 ### Reserved Types
 
 Reserved types may not be written by any implementation. They are reserved for


src/docs/sdk/envelopes.mdx
```diff
@@ -465,6 +465,26 @@ The item contains a user feedback / user report JSON payload:
...
+### Client Report
+
+Item type `"client_report"` contains a client report payload encoded in JSON.
+
+See the <Link to="/sdk/client-reports/">client reports</Link> documentation for the payload
+details.
+
+**Constraints:**
+
+- This Item may occur at most once per Envelope.
```
mattjohnsonpint, line 477: Is this correct, or could multiple client reports be sent in a single envelope?

src/docs/sdk/envelopes.mdx
```diff
@@ -465,6 +465,26 @@ The item contains a user feedback / user report JSON payload:
...
+- This Item can either be included in an Envelope with other Items, or it may be sent by itself.
+
+**Envelope Headers:**
+
+*None*
```
mattjohnsonpint, line 482: I don't believe a client report sent by itself would require any specific envelope headers, since there's no event id specifically associated with the report, and the timestamp is already in the report payload.  Correct?

AbhiPrasad: afaik this is correct

src/docs/sdk/envelopes.mdx
philipphofmann: I'm not sure. @jan-auer or @untitaker, I'm pretty sure one of you can answer that.

philipphofmann (approved):
Apart from the open question, `multiple client reports be sent in a single envelope?` LGTM. Thanks for helping to improve our docs, @mattjohnsonpint 👏

src/docs/sdk/envelopes.mdx
untitaker: You can technically send multiple client reports. If it makes things easier for e.g. multi-language SDKs, sure, but generally please try to avoid sending more client reports than necessary.

PR #557, opened 2022-04-12
doc: Clarify the existence of client_os

untitaker:
See also https://www.notion.so/sentry/Conflicting-OS-contexts-6b4b157034c54c02a97ea8523bee265d#6aa1e8cc97f4444bb33d180ecf7c3adf

### DIFF
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -208,7 +208,7 @@ confusion about which OS context represents what. So here's some examples:
   the SDK (if at all).
 
   Additionally, the Sentry server will attempt to parse the `User-Agent` header
-  from the event's [./request](Request Interface) and create a _secondary_ OS
+  from the event's [Request Interface](/sdk/event-payloads/request/) and create a _secondary_ OS
   context under the non-default key `"client_os"`.
 
 * In events reported from a JS web frontend, the SDK typically reports no OS
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -196,11 +196,44 @@ The `type` and default key is `"device"`.
 
 ## OS Context
 
-OS context describes the operating system on which the event was created. In web
-contexts, this is the operating system of the browser (generally pulled from the
-User-Agent string).
+OS context under the default key `"os"` describes the operating system on which
+the crash happened/the event was created.
 
-The `type` and default key is `"os"`.
+The `type` and default key is `"os"`. However, since contexts can be set
+multiple times under different keys, there has historically been a lot of
+confusion about which OS context represents what. So here's some examples:
+
+* In events reported from a Python/ASPNET/Rails backend app, the OS context
+  under the default key `"os"` is the server's operating system, and is set by
+  the SDK (if at all).
+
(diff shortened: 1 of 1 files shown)


bruno-garcia:
Adding @mattjohnsonpint just for his info but he's OOO for a week

jjbayer (commented):
Thanks for documenting this! While it's good for clarifying the current situation, in the long run I would not mention `server_os`, or even the possibility to create keys other than `os` or `client_os`.

untitaker:
>While it's good for clarifying the current situation, in the long run I would not mention server_os, or even the possibility to create keys other than os or client_os.

I think that people do not only look at these docs to figure out what to send, but also to explain what they're seeing in the JSON that they can download from the server. SDK devs are sometimes curious how they can replicate what they see in web frontend events and generally do review how the event looks in issue details

jjbayer (approved):

jjbayer (approved):

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -196,11 +196,44 @@ The `type` and default key is `"device"`.
...
-User-Agent string).
+OS context under the default key `"os"` describes the operating system on which
+the crash happened/the event was created.
 
-The `type` and default key is `"os"`.
+The `type` and default key is `"os"`. However, since contexts can be set
+multiple times under different keys, there has historically been a lot of
+confusion about which OS context represents what. So here's some examples:
+
+* In events reported from a Python/ASPNET/Rails web backend, the OS context
```
bruno-garcia, line 206:
```suggestion
* In events reported from a Python/ASP.NET/Rails web backend, the OS context
```

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -196,11 +196,44 @@ The `type` and default key is `"device"`.
...
+* In events reported from a JS web frontend, the SDK typically reports no OS
+  context.
+
+  The server however knows by looking at the platform (`"javascript"`) that the
+  incoming User-Agent can only come from the crashing device, and creates the
+  User-Agent based OS context under the default key `"os"`.
+
+To summarize:
+
+* `"os"` key for the crashing device's OS.
```
bruno-garcia, line 223:
```suggestion
* `"os"` key for the device generating the event.
```

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -196,11 +196,44 @@ The `type` and default key is `"device"`.
...
+* `"client_os"` key for an adjacent client device's OS (that is _not_ the
+  crashing device) if it's important. The Sentry server sets this as part of
+  User-Agent parsing, but SDKs can set this too.
```
bruno-garcia, line 226:
```suggestion
* `"client_os"` key for an adjacent client device's OS (that is _not_ the
  device creating the Sentry event instance) if it's important. The Sentry server sets this as part of
  User-Agent parsing, but SDKs can set this directly too.
```

bruno-garcia (approved):
Thanks Markus. Left some suggestions: Prefer "the device creating the event" as opposed to "crashing device" since on the case of server, we can be creating a transaction or message or 'error' event so the word 'crashing' here is a bit off

PR #560, opened 2022-04-19
Add a blurb about increasing memory for docker

barkbarkimashark:
Our internal docs recommends us to increase docker memory limits:
https://www.notion.so/sentry/Dev-Environment-98b18b1e96e04f018e6e3d5d87f2889e#c5daa8c7cae647d4893348b6e5eaf97b

We should probably do the same on the environments we setup for developing sentry.

### DIFF
--- a/src/docs/environment/index.mdx
+++ b/src/docs/environment/index.mdx
@@ -41,9 +41,12 @@ brew bundle --verbose
 ### Docker (Mac specific)
 
 <Alert title="Note" level="info">
-  It's recommended to increase the Docker memory limit to something higher than the default(2048MB).
+  It's recommended to increase the Docker memory limit to something higher than the default (2048MB).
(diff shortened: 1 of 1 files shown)


MeredithAnya:
Do you think it'd be reasonable to add a sentence for if people want to adjust the resources via docker desktop? (that's what I had done since it's relatively straightforward)

Maybe just like:
`On Docker Desktop, you can adjust the memory limits by going to Preferences > Resources > Memory`

barkbarkimashark:
> Do you think it'd be reasonable to add a sentence for if people want to adjust the resources via docker desktop? (that's what I had done since it's relatively straightforward)
> 
> Maybe just like: `On Docker Desktop, you can adjust the memory limits by going to Preferences > Resources > Memory`

Good call, just added!

PR #562, opened 2022-04-21
Scrap "type(scope): " from PR format

chadwhitacre:
We had some discussion about this a while ago and people seemed to agree that this is busywork that creates noise without adding much value. Can we overcome months of additional inertia? :)

**Old:** https://develop.sentry.dev/commit-messages/#type

**New:** https://develop-git-cwlw-scrap-types-and-scopes.sentry.dev/commit-messages/#commit-message-format


chadwhitacre:
This quickly devolved into a bikeshed in internal Slack (surprise surprise) so I am closing. I don't want that ~blood~ _paint_ on my hands.

PR #564, opened 2022-04-25
fix: Document min Sentry version for client reports

philipphofmann:
#skip-changelog


jan-auer (approved):
That's harsh :)

philipphofmann:
> That's harsh :)

Because I mention the bug in Relay?

PR #567, opened 2022-04-27
fix(Jira Server): Update docs link

ceorourke:
Update the link here to the correct one so that it actually goes to the intended anchor - it was redirecting to https://docs.sentry.io/product/integrations/issue-tracking/jira/


leeandher (approved):
Nice catch!

PR #568, opened 2022-04-28
Client reports get never rate limited

philipphofmann:
Document that client reports never get rate limited.

### DIFF
--- a/src/docs/sdk/client-reports.mdx
+++ b/src/docs/sdk/client-reports.mdx
@@ -8,8 +8,6 @@ about themselves to Sentry.  They are currently mainly used to emit outcomes
 for events that were never sent.  Chained relays are also able to emit these
 client reports to inform the next relay in chain about _some_ outcomes.
 
-It is guaranteed that client reports never get rate limited by the server.
-
 Due to a bug in Relay, which discards envelopes containing unknown envelope
 items, the minimum required version of Sentry for client reports is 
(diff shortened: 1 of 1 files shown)


jan-auer (approved):
Technically, this is incorrect. We're minimizing the possibility of client reports getting rate limited, there are no product rate limits, and as such clients can assume that client reports go through (based on the best-effort assumption). So, feel free to clarify or merge as-is, based on your understanding.

src/docs/sdk/client-reports.mdx
```diff
@@ -129,7 +129,8 @@ an envelope item, for example, with the response 429, the client SDK must not re
 this as the server already does. Still, the SDK must record lost envelope items
 when dropping them itself, for example, caused by an active rate limit. SDKs can put
 a simple optional check for HTTP status codes in place where any code `>= 400` except
-`429` will be recorded as `network_error`.
+`429` will be recorded as `network_error`. It is guaranteed that client reports never
```
AbhiPrasad, line 132: Can it still get rate-limited by ngnix?

philipphofmann:
Better now @jan-auer?

src/docs/sdk/client-reports.mdx
jan-auer: ^ this, exactly. those high-level limits are going to be increased more and more over time since we will try to minimize this from happening.

jan-auer (approved):
works for me :)  Thank you for the quick turnaround!

PR #573, opened 2022-05-03
ref(measurements): Link to Relay MetricUnit and update wording [INGEST-1131]

jan-auer:
Links to Relay's canonical `MetricUnit` documentation and updates the notice on custom units. Custom units are in fact supported in ingestion, but SDKs are encouraged not to expose them for the time being where possible. Also, there's now a range of information and fraction units in addition to durations.

This paragraph deliberately doesn't give SDK implementation advice since this document - like all payload docs - are more likely to be consumed by end users.

See https://github.com/getsentry/relay/pull/1256

### DIFF
--- a/src/docs/sdk/event-payloads/properties/measurements.mdx
+++ b/src/docs/sdk/event-payloads/properties/measurements.mdx
@@ -27,15 +27,14 @@ Standard measurement keys currently supported are from the following list taken 
 ]
 ```
 
-Custom measurements need units to be specified. Units currently supported are durations from the list `["ns" , "ms" , "s"]` taken from [here](https://github.com/getsentry/relay/blob/1e45a8817e45408ecbdbd3f0a679a5f5e4885290/relay-metrics/src/protocol.rs#L14-L21). Note that in the future we will extend the available units (for instance, memory units) but we will not implement a generic user-defined unit interface.
-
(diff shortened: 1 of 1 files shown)


src/docs/sdk/event-payloads/properties/measurements.mdx
```diff
@@ -27,8 +27,7 @@ Standard measurement keys currently supported are from the following list taken
 ]
 ```
 
-Custom measurements need units to be specified. Units currently supported are durations from the list `["ns" , "ms" , "s"]` taken from [here](https://github.com/getsentry/relay/blob/1e45a8817e45408ecbdbd3f0a679a5f5e4885290/relay-metrics/src/protocol.rs#L14-L21). Note that in the future we will extend the available units (for instance, memory units) but we will not implement a generic user-defined unit interface.
-
+Custom measurements need units to be specified. The full list of supported units is specified on [Relay's `MetricUnit`](https://getsentry.github.io/relay/relay_metrics/enum.MetricUnit.html). Sentry's event ingestion supports arbitrary custom units, but many SDKs will not expose a generic user-defined unit interface.
```
AbhiPrasad, line 30: We'll have to change the `measurements` json example below to use the full name for `ms`

AbhiPrasad (approved):

src/docs/sdk/event-payloads/properties/measurements.mdx
jan-auer: Thanks, updated!

PR #598, opened 2022-06-03
Removing Raven reference from Documentation

drguthals:
Since [Raven.js is deprecated](https://docs.sentry.io/clients/javascript/) should we be removing this reference for clarity?


drguthals:
NVM I see #598 already does this.

PR #609, opened 2022-06-17
Switch Forum link to GitHub discussions

edhgoose:
The forum is in read only mode. Therefore users should be sent to GitHub Discussions instead.

It looks like https://github.com/getsentry/develop/blob/c36c3ad4a58b6a4fab7d3c3ef07b4030f10572cc/src/components/sidebar.tsx#L298 might need to change too - but I'm not sure if that should be to the same place?


chadwhitacre:
Thanks @edhgoose. The flow is a bit different for self-hosted, I'll accept this PR and then follow up with new language.

cc: @drguthals

PR #620, opened 2022-06-27
fix(docs): Add new API Scopes for configuring Vercel Integration

leeandher:
Vercel recently introduced [API Scopes](https://vercel.com/changelog/enhanced-security-with-new-api-scopes-for-integrations) for integrations on their marketplace, meaning Self-Hosted users will need instructions on which scopes to set up when they create their own apps.

### DIFF
--- a/src/docs/integrations/vercel.mdx
+++ b/src/docs/integrations/vercel.mdx
@@ -6,20 +6,31 @@ title: "Vercel Integration"
 
 To use the Vercel integration you'll need to create your own integration in Vercel. To start, click "Create" in the [Integrations Developer Console](https://vercel.com/dashboard/integrations/console).
 
-When configuring the app:
-* Set the `Redirect URL` to `{YOUR_DOMAIN}/extensions/vercel/configure/`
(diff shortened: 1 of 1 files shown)


src/docs/integrations/vercel.mdx
```diff
@@ -6,20 +6,32 @@ title: "Vercel Integration"
...
+
+| API Scope                     | Value      |
+| ----------------------------- | ---------- |
+| Integration Configuration     | Read/Write |
+| Projects                      | Read/Write |
+| Project Environment Variables | Read/Write |
+| Teams                         | Read       |
+| Current User                  | Read       |
+
+- Set the `Webhook URL` to `{YOUR_DOMAIN}/extensions/vercel/delete/` and subscribe to the following events:
```
ceorourke, line 22: This will change after we finish up that old webhook stuff, right? We want to use the "generic webhook" instead of the delete one, so maybe this PR should hang out for a bit until that change lands.

AniketDas-Tekky (approved):

src/docs/integrations/vercel.mdx
leeandher: That is true, I don't mind opening a separate PR for that though because currently, users integrating with Sentry will see these scopes and not know which to grant.

PR #622, opened 2022-06-28
Document self-hosted installation behind a proxy

chadwhitacre:
Here's the documentation to go with https://github.com/getsentry/self-hosted/pull/1543. I pulled it from https://github.com/getsentry/self-hosted/commit/ee2be4623417f155da35baccd426bc62627d37ca and edited it somewhat.

cc: @goganchic @BYK @aminvakil

### DIFF
--- a/src/docs/self-hosted/index.mdx
+++ b/src/docs/self-hosted/index.mdx
@@ -10,7 +10,7 @@ Our recommendation is to download the [latest release of the self-hosted reposi
 
 ### Installing Behind a Proxy
 
-In some enterprise setups there is no direct Internet connection, you must use an HTTP proxy server. How do you install Sentry in this environment? Let us assume: 
+In some enterprise setups there is no direct Internet connection, so you must use an HTTP proxy server. How do you install Sentry in this environment? Let us assume: 
 
 1. Your Sentry installation is running on Linux.
 1. `http://proxy:3128` is your proxy address.
--- a/src/docs/self-hosted/index.mdx
+++ b/src/docs/self-hosted/index.mdx
@@ -8,6 +8,42 @@ In addition to making its source code available publicly, Sentry offers and main
 
 Our recommendation is to download the [latest release of the self-hosted repository](https://github.com/getsentry/self-hosted/releases/latest), and then run `./install.sh` inside this directory. This script will take care of all the things you need to get started, including a base-line configuration, and then will tell you to run `docker-compose up -d` to start Sentry. Sentry binds to port `9000` by default. You should be able to reach the login page at [http://127.0.0.1:9000](http://127.0.0.1:9000/).
 
+### Installing Behind a Proxy
+
+In some enterprise setups there is no direct Internet connection, you must use an HTTP proxy server. How do you install Sentry in this environment? Let us assume: 
+
+1. Your Sentry installation is running on Linux.
+1. `http://proxy:3128` is your proxy address.
+1. `127.0.0.0/8` is the only network that should be accessed without a proxy.
+
+Here are the steps to follow:
+
+1. Set `http_proxy`, `https_proxy` and `no_proxy` variables in the `/etc/environment` file.
+2. To make the `docker pull` command respect your proxy settings, create a `/etc/systemd/system/docker.service.d/http-proxy.conf` file with these contents:
+```
+[Service]
+Environment="HTTP_PROXY=http://proxy:3128"
+Environment="HTTPS_PROXY=http://proxy:3128"
+Environment="NO_PROXY=127.0.0.0/8"
+```
+3. Run `systemctl daemon-reload` and restart Docker with `systemctl restart docker.service`.
+4. To add your proxy environment variables into Sentry's Docker containers, create `~/.docker/config.json` with these contents:
+```json
+{
+ "proxies":
+ {
+   "default":
+   {
+     "httpProxy": "http://proxy:3128",
+     "httpsProxy": "http://proxy:3128",
+     "noProxy": "127.0.0.0/8"
+   }
+ }
+}
+```
+
+From there you can run `./install.sh` like usual.
+
 ## Configuration
 
 You very likely will want to adjust the default configuration for Sentry. These facilities are available for that purpose:


chadwhitacre:
I considered putting this in "Releases & Upgrades" but it's more about installation. I also considered giving it a separate page like we do for other configs, but it's really about install-time vs. run-time config.

src/docs/self-hosted/index.mdx
```diff
@@ -8,6 +8,42 @@ In addition to making its source code available publicly, Sentry offers and main
 
 Our recommendation is to download the [latest release of the self-hosted repository](https://github.com/getsentry/self-hosted/releases/latest), and then run `./install.sh` inside this directory. This script will take care of all the things you need to get started, including a base-line configuration, and then will tell you to run `docker-compose up -d` to start Sentry. Sentry binds to port `9000` by default. You should be able to reach the login page at [http://127.0.0.1:9000](http://127.0.0.1:9000/).
 
+### Installing Behind a Proxy
+
+In some enterprise setups there is no direct Internet connection, you must use an HTTP proxy server. How do you install Sentry in this environment? Let us assume:
```
ethanhs, line 13:
```suggestion
In some enterprise setups there is no direct Internet connection, so you must use an HTTP proxy server. How do you install Sentry in this environment? Let us assume: 
```

src/docs/self-hosted/index.mdx
```diff
@@ -8,6 +8,42 @@ In addition to making its source code available publicly, Sentry offers and main
...
+1. `127.0.0.0/8` is the only network that should be accessed without a proxy.
+
+Here are the steps to follow:
+
+1. Set `http_proxy`, `https_proxy` and `no_proxy` variables in the `/etc/environment` file.
+2. To make the `docker pull` command respect your proxy settings, create a `/etc/systemd/system/docker.service.d/http-proxy.conf` file with these contents:
+```
+[Service]
+Environment="HTTP_PROXY=http://proxy:3128"
+Environment="HTTPS_PROXY=http://proxy:3128"
```
ethanhs, line 26:
```suggestion
Environment="HTTPS_PROXY=https://proxy:3128"
```

I definitely could be mistaken, but I believe `HTTPS_PROXY` *must* be set to an `https` URL.

src/docs/self-hosted/index.mdx
```diff
@@ -8,6 +8,42 @@ In addition to making its source code available publicly, Sentry offers and main
...
+3. Run `systemctl daemon-reload` and restart Docker with `systemctl restart docker.service`.
+4. To add your proxy environment variables into Sentry's Docker containers, create `~/.docker/config.json` with these contents:
+```json
+{
+ "proxies":
+ {
+   "default":
+   {
+     "httpProxy": "http://proxy:3128",
+     "httpsProxy": "http://proxy:3128",
```
ethanhs, line 38:
```suggestion
     "httpsProxy": "https://proxy:3128",
```

ethanhs (approved):
The text and content LGTM. I made 3 minor suggestions.

BYK (approved):
LGTM

src/docs/self-hosted/index.mdx
goganchic: If I'm not mistaken `HTTP_PROXY` variable is used for HTTP requests and `HTTPS_PROXY` - for HTTPS requests (found in [curl man page](https://curl.se/docs/manpage.html)). So it's possible to use http/https proto for HTTP_PROXY and for HTTPS_PROXY.

ethanhs: Yup, that definitely seems to be correct, I will resolve these reviews then.

chadwhitacre: Hmmm ... reading the curl docs I see "http_proxy is an exception as it is only available in lower case." Do we need to change that here? Maybe suggest setting both?

ethanhs: It seems according to the [docker docs](https://docs.docker.com/network/proxy/#set-the-environment-variables-manually) that the upper case usage is correct.

chadwhitacre:
Merged! Thanks for the contribution @goganchic!

src/docs/self-hosted/index.mdx
```diff
@@ -8,6 +8,42 @@ In addition to making its source code available publicly, Sentry offers and main
...
+
+In some enterprise setups there is no direct Internet connection, so you must use an HTTP proxy server. How do you install Sentry in this environment? Let us assume: 
+
+1. Your Sentry installation is running on Linux.
+1. `http://proxy:3128` is your proxy address.
+1. `127.0.0.0/8` is the only network that should be accessed without a proxy.
+
+Here are the steps to follow:
+
+1. Set `http_proxy`, `https_proxy` and `no_proxy` variables in the `/etc/environment` file.
```
aminvakil, line 21: Is this really a necessity?

We're using http proxy in some of our servers which are using docker and we have only changed docker systemd file.

aminvakil (commented):
Sorry, I've been late to party!

src/docs/self-hosted/index.mdx
goganchic: When you change systemd file - it makes docker daemon to use proxy for `docker pull`. When you config proxy in `.docker/config.json` - it makes docker to set env proxy variables during `docker run`. `/etc/environment` file is required for `docker build`, so without it [this code](https://github.com/getsentry/self-hosted/blob/e028a5ed52f835ef2fefcab0104ff047889e5470/cron/Dockerfile#L4) fails.

aminvakil: Ah, I didn't know that `docker build` is different, thanks!

https://github.com/getsentry/self-hosted/blob/e028a5ed52f835ef2fefcab0104ff047889e5470/install/build-docker-images.sh#L6 :
`$dc build --build-arg "http_proxy=${http_proxy:-}" --build-arg "https_proxy=${https_proxy:-}" --build-arg "no_proxy=${no_proxy:-}" --force-rm web`

So if changing `/etc/environment` is necessary why are we passing these build arguments to `docker build` and if we are passing these build arguments to `docker build` why we need to change `/etc/environment`?

aminvakil: Ah, I didn't know that `docker build` is different, thanks!

https://github.com/getsentry/self-hosted/blob/e028a5ed52f835ef2fefcab0104ff047889e5470/install/build-docker-images.sh#L6 :
`$dc build --build-arg "http_proxy=${http_proxy:-}" --build-arg "https_proxy=${https_proxy:-}" --build-arg "no_proxy=${no_proxy:-}" --force-rm web`

So if changing `/etc/environment` is necessary why are we passing these build arguments to `docker build` and if we are passing these build arguments to `docker build` why we need to change `/etc/environment`?

goganchic: `/etc/environment` provide env variables for shell and the command `$dc build --build-arg "http_proxy=${http_proxy:-}" --build-arg "https_proxy=${https_proxy:-}" --build-arg "no_proxy=${no_proxy:-}" --force-rm web` passes this variables to `docker build`. But actually we can remove all `--build-arg` from `docker build` and remove proxy args from https://github.com/getsentry/self-hosted/blob/e028a5ed52f835ef2fefcab0104ff047889e5470/install/install-wal2json.sh#L10, because `./docker/config.json` does all required things. Actually we don't need `/etc/environment`, we need only `~/.docker/config.json` and `/etc/systemd/system/docker.service.d/http-proxy.conf`.

goganchic: https://github.com/getsentry/develop/pull/628

PR #623, opened 2022-06-29
fix(gitlab): Update config docs

ceorourke:
Remove mention of "expire access tokens" checkbox as it no longer exists, and the "confidential" box is checked by default so just say to make sure it's checked. 

See: https://github.com/getsentry/sentry/pull/36234


Nadia740 (changes requested):
[]()

PR #624, opened 2022-06-30
feat(transactions): Add transaction_info.source [INGEST-1427]

jjbayer:
Add a `source` label to every transaction that indicates how the transaction name was generated, as decided in https://www.notion.so/sentry/Annotating-Transaction-Names-a08a05ff97b348b0903de03a17baa731.

Rendered version is [here](https://develop-git-feat-transaction-source.sentry.dev/sdk/event-payloads/properties/transaction_info/) and will be included on the "Transaction Type" page.

### DIFF
--- a/src/docs/sdk/event-payloads/properties/transaction_info.mdx
+++ b/src/docs/sdk/event-payloads/properties/transaction_info.mdx
@@ -18,11 +18,11 @@ from the transaction name, or replace the entire name with a placeholder.
 
 | Source             | Description                                                                                                                     | <div style="width:220px">Examples</div>                |
 | ------------------ | ------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------ |
-| `raw-url`          | Raw URL, potentially containing identifiers.                                                                                    | `/auth/login/john123/`<br />`GET /auth/login/john123/` |
-| `raw-url-fallback` | The SDK is configured to send `url-pattern`, but has to fall back to a raw URL for this particular transaction.                 |
-| `url-pattern`      | URL pattern / route                                                                                                             | `/auth/login/:userId/`<br />`GET /auth/login/{user}/`  |
-| `view`             | Name of the view handling the request.                                                                                          | `UserListView`                                         |
 | `custom`           | User-defined name, see [`setTransactionName()`](https://docs.sentry.io/platforms/javascript/enriching-events/transaction-name/) | `my_transaction`                                       |
+| `url`              | Raw URL, potentially containing identifiers.                                                                                    | `/auth/login/john123/`<br />`GET /auth/login/john123/` |
+| `raw-url-fallback` | The SDK is configured to send a `route`, but has to fall back to a raw URL for this particular transaction.                     |
+| `route`            | Parametrized URL / route                                                                                                        | `/auth/login/:userId/`<br />`GET /auth/login/{user}/`  |
+| `view`             | Name of the view handling the request.                                                                                          | `UserListView`                                         |
 | `unknown`          | This is the default value set by Relay for legacy SDKs.                                                                         |                                                        |
 | `component`        | Named after a software component, such as a function or class name.                                                             | `AuthLogin.login`<br /> `LoginActivity.login_button`   |
-| `background-task`  | Name of a background task (e.g. a Celery task)                                                                                  | `sentry.tasks.do_something`                            |
+| `task`             | Name of a background task (e.g. a Celery task)                                                                                  | `sentry.tasks.do_something`                            |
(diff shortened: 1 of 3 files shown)


src/docs/sdk/event-payloads/properties/transaction_info.mdx
```diff
@@ -0,0 +1,28 @@
...
+`transaction_info.source`
+
+: _Required_. Contains information about how the name of the transaction was determined.
+This will be used by the server to decide whether or not to scrub identifiers
+from the transaction name, or replace the entire name with a placeholder.
+
+| Source             | Description                                                                                                                     | <div style="width:220px">Examples</div>                |
+| ------------------ | ------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------ |
+| `raw-url`          | Raw URL, potentially containing identifiers.                                                                                    | `/auth/login/john123/`<br />`GET /auth/login/john123/` |
+| `raw-url-fallback` | The SDK is configured to send `url-pattern`, but has to fall back to a raw URL for this particular transaction.                 |
```
jjbayer, line 22: I figured that for data analysis, it makes sense to distinguish between the case where the SDK is configured to send the raw URL, and the case where it is configured to send a parametrized URL but has to fall back to a raw URL instead. @AbhiPrasad what do you think, can these two cases even be distinguished in the JS SDK?

src/docs/sdk/event-payloads/properties/transaction_info.mdx
```diff
@@ -0,0 +1,28 @@
...
+| `url-pattern`      | URL pattern / route                                                                                                             | `/auth/login/:userId/`<br />`GET /auth/login/{user}/`  |
```
jan-auer, line 23: Could we simplify some of these names, for instance this one to `route` and `raw-url` to just `url`?

src/docs/sdk/event-payloads/properties/transaction_info.mdx
jan-auer: I agree that we probably do not need to distinguish these cases. The actual source of this transaction name is the URL, and it's irrelevant whether it was applied as a fallback, as part of a multi-step process, or directly.

src/docs/sdk/event-payloads/properties/transaction_info.mdx
```diff
@@ -0,0 +1,28 @@
...
+| `view`             | Name of the view handling the request.                                                                                          | `UserListView`                                         |
+| `custom`           | User-defined name, see [`setTransactionName()`](https://docs.sentry.io/platforms/javascript/enriching-events/transaction-name/) | `my_transaction`                                       |
```
jan-auer, line 25: Would move this to the top to promote its significance. This is the most important one to get right in the implementation.

jan-auer (commented):
Should we also update other places, such as:

- [Unified API](https://develop.sentry.dev/sdk/unified-api/): add the source as internal second parameter
- [Guidelines for Performance Monitoring](https://develop.sentry.dev/sdk/performance/): Mention that the transaction source should be set.

src/docs/sdk/event-payloads/properties/transaction_info.mdx
jjbayer: Sure why not

jjbayer:
> Should we also update other places, such as:
> 
> * [Unified API](https://develop.sentry.dev/sdk/unified-api/): add the source as internal second parameter
> * [Guidelines for Performance Monitoring](https://develop.sentry.dev/sdk/performance/): Mention that the transaction source should be set.

@jan-auer I added a reference in "Guidelines for PM". Not sure about the Unified API, do you mean as a second parameter to `set_transaction()`? Would this not automatically become public API? I assume we want to prevent the user from setting the `source` themselves.

src/docs/sdk/event-payloads/properties/transaction_info.mdx
AbhiPrasad: It's feasible to distinguish between them, but I think it's also unnecessary (and a waste of bundle size) - let's stick with just the `url` source!

src/docs/sdk/event-payloads/properties/transaction_info.mdx
```diff
@@ -0,0 +1,28 @@
+`transaction_info`
+
+: _Required_. Additional information about the name of the transaction.
```
untitaker, line 3: hmm, is it required? pretty sure it can't/won't be

untitaker (approved):
I think we should add this to the public docs, after all SDKs need to know how to send this

src/docs/sdk/event-payloads/properties/transaction_info.mdx
jan-auer: Alright, removing url-fallback then.

jan-auer (approved):

PR #628, opened 2022-07-03
Improve self-hosted docs about installation behind the proxy

goganchic:
necessary code changes: https://github.com/getsentry/self-hosted/pull/1560

### DIFF
--- a/src/docs/self-hosted/index.mdx
+++ b/src/docs/self-hosted/index.mdx
@@ -10,7 +10,7 @@ Our recommendation is to download the [latest release of the self-hosted reposi
 
 ### Installing Behind a Proxy
 
-In some enterprise setups there is no direct Internet connection, so you must use an HTTP proxy server. How do you install Sentry in this environment? Let us assume: 
+In some enterprise setups there is no direct Internet connection, so you must use an HTTP proxy server. How do you install Sentry in this environment? Let us assume:
 
 1. Your Sentry installation is running on Linux.
 1. `http://proxy:3128` is your proxy address.
@@ -24,7 +24,7 @@ Here are the steps to follow:
 [Service]
 Environment="HTTP_PROXY=http://proxy:3128"
 Environment="HTTPS_PROXY=http://proxy:3128"
-Environment="NO_PROXY=127.0.0.0/8"
+Environment="NO_PROXY=relay,web,sentry,127.0.0.0/8"
 ```
 3. Run `systemctl daemon-reload` and restart Docker with `systemctl restart docker.service`.
 4. To add your proxy environment variables into Sentry's Docker containers, create `~/.docker/config.json` with these contents:
@@ -36,7 +36,7 @@ Environment="NO_PROXY=127.0.0.0/8"
    {
      "httpProxy": "http://proxy:3128",
      "httpsProxy": "http://proxy:3128",
-     "noProxy": "127.0.0.0/8"
+     "noProxy": "relay,web,sentry,127.0.0.0/8"
    }
  }
 }
--- a/src/docs/self-hosted/index.mdx
+++ b/src/docs/self-hosted/index.mdx
@@ -30,7 +30,7 @@ Environment="NO_PROXY=127.0.0.0/8"
 
 From there you can run `./install.sh` like usual.
 
-By default Sentry sends anonymous usage statistics to the Sentry team. It helps to improve the product. To make this feature work it's required to allow Docker containers to access the Internet. To add your proxy environment variables into Sentry's Docker containers, create `~/.docker/config.json` with these contents:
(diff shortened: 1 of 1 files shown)


src/docs/self-hosted/index.mdx
```diff
@@ -36,7 +35,7 @@ Environment="NO_PROXY=127.0.0.0/8"
    {
      "httpProxy": "http://proxy:3128",
      "httpsProxy": "http://proxy:3128",
-     "noProxy": "127.0.0.0/8"
+     "noProxy": "relay,web,sentry,127.0.0.0/8"
```
aminvakil, line 38: Have you tested a clean working installation with these variables? Are these enough? We have `kafka:9092` in `relay/config.yml` for example, shouldn't we also add `kafka` to `noProxy`?

aminvakil (commented):
Thank you for keeping to improve docs!

src/docs/self-hosted/index.mdx
goganchic: I'd removed all docker containers, images, volumes and networks and reinstalled Sentry one more time, checked web, created test event. I'd found no errors in docker logs.

goganchic: Please wait a bit, I need to recheck it one more time.

goganchic: From the one hand proxies from ~/.docker/config.json work only for docker run command and do not work for docker build. It is written in the [docker docs](https://docs.docker.com/network/proxy/):

> When you create or start new containers, the environment variables are set automatically within the container.

From the other hand we need a connection to the Internet only during build stage. So we do not need `~/.docker/config.json` at all. We need to set correct env variables through `/etc/environment` and that's it. In this commit https://github.com/getsentry/self-hosted/commit/2672ce545227f9a40b90178006c5ab25f7860a64 we do not need changes in `/etc/apt/apt.conf`, because all required variables will be passed through `/etc/environment`.

Sorry for my mistake.

goganchic: @aminvakil  I'd updated PR, please take a look.

src/docs/self-hosted/index.mdx
```diff
@@ -10,11 +10,11 @@ Our recommendation is to download the [latest release of the self-hosted reposi
...
 ### Installing Behind a Proxy
 
-In some enterprise setups there is no direct Internet connection, so you must use an HTTP proxy server. How do you install Sentry in this environment? Let us assume: 
+In some enterprise setups there is no direct Internet connection, so you must use an HTTP proxy server. How do you install Sentry in this environment? Let us assume:
 
 1. Your Sentry installation is running on Linux.
-1. `http://proxy:3128` is your proxy address.
-1. `127.0.0.0/8` is the only network that should be accessed without a proxy.
+2. `http://proxy:3128` is your proxy address.
+3. `127.0.0.0/8` is the only network that should be accessed without a proxy.
```
chadwhitacre, line 17: Fwiw, Markdown does the right thing with these even if they're all the same number.

chadwhitacre:
Catching up on the threads ... what's our status here? https://github.com/getsentry/self-hosted/pull/1560 is closed, do we still want to remove step (4) of the docs with this PR?

goganchic:
Yes. Before https://github.com/getsentry/self-hosted/pull/1560 I thought that step 1 is redundant, but actually it's required and step 4 is not.

chadwhitacre:
I had the same thought. Not everyone will care to do that but it would be good to note in a comment, yes.

Thanks! :)

goganchic:
@chadwhitacre I've added the information about the configuration access for Sentry Beacon.

src/docs/self-hosted/index.mdx
```diff
@@ -27,7 +27,11 @@ Environment="HTTPS_PROXY=http://proxy:3128"
 Environment="NO_PROXY=127.0.0.0/8"
 ```
 3. Run `systemctl daemon-reload` and restart Docker with `systemctl restart docker.service`.
-4. To add your proxy environment variables into Sentry's Docker containers, create `~/.docker/config.json` with these contents:
+
+From there you can run `./install.sh` like usual.
+
+By default Sentry sends anonymous usage statistics to the Sentry team. It helps to improve the product. To make this feature work it's required to allow Docker containers to access the Internet. To add your proxy environment variables into Sentry's Docker containers, create `~/.docker/config.json` with these contents:
```
aminvakil, line 33: From this sentence it seems like sentry containers only want to access the Internet for "anonymous usage statistics", is it so?

Let's say if users want to integrate their sentry with a third party service like slack, I think they need to use proxy there as well.

This should be added to the sentence too IMHO.

goganchic: fair enough

aminvakil (approved):
LGTM!

@goganchic Thanks for improving docs!

PR #632, opened 2022-07-07
docs: Help note for Firefox users

armenzg:
Some Firefox users may use a master password and the process to create a local certificate will prompt for such password.

Unfortunately, the wording is not clear and this note should help.

### DIFF
--- a/src/docs/frontend/development-server.mdx
+++ b/src/docs/frontend/development-server.mdx
@@ -34,4 +34,4 @@ You can now run the dev server with `yarn dev-ui` and open [https://localhost:79
 
 ![lock icon in address bar](../img/addressBarLockIcon.png)
 
-
+**NOTE**: For Firefox users that use the master password you will be prompted for it with this message: "Enter Password or Pin for "NSS Certificate DB":"


asottile-sentry (approved):

billyvg (approved):

gitressa:
Thanks for working on this @armenzg. It's good that the tip is added on https://develop.sentry.dev/frontend/development-server/.

I just ran into the cryptic `Enter Password or Pin for "NSS Certificate DB":` message, after installing mkcert from https://github.com/FiloSottile/mkcert/.

mkcert-Enter-Password-Pin-NSS-Certificate-DB

Do you think it could be added in the [README.md](https://github.com/FiloSottile/mkcert/blob/master/README.md) as well, since that was my entry point and not develop.sentry.dev?

armenzg:
> Thanks for working on this @armenzg. It's good that the tip is added on https://develop.sentry.dev/frontend/development-server/.
> 
> I just ran into the cryptic `Enter Password or Pin for "NSS Certificate DB":` message, after installing mkcert from https://github.com/FiloSottile/mkcert/.
> 
> mkcert-Enter-Password-Pin-NSS-Certificate-DB
> 
> Do you think it could be added in the [README.md](https://github.com/FiloSottile/mkcert/blob/master/README.md) as well, since that was my entry point and not develop.sentry.dev?

Possibly! Would you want to open a PR with them?
I think [this issue](https://github.com/FiloSottile/mkcert/issues/50) would be related to adding it.

gitressa:
I totally overlooked it, but there's already a MR in progress: https://github.com/FiloSottile/mkcert/pull/100.
Have a great day!

Issue #634, opened 2022-07-07
Including `transaction` field in Dynamic Sampling Context must depend on transaction source


vladanpaunovic:
In some SDKs (mainly JS Browser) we run into the issue of having "low quality" transaction names. These are transaction names that are basically useless to customers because they mess up grouping and by extension will mess up the sampling rules for dynamic sampling.

"Low quality transactions" are transactions that have a high cardinality in nature which boils down to URLs as transaction names (e.g. `"example.com/resources/1398074123749127834/container/879872937819"`). In order to filter these bad transactions names out, we only wanna include only low cardinality ones, i.e. parameterized or named in some other way (e.g. `"example.com/resources/:resouceId/container/:containerId"`).

This all should be documented.

PR #635, opened 2022-07-08
Add transaction back to DSC specification

lforst:
Resolves https://github.com/getsentry/develop/issues/634

### DIFF
--- a/src/docs/sdk/performance/dynamic-sampling-context.mdx
+++ b/src/docs/sdk/performance/dynamic-sampling-context.mdx
@@ -160,7 +160,7 @@ SDKs may use the following keys to set entries on `baggage` HTTP headers:
 - `sentry-release`
 - `sentry-environment`
 - `sentry-user_segment`
-- `sentry-transaction` (**only inclue iff name has [good quality](#note-on-low-quality-transaction-names)**)
+- `sentry-transaction` (**only include if name has [good quality](#note-on-low-quality-transaction-names)**)
 
 SDKs must set all of the keys in the form of "`sentry-[name]`".
 The prefix "`sentry-`" acts to identify key-value pairs set by Sentry SDKs.
--- a/src/docs/sdk/performance/dynamic-sampling-context.mdx
+++ b/src/docs/sdk/performance/dynamic-sampling-context.mdx
@@ -145,7 +145,7 @@ The value of this envelope header is a JSON object with the following fields:
 - `release` (string) - The release name as specified in client options`.
 - `environment` (string) - The environment name as specified in client options.
 - `user_segment` (string) - User segment as set by the user with <Link to="/sdk/unified-api/#scope">`scope.set_user`</Link>.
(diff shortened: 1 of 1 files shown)


Lms24 (approved):
I think this is a good general description for "good quality" transaction names. 

Wondering if the dependence for `transaction` being added to the DSC on the transaction source should be a hard requirement instead of a suggestion. But as long as Relay doesn't make sampling decisions based on the source (which IIRC and according to the spec) I think we can leave this open to SDK implementors.

lforst:
> Wondering if the dependence for `transaction` being added to the DSC on the transaction source should be a hard requirement instead of a suggestion. But as long as Relay doesn't make sampling decisions based on the source (which IIRC and according to the spec) I think we can leave this open to SDK implementors.

The reason I wrote it like this and also made it a hard-requirement to have the transaction on the DSC based on transaction source, is because not all SDKs have the problem of "low quality transaction names" like the JS SDK with URLs and additionally not all SDKs will have to provide a transaction source until relay makes it required.
In my head, transaction source and inclusion of TXs on DSC are two different concepts, but the one can support the decision in the other one.

untitaker (approved):

src/docs/sdk/performance/dynamic-sampling-context.mdx
```diff
@@ -105,7 +105,30 @@ After the DSC of a particular trace has been frozen, API calls like `set_user` s
...
 
-All of the values in the payloads below are required (non-optional) in a sense, that when they are known to an SDK at the time a transaction envelope is sent to Sentry, or at the time a baggage header is propagated, they must also be included in said envelope or baggage.
+All of the values in the payload schemas below are required (non-optional) in a sense, that when they are known to an SDK at the time a transaction envelope is sent to Sentry, or at the time a baggage header is propagated, they must also be included in said envelope or baggage.
+
+<Alert level="warning">
+
+### Note on "low-quality" transaction names:
+
+UX wise for the Dynamic Sampling product, we depend on transaction names (i.e. the `transaction` parameter of the DSC) to have "good quality".
+**Good quality transaction names** are descriptive, have proper grouping on Sentry, have low cardinality, and do not contain PII.
```
jjbayer, line 115:
```suggestion
**Good quality transaction names** are descriptive, have proper grouping on Sentry, have low cardinality, and do not contain PII or other identifiers.
```

src/docs/sdk/performance/dynamic-sampling-context.mdx
```diff
@@ -136,6 +160,7 @@ SDKs may use the following keys to set entries on `baggage` HTTP headers:
 - `sentry-release`
 - `sentry-environment`
 - `sentry-user_segment`
+- `sentry-transaction` (**only inclue iff name has [good quality](#note-on-low-quality-transaction-names)**)
```
romtsn, line 163:
```suggestion
- `sentry-transaction` (**only include if name has [good quality](#note-on-low-quality-transaction-names)**)
```

PR #640, opened 2022-07-13
Add docs for migration testing [INC-160 INGEST-1436]

jjbayer:
Document testing migrations with the `TestMigrations` base class.

Preview of the relevant section is here: https://develop-git-test-migrations-testing.sentry.dev/database-migrations/#testing

### DIFF
--- a/src/docs/database-migrations.mdx
+++ b/src/docs/database-migrations.mdx
@@ -60,6 +60,33 @@ Always commit the changes to `migrations_lockfile.txt` with your migration.
 
 There are some things we need to be careful about when running migrations.
 
+### Testing
+
+Database migrations are risky operations that can lead to irreversible data loss or corruption. This is especially true for [data migrations](https://docs.djangoproject.com/en/4.0/topics/migrations/#data-migrations).
+For this reason, every migration should have a corresponding integration test.
+
+To test your migration, derive a test case from `TestMigrations` and add it to `tests/sentry/migrations`.
+For example:
+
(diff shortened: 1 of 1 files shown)


src/docs/database-migrations.mdx
```diff
@@ -60,6 +60,33 @@ Always commit the changes to `migrations_lockfile.txt` with your migration.
 
 There are some things we need to be careful about when running migrations.
 
+### Testing
```
jjbayer, line 63: I put this at the top of the "Guidlines" section because it seems like the most important guideline to me, but I'm open to moving it further down.

wedamija (approved):
Looks good to me. A couple of other things that we might mention:
 - These tests can be a little brittle due to some interactions with the `django-pg-zero-downtime-migrations` package. If old tests start breaking, it can be worthwhile to just delete them rather than spending the time to fix.
 - I've added a `setup_initial_state` method that runs before any migration rollbacks are run. This can be useful if you want to use the `self.create_*` helper methods to create complex objects. Ideally, things should just be created in `setup_before_migration` though.
 - Objects created in `setup_before_migration` should only be created via `apps.get_model`. Using the `self.create_*` helpers here can cause test breakages if a model has fields added or removed.

jjbayer:
> Looks good to me. A couple of other things that we might mention:

@wedamija thanks, I added a `Notes` section:
image

wedamija (approved):

PR #646, opened 2022-07-14
Fix: fix typo

dashed:


AbhiPrasad (approved):
Yeah I think exclusive or is more semantically correct - but this makes it more clear

Issue #636, opened 2022-07-11
(DEV-DOCS) Control Sending Trace Data


smeubank:
### Problem statement

SDK developers should have a specification 
which guides them on the purpose and direction for implementing the option agreed upon to control sending trace data to outgoing requests, 
so that Sentry customers can control where data is sent and how traces are constructed from individual web applications using Sentry SDKs 

### Reference
[DACI Control Sending Trace Data](https://www.notion.so/sentry/DACI-Control-Sending-Trace-Data-63fd04ea885443ee99514c15a1e91499)

PR #648, opened 2022-07-18
Add spec for tracePropagationTargets SDK Option

Lms24:
This PR adds the spec for the new `tracePropagationTargets` SDK option. It is the result of the discussions we had about  giving users the ability to control where trace data is being propagated to (see [DACI Control Sending Trace Data](https://www.notion.so/sentry/DACI-Control-Sending-Trace-Data-63fd04ea885443ee99514c15a1e91499)).

resolves: #636

### DIFF
--- a/src/docs/sdk/performance/dynamic-sampling-context.mdx
+++ b/src/docs/sdk/performance/dynamic-sampling-context.mdx
@@ -59,7 +59,7 @@ Baggage is a standard HTTP header with URI encoded key-value pairs.
 For the propagation of DSC, SDKs first read the DSC from the baggage header of incoming requests/messages.
 To propagate DSC to downstream SDKs/services, we create a baggage header (or modify an existing one) through HTTP request instrumentation.
 
-The `baggage` header should only be attached to an outgoing request, if the request's URL matches at least one entry of the [`tracePropagationTargets`](/sdk/performance/#tracepropagationtargets) SDK option (or its fallback default behaviour).
+The `baggage` header should only be attached to an outgoing request if the request's URL matches at least one entry of the [`tracePropagationTargets`](/sdk/performance/#tracepropagationtargets) SDK option.
 
 <Alert level="info">
 
--- a/src/docs/sdk/performance/dynamic-sampling-context.mdx
+++ b/src/docs/sdk/performance/dynamic-sampling-context.mdx
@@ -59,6 +59,8 @@ Baggage is a standard HTTP header with URI encoded key-value pairs.
 For the propagation of DSC, SDKs first read the DSC from the baggage header of incoming requests/messages.
 To propagate DSC to downstream SDKs/services, we create a baggage header (or modify an existing one) through HTTP request instrumentation.
 
+The `baggage` header should only be attached to an outgoing request, if the request's URL matches at least one entry of the [`tracePropagationTargets`](#tracepropagationtargets) SDK option (or its fallback default behaviour).
+
 <Alert level="info">
 
 Other vendors might also be using the `baggage` header.
--- a/src/docs/sdk/performance/dynamic-sampling-context.mdx
+++ b/src/docs/sdk/performance/dynamic-sampling-context.mdx
@@ -59,7 +59,7 @@ Baggage is a standard HTTP header with URI encoded key-value pairs.
 For the propagation of DSC, SDKs first read the DSC from the baggage header of incoming requests/messages.
 To propagate DSC to downstream SDKs/services, we create a baggage header (or modify an existing one) through HTTP request instrumentation.
 
-The `baggage` header should only be attached to an outgoing request, if the request's URL matches at least one entry of the [`tracePropagationTargets`](#tracepropagationtargets) SDK option (or its fallback default behaviour).
+The `baggage` header should only be attached to an outgoing request, if the request's URL matches at least one entry of the [`tracePropagationTargets`](/sdk/performance/#tracepropagationtargets) SDK option (or its fallback default behaviour).
 
 <Alert level="info">
 
--- a/src/docs/sdk/performance/index.mdx
+++ b/src/docs/sdk/performance/index.mdx
@@ -42,25 +42,23 @@ For example, users can specify this property to keep trace propagation within th
 
 This option takes an array of strings and/or regular expressions, and SDKs should only add trace headers to an outgoing request if the request's URL matches (or, in the case of string literals, contains) at least one of the items from the array.
 
+SDKs may choose a default value which makes sense for their use case.
+Most SDKs default to the regex `/.*/` (meaning they attach headers to all outgoing requests), but deviation is allowed if necessary.
+For example, because of CORS, browser-based SDKs default to only adding headers to domain-internal requests.
+
 See [`sentry-trace`](#header-sentry-trace) and [`baggage`](/sdk/performance/dynamic-sampling-context/#baggage) for more details on the individual headers which are attached to outgoing requests.
 
 #### Example
 
+The following example shows which URLs of outgoing requests would (not) match a given `tracePropagationTargets` array.
+
 ```JS
-// Entries can be strings or RegEx
+// Entries can be strings or regex
 tracePropagationTargets: ['localhost', /^\// ,/myApi.com\/v[2-4]/]
 
 URLs matching: 'localhost:8443/api/users', '/api/envelopes', 'myApi.com/v2/projects'
 URLs not matching: 'someHost.com/data', 'myApi.com/v1/projects'
 ```
-
-#### Default Behaviour
-
-If the `tracePropagationTargets` option is not specified by users, SDKs should fall back to their default behaviour for trace propagation.
-This default behaviour is not standardized and can vary across SDKs.
-For instance, Server-side SDKs may propagate trace headers to all outgoing requests by default while client-side SDKs (e.g. Javascript) can fall back to a default array of allowed domains (e.g. `localhost` and `/^\//` in JavaScript).
-Choosing a sensible default behaviour is up to the SDK maintainers, however, special limitations, such as CORS issues for frontend web SDKs must be considered.
-SDKs that previously supported the `tracingOrigins` property (see deprecation notice below) can continue using the already established default behaviour for `tracePropagationTargets`
 
 <Alert level="info" title="Deprecation of tracingOrigins">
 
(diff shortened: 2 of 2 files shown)


sl0thentr0py (approved):

src/docs/sdk/performance/dynamic-sampling-context.mdx
```diff
@@ -59,6 +59,8 @@ Baggage is a standard HTTP header with URI encoded key-value pairs.
 For the propagation of DSC, SDKs first read the DSC from the baggage header of incoming requests/messages.
 To propagate DSC to downstream SDKs/services, we create a baggage header (or modify an existing one) through HTTP request instrumentation.
 
+The `baggage` header should only be attached to an outgoing request, if the request's URL matches at least one entry of the [`tracePropagationTargets`](/sdk/performance/#tracepropagationtargets) SDK option (or its fallback default behaviour).
```
lobsterkatie, line 62:
```suggestion
The `baggage` header should only be attached to an outgoing request if the request's URL matches at least one entry of the [`tracePropagationTargets`](/sdk/performance/#tracepropagationtargets) SDK option.
```

All of our options have default values, and the default behavior is/should be just following the default value, so I don't think saying so is necessary here.

src/docs/sdk/performance/index.mdx
```diff
@@ -32,6 +34,41 @@ Optionally, the `tracesSampler` callback can also return a boolean to force a sa
...
 See more about how sampling should be performed below.
 
+### `tracePropagationTargets`
+
+Sentry SDKs propagate trace information via headers of outgoing Http requests to downstream SDKs (see [`sentry-trace`](#header-sentry-trace) and [`baggage`](/sdk/performance/dynamic-sampling-context/#baggage) for more details).
+The `tracePropagationTargets` option gives users a mechanism of controlling to which outgoing Http requests these headers should be attached.
+For example, users can specify this property to keep trace propagation within their infrastructure, thereby stopping it from being sent to third party services.
+
+This option takes an array of strings or regular expressions to match against the URLs of outgoing requests.
+If this option is specified by users, SDKs may only add the trace headers to a request if the the request's URL matches (contains) at least one of the items from the array.
```
lobsterkatie, line 44:
```suggestion
Sentry SDKs propagate trace information to downstream SDKs via headers on outgoing HTTP requests. The `tracePropagationTargets` option gives users a mechanism of controlling to which outgoing HTTP requests these headers should be attached. For example, users can specify this property to keep trace propagation within their infrastructure, thereby preventing data within the headers from being sent to third party services.

See [`sentry-trace`](#header-sentry-trace) and [`baggage`](/sdk/performance/dynamic-sampling-context/#baggage) for more details on the individual headers.

This option takes an array of strings and/or regular expressions, and SDKs should only add trace headers to an outgoing request if the the request's URL matches (or, in the case of string literals, contains) at least one of the items from the array.
```

(This is pretty much all of your sentences, just rearranged a bit.)

src/docs/sdk/performance/index.mdx
```diff
@@ -32,6 +34,41 @@ Optionally, the `tracesSampler` callback can also return a boolean to force a sa
...
+URLs not matching: 'someHost.com/data', 'myApi.com/v1/projects'
+```
+
+#### Default Behaviour
+
+If the `tracePropagationTargets` option is not specified by users, SDKs should fall back to their default behaviour for trace propagation.
+This default behaviour is not standardized and can vary across SDKs.
+For instance, Server-side SDKs may propagate trace headers to all outgoing requests by default while client-side SDKs (e.g. Javascript) can fall back to a default array of allowed domains (e.g. `localhost` and `/^\//` in JavaScript).
+Choosing a sensible default behaviour is up to the SDK maintainers, however, special limitations, such as CORS issues for frontend web SDKs must be considered.
+SDKs that previously supported the `tracingOrigins` property (see deprecation notice below) can continue using the already established default behaviour for `tracePropagationTargets`
```
lobsterkatie, line 62: Given my thoughts about defaults, I think we'd be fine doing away with this section entirely and just adding a sentence or two above the example, something like

> SDKs may choose a default value which makes sense for their use case. Most SDKs default to the regex `/.*/` (attach headers to all outgoing requests), but deviation is allowed if necessary. (For example, because of CORS, browser-based SDKs default to only adding headers to internal requests.)

WDYT?

(If you do go this route, feel free to use or not use as much of the above as you'd like.)

src/docs/sdk/performance/index.mdx
```diff
@@ -32,6 +34,41 @@ Optionally, the `tracesSampler` callback can also return a boolean to force a sa
...
+
+<Alert level="info" title="Deprecation of tracingOrigins">
+
+This Option replaces the non-standardized `tracingOrigins` option which was previously used in some SDKs.
+SDKs that support `tracingOrigins` are encouraged to deprecate and and eventually remove `tracingOrigins` in favour `tracePropagationTargets`.
+In case both options are specified by users, SDKs should only rely on the `tracePropagationTargets` array.
```
lobsterkatie, line 68:
```suggestion
This option replaces the non-standardized `tracingOrigins` option which was previously used in some SDKs. SDKs that support `tracingOrigins` are encouraged to deprecate and and eventually remove `tracingOrigins` in favour `tracePropagationTargets`. In case both options are specified by users, SDKs should only rely on the `tracePropagationTargets` array.
```

(No change other than pulling the extra newlines.)

src/docs/sdk/performance/index.mdx
```diff
@@ -160,6 +197,8 @@ To offer a minimal compatibility with the [W3C `traceparent` header](https://www
 To avoid confusion with the W3C `traceparent` header (to which our header is similar but not identical), we call it simply `sentry-trace`.
 No version is being defined in the header.
 
+The `sentry-trace` header should only be attached to an outgoing request, if the request's URL matches at least one entry of the [`tracePropagationTargets`](#tracepropagationtargets) SDK option (or its fallback default behaviour).
```
lobsterkatie, line 200:
```suggestion
The `sentry-trace` header should only be attached to an outgoing request if the request's URL matches at least one entry of the [`tracePropagationTargets`](#tracepropagationtargets) SDK option.
```

lobsterkatie (commented):
This looks good so far!

My one substantive piece of feedback is that I think trying to differentiate between a default _value_ for `tracePropogationTargets` and it's default _behavior_ is over-complicating things. We have our weird span-creation thing (which we're going to fix), but every other SDK which implemented `tracingOrigins` did it the way that `tracePropagationTargets` will, so "default behavior" is really just abiding by a default value for the option. Thinking about it that way lets us tighten things up a bit by cutting out some of the `this (or this)` stuff, as you'll see I've suggested below.

Other than that, all of my suggestions below are just rearrangements of what you already had.

src/docs/sdk/performance/index.mdx
Lms24: What I was going for with the newlines was easier readability and for instance less clutter when reviewers would e.g. make wording suggestions in one sentence. 
I think from a MD rendering perspective, having one or no newline after each sentence shouldn't make any difference (as long as it's just one newline obv.). 
Came across this in various PRs of our docs repos and I thought it makes sense. But admittedly, I didn't look if we have a convention for this or not.

src/docs/sdk/performance/dynamic-sampling-context.mdx
Lms24: Good point

src/docs/sdk/performance/index.mdx
Lms24: Ok, did some googling and it turns out that this apparently is a debated topic (example [one](https://github.com/github/choosealicense.com/issues/505), [two](https://twitter.com/jakevdp/status/796817091684769792), + a ton more). Another advantage is less conflicts/better diffing in git.

src/docs/sdk/performance/index.mdx
Lms24: [0c98f1c](https://github.com/getsentry/develop/pull/648/commits/0c98f1c4446f7f3aa8292a36f2754e9d2240a262)

src/docs/sdk/performance/index.mdx
Lms24: I think that's a good change. Took your example and adjusted it slightly

Lms24:
Thanks for the great review!

>My one substantive piece of feedback is that I think trying to differentiate between a default value for tracePropogationTargets and it's default behavior is over-complicating things.

Yes that's a good point. Let's just go with default _values_ (which, thinking about it again, makes much more sense). 

Adjusted in [37d7a33](https://github.com/getsentry/develop/pull/648/commits/37d7a3364105ef19b7b3031949fce62e10f1422a)

src/docs/sdk/performance/index.mdx
lobsterkatie: It's interesting - this is the first time I've ever seen it done. Normally people do hard wrapping, which is _awful_ for diffing, and (though I can't find it anywhere now among our 17 docs style guides) I could swear that at one point we actually had "one line per paragraph" as part of our docs guidelines.

I do see the point about diffing. At the same time, I find it significantly harder to read, both from a shape perspective (my eyes have to move a different amount for every line) and from a semantic perspective (the newline makes my brain pause for a second, so the text doesn't flow). It also means one has to think about which newlines actually are paragraph breaks and which aren't, especially given the line padding the GH does (which makes it look a little like each line is a new paragraph).

It's true that GH didn't do a great job of this in the suggestion above, but I generally find that the brighter highlighting that it/VSCode/Fork/etc normally do is enough:

image

So IDK. Unless/until we reach a point where we decide that's what we do, everywhere (and we all adjust to reading that way), my preference would be to leave them out. But your PR, your call. 🙂

lobsterkatie (approved):

src/docs/sdk/performance/index.mdx
Lms24: Ok, let's go with consistency (changed it in [bd4a295](https://github.com/getsentry/develop/pull/648/commits/bd4a295e5f155d08c19f09aecd9ff18d2f82a6e3)). 

I think it might be worth having this discussion in the future (maybe if we ever get a #docsengineer 👀 ). I personally find the one-sentence-lines easier to read because I'm not a fan of eternally long lines (also with auto wrap and stuff) but I agree that consistency is more important. (Maybe it's just PTSD from my LaTeX days at university where this was even more of a problem)

adinauer (approved):

PR #651, opened 2022-07-20
proto field is required for DS

bruno-garcia:


marandaneto (approved):

untitaker (approved):

untitaker:
To clarify: dynamic sampling still works even without that field. you just won't get unsampled numbers such as transactions count.

PR #655, opened 2022-07-21
Note on using Ngrok for local dev and installing Slack App

armenzg:
If this step is not taken, the installation of the Slack App will complain that the redirect URI points to http://localhost:8000.


leeandher (approved):
Thanks for adding this!

armenzg:
Thanks for getting me the answer I needed!

PR #658, opened 2022-08-02
Add `app_memory` to `app` context

timfish:
Currently the `device` context has properties for the device memory usage but it can be useful to know how much memory an app was using when an event or transaction was captured. 

Open to better naming than `memory_used` 🤔

Related issues:
https://github.com/getsentry/sentry-cocoa/issues/999

### DIFF
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -354,7 +354,7 @@ The `type` and default key is `"app"`.
 
 : _Optional_. Internal build identifier, as it appears on the platform.
 
-`memory_used`
+`app_memory`
 
 : _Optional_. Amount of memory used by the application in bytes.
 
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -354,6 +354,10 @@ The `type` and default key is `"app"`.
 
 : _Optional_. Internal build identifier, as it appears on the platform.
 
+`memory_used`
+
+: _Optional_. Amount of memory used by the application in bytes.
+
 ## Browser Context
 
 Browser context carries information about the browser or user agent for


bruno-garcia (approved):

bruno-garcia:
@jan-auer @mattjohnsonpint @marandaneto  thoughts?

timfish:
Might also be worth including `storage_used` to help with this:
https://github.com/getsentry/team-mobile/issues/8

bruno-garcia:
Dotnet sdk has "Memory Info", i believe the 
 units are not well defined. So the ui doesn't present KB/MB etc, just shows as-is

AbhiPrasad:
I'm +1 on adding this, and I think `memory_used` effectively articulates what it's doing.

If we are worried about units not being well defined, we could also just include the unit as part of the value.
For example, `"memory_used": "30 kB"`. Thoughts on that? We are probably not going to index this - so I don't think we need to worry about high cardinality concerns.

timfish:
Everything in `contexts` is in bytes apart from `gpu.memory_size` [which is in Megabytes](https://develop.sentry.dev/sdk/event-payloads/contexts/#gpu-context).

marandaneto:
Android SDK already reports `contexts.device.memory_size`, `contexts.device.free_memory` and `contexts.device. usable_memory`, all in bytes.

timfish:
There are a large number of properties documented in `contexts.device` that are not specifically handled in Relay or Sentry right now such as `boot_time`, `processor_count`, `processor_frequency` and all the `supports_*`.

AbhiPrasad:
Yeah only a certain few fields are indexed. If you want to see the full list -> https://github.com/getsentry/sentry/blob/master/src/sentry/snuba/events.py

https://github.com/getsentry/sentry/blob/0ce68802e966fe49a595ee1697804929c2a55f50/src/sentry/snuba/events.py#L240-L316

AbhiPrasad:
We can open up a GH issue on the Sentry repo to add `memory_used` to the UI after we merge the spec in!

brustolin:
I want to suggest to keep pattern and called it "app_memory"

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -354,6 +354,10 @@ The `type` and default key is `"app"`.
 
 : _Optional_. Internal build identifier, as it appears on the platform.
 
+`memory_used`
```
brustolin, line 357:
```suggestion
`app_memory`
```

brustolin:
@AbhiPrasad, @bruno-garcia, What else do we need to merge this?

AbhiPrasad:
I'm going to go ahead and merge this! I'll take care of updating relay.

PR #667, opened 2022-08-10
Discourage use of transaction source unknown

jjbayer:
`unknown` should only be set by Relay, as a value for SDKs that do not send the `source` attribute.

image


sl0thentr0py (approved):
thx @jjbayer!

Issue #657, opened 2022-07-29
Add documentation around the craft publish process


AbhiPrasad:
1. Setting up https://github.com/getsentry/craft (mostly just a summary of the README)
  a. What changelog policy to use
  b. What targets to set
  c. Setting up a version bumping script, and some examples
2. Using the publish workflow https://github.com/getsentry/publish
  a. When to request a release and how to fill out the action form
  b. How does release process differ for alphas/betas
  c. Who can approve the release https://github.com/orgs/getsentry/teams/release-approvers
  d. Troubleshooting a failed release

PR #674, opened 2022-08-18
add craft quick-start

asottile-sentry:


AbhiPrasad:
Ref: https://github.com/getsentry/develop/issues/657

Maybe we keep this under SDK docs for now, but it might even make sense to make a brand new section around it.

chadwhitacre:
How does this relate to the Craft README?

Also what's with the lack of capitalization? ☹️

chadwhitacre:
> How does this relate to the Craft README?

Hrm, actually seems like it might overlap more with [the `publish` README](https://github.com/getsentry/publish). Is this intended to replace that?

asottile-sentry:
this is the ticket I'm trying to solve: https://getsentry.atlassian.net/browse/DEVINFRA-21

it doesn't make sense in `craft` as that's a generic release tool

the `publish` readme is similar, but it's also hidden away behind a big (imo kinda useless?) state diagram

chadwhitacre:
A new section on dev docs makes sense to me, I think whatever it duplicates from the `publish` README should be removed from the `publish` README with a link over to dev docs.

Creating the big state diagram is how I wrapped my head around how the `publish` repo works, if nobody else finds it valuable we can get rid of it.

JonasBa:
Just want to chip in here, I recently setup craft and have hit some errors along the way. Having a quick-start guide like this that explains the flow would have been great. In the end, I just asked my q in slack

src/docs/sdk/craft-quick-start.mdx
```diff
@@ -0,0 +1,146 @@
...
+```bash
+git tag 0.0.0 "$(git log -1 --reverse --format=%h)"
+git push origin --tags
+```
+
+[a limitation]: https://github.com/getsentry/craft/issues/342
+
+## ci requirements
+
+first we need to ensure your repository responds to `release/**` branches
```
mattgauntseo-sentry, line 29:
```suggestion
first we need to ensure your repository has a GitHub action that responds to `release/**` branches
```

src/docs/sdk/craft-quick-start.mdx
```diff
@@ -0,0 +1,146 @@
...
+and produces an artifact with the right name
+
+for our example repository that's done using [push: branches] and an
```
mattgauntseo-sentry, line 32:
```suggestion
for our example repository that's done by ensuring the following in a GitHub action:
1. [push: branches]
    ```yaml
    on:
      pull_request:
      push:
        branches: [main, test-me-*, release/**]
    ```
```

src/docs/sdk/craft-quick-start.mdx
```diff
@@ -0,0 +1,146 @@
...
+[artifact instruction]
```
mattgauntseo-sentry, line 33:
```suggestion
1. [artifact instruction]
    ```yaml
        - uses: actions/upload-artifact@v3
          with:
            name: ${{ github.sha }}
            path: dist/*
    ```
```

src/docs/sdk/craft-quick-start.mdx
```diff
@@ -0,0 +1,146 @@
...
+        with:
+          version: ${{ github.event.inputs.version }}
+          force: ${{ github.event.inputs.force }}
+```
+
+### commit + PR
+
+**commit those three files and make a pull request**
+
+here's [an example PR] and the [followup to fix fetch-depth]
```
mattgauntseo-sentry, line 123: Why include the fetch-depth PR?

src/docs/sdk/craft-quick-start.mdx
```diff
@@ -0,0 +1,146 @@
...
+
+you can do this self-service via the settings page of your repository:
+`https://github.com/getsentry/REPONAME_HERE/settings/access`
+
+## making your first release!
+
+navigate to the actions tab of your repository, locate the release workflow,
+and create the first release!  I used `1.0.0` as the first version.
+
+this will create an [issue in publish] which you'll need an approver to
```
mattgauntseo-sentry, line 143: This suggests that the "deployer" should reach out to someone, is that the case or should they wait for an approver to eventually get to it?

mattgauntseo-sentry (approved):
Feel free to ignore the suggestions, I favor inlining code where possible to avoid making folks jump between pages, especially in this case where we really want folks to get up and running quickly.

src/docs/sdk/craft-quick-start.mdx
asottile-sentry: idk, I usually wait

src/docs/sdk/craft-quick-start.mdx
asottile-sentry: because it's needed -- and I forgot it in the first PR

src/docs/sdk/craft-quick-start.mdx
asottile-sentry: there's other ways to do this so I didn't want to prescribe the exact solution

PR #675, opened 2022-08-19
Added logs.mdx file about log rotation by a docker daemon

vzhgit:
Hi.
Sorry for delay, too much work. @aminvakil I mention you because you asked me 
https://github.com/getsentry/self-hosted/issues/1594#issuecomment-1202229863


src/docs/self-hosted/index.mdx
```diff
@@ -82,6 +82,7 @@ Here is further information on specific configuration topics related to self-hos
 - [Email](/self-hosted/email/)
 - [Geolocation](/self-hosted/geolocation/)
 - [Single Sign-On (SSO)](/self-hosted/sso/)
+- [Logs rotaion](/self-hosted/logs/)
```
aminvakil, line 85:
```suggestion
- [Log rotation](/self-hosted/log-rotation/)
```

src/docs/self-hosted/logs.mdx
```diff
@@ -0,0 +1,22 @@
+---
+title: "Logs rotation"
```
aminvakil, line 2:
```suggestion
title: "Log rotation"
```

src/docs/self-hosted/logs.mdx
```diff
@@ -0,0 +1,22 @@
...
+---
+
+By default docker writes logs of a container to one log file which can be giant - tens of gigabytes.
```
aminvakil, line 5:
```suggestion
By default docker writes logs of each container into a single file which can be giant - tens of gigabytes.
```

src/docs/self-hosted/logs.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+To reduce a used space for logs you can
```
aminvakil, line 7:
```suggestion
To reduce used disk space of these logs you can 
```

src/docs/self-hosted/logs.mdx
```diff
@@ -0,0 +1,22 @@
+---
+title: "Logs rotation"
+---
+
+By default docker writes logs of a container to one log file which can be giant - tens of gigabytes.
+
+To reduce a used space for logs you can 
+ * modify docker-compose.yaml file by adding into a common section for each container following logging settings:
```
aminvakil, line 8: We use `.yml` extension in self-hosted, right now, I'm changing it just for convenience.
```suggestion
 * modify docker-compose.yml file and add following logging settings into a common section for each container:
```

src/docs/self-hosted/logs.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+```
+x-restart-policy: &restart_policy
```
aminvakil, line 11: This documentation is about log policy, not changing restart policies.
```suggestion
x-log-policy: &log_policy
```

src/docs/self-hosted/logs.mdx
```diff
@@ -0,0 +1,22 @@
...
+  restart: unless-stopped
```
aminvakil, line 12: I think it would be better not to add this here.
```suggestion
```

src/docs/self-hosted/logs.mdx
```diff
@@ -0,0 +1,22 @@
...
+  logging:
+    driver: "json-file"
+    options:
+      max-size: "20m"
+      max-file: "10"
+```
+
+And docker daemon will automatically control size of a log file for each container from the yaml file: 20Mb * 10 = 200Mb for each container. <a href="https://docs.docker.com/compose/compose-file/compose-file-v3/#logging">Offical documentaion</a> about those settings.
```
aminvakil, line 20:
```suggestion
And docker daemon will automatically manage disk space used of each container's log file from the yaml file: 20Mb * 10 = 200Mb for each container. You can take a look at <a href="https://docs.docker.com/compose/compose-file/compose-file-v3/#logging">Offical documentaion</a> about those settings.
```

src/docs/self-hosted/logs.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+ * modify a docker daemon <a href="https://docs.docker.com/config/containers/logging/configure/"><code>log settings</code></a> to apply them to all newly created containers.
```
aminvakil, line 22:
```suggestion
 * modify docker daemon <a href="https://docs.docker.com/config/containers/logging/configure/"><code>log settings</code></a> to apply them to all newly created containers system-wide.
```

aminvakil (changes requested):
Thanks for adding this! I've added a couple of notes and suggested a couple of changes.

src/docs/self-hosted/logs.mdx
vzhgit: Please see next comment

src/docs/self-hosted/logs.mdx
vzhgit: My original proposal to solve that problem was with minimal actions because “&restart_policy” is already used by each container.
If we're creating a new common section, we should reformulate following phrase:

    modify docker-compose.yml file and add following common logging setting and apply it either for each container or only for containers with huge log files:

vzhgit (commented):
Sorry for mess. Maybe I did something in a wrong way. It's my first PR with suggestions.

aminvakil:
> Sorry for mess. Maybe I did something in a wrong way. It's my first PR with suggestions.

It's totally fine! You've just marked all my comments as resolved and nothing has been changed, mark them as unresolved and "apply my suggestion" if you agree, comment on my comment if you disagree.

More info: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request#applying-suggested-changes

src/docs/self-hosted/logs.mdx
aminvakil: It's not clear enough to me this way, also it will affect all containers as maybe users only want to apply this policy to some containers.

PR #686, opened 2022-09-13
feat: Initial OpenTelemetry Spec

AbhiPrasad:
https://develop-git-abhi-otel.sentry.dev/sdk/performance/opentelemetry/

This PR outlines an initial spec for OpenTelemetry usage in Sentry. It focuses on outlining the Span protocol (how to map from otel span -> sentry span), as well as mapping out some high level sections for us to work on here.

### Open Questions

There are still some open questions we need to answer together at this stage.

1. At a high level we established that we would like the OpenTelemetry SDK to be responsible to create transactions. Therefore we must decide how this is going to be done

If there are two traces ongoing, do we create two transactions? If we do, which one becomes the active transaction? Do we store state about all ongoing transactions? 

My opinion: We track all transactions in a set, and remove them when we get finished. So instead of assigning to the active transaction on the scope, we assign to transactions the span processor is aware about. Pseudo-algo:

```py
# Ideally it's a Weak Map that holds WeakRef
map = new Map();

def on_span_start(otel_span):
  transaction = map.get(otel_span.trace_id)
  if transaction:
    return transaction

  new_transaction = transaction_from_span(otel_span)
  map.add(new_transaction.trace_id, new_transaction)
  if (map.length === 1):
    make_active_transaction(new_transaction)
    
def on_span_finish(otel_span):
  transaction = map.get(otel_span.trace_id)
  if transaction
    if transaction.span_id == otel_span.span_id:
      transaction.finish(otel_span.end_timestamp)
      map.delete(transaction.trace_id)
      if get_active_transaction(transaction) == transaction
        make_active_transaction(nil)
    else:
      add_span_to_transaction(otel_span)
  # if no transaction exists, that means the transaction was already sent to Sentry and finished. 
  # Create a new transaction from this span and send it to Sentry
  new_transaction = transaction_from_span(otel_span)
  new_transaction.finish(otel_span.end_timestamp)
```

2. Right now I've briefly just said we rely on a combination of opentelemetry span kind, span status, span attributes and span name to decide sentry span operation and span description. Getting a correct span operation is imperative to making sure we can make use of Sentry product features (span operations, performance issues, etc) - so we'll need to make sure these are consistent.

OpenTelemetry has a [list of semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/README.md) that identify the type of span based on span attributes, but this is annoying to keep duplicating in every single SDK. The easiest solution here is just to push all of that work onto Relay.

So,

1. SDK maps all attributes -> tags
2. In Relay we detect a transaction as coming from opentelemetry:
3. In Relay, we process the tags to generate ops, descriptions and status for the spans (and also maybe transaction)

With dynamic sampling this is also not a problem, since we can just use the OpenTelemetry Span name as the transaction name, so everything should be consistent.

We did some of this mapping work in the [Sentry Exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/e043da8ff3d4a28ed33a11584e625a5404a6049a/exporter/sentryexporter/sentry_exporter.go#L269-L275). For the purposes of the MVP in Ruby/Python, we can prob do this mapping in the SDK though.

### Final Thoughts

My goal here is to get this merged in after the protocol gets approved. We can then split up the work around each of the different sections and iterate on this together.


src/docs/sdk/performance/opentelemetry.mdx
```diff
@@ -0,0 +1,46 @@
+---
+title: "OpenTelemetry Support"
+---
+
+<Alert level="warning">
+
+This page is under active development. Specifications are not final and subject to change.
+
+"Proof is in the progress" - Drake
```
vladanpaunovic, line 9: 🎤

sl0thentr0py: started from the bottom

src/docs/sdk/performance/opentelemetry/span-protocol.mdx
```diff
@@ -0,0 +1,342 @@
...
+return sentryStatus
+```
+
+### Span Events
+
+OpenTelemetry, has the concept of [Span Events](https://github.com/open-telemetry/opentelemetry-proto/blob/724e427879e3d2bae2edc0218fff06e37b9eb46e/opentelemetry/proto/trace/v1/trace.proto#L193-L214). As per the spec:
+
+> An event is a human-readable message on a span that represents “something happening” during it’s lifetime
+
+In Sentry, we have two options for how to treat span events. First, we can add them as breadcrumbs to the transaction the span belongs to. Second, we can create an artificial "point-in-time" span (a span with 0 duration), and add it to the span tree. TODO on what approach we take here.
```
vladanpaunovic, line 338: Adding a _point-in-time_ span this way seems a bit hacky to me.

When you say "_we can add them as breadcrumbs to the transaction_", you mean we can add them as child transactions, not via `addBreadcrumb()`

Maybe clarifying this a bit so people (like me) don't mistake spans for breadcrumbs :)

src/docs/sdk/performance/opentelemetry/index.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+TODO: Talk about the approach we are using, based on Matt's hackweek project - https://github.com/getsentry/sentry-ruby/pull/1876
+
+## Span Protocol
+
+import "./span-protocol.mdx"
+
+## Transaction Protocol
+
+There is no concept of a transaction within OpenTelemetry, so we rely on promoting spans to become transactions. The span `description` becomes the transaction `name`, and the span `op` becomes the transaction `op`. Therefore, OpenTelemetry spans must be mapped to Sentry spans before they can be promoted to become a transaction.
```
vladanpaunovic, line 31: When talking about this, I would make it clearer on what spans are Otel spans and what spans are Sentry spans.

I can kinda figure it out, but it feels like I am guessing.

src/docs/sdk/performance/opentelemetry/index.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+## OpenTelemetry Context
+
+Aside from information from Spans and Transactions, OpenTelemetry has meta-level information about the SDK, resource, and service that generated spans. To track this information, we generate a new OpenTelemetry Event Context.
+
+The existence of this context on an event (transaction or error) is how the Sentry backend will know that the incoming event is an OpenTelemetry event.
```
vladanpaunovic, line 37: You can link to event payload here and move the Otel context to event payload page.

src/docs/sdk/performance/opentelemetry/index.mdx
```diff
@@ -0,0 +1,80 @@
...
+        auto_version?: string;
+    };
+
+    // Rest of attribute keys excluding service and sdk keys
+    // https://github.com/open-telemetry/opentelemetry-proto/blob/724e427879e3d2bae2edc0218fff06e37b9eb46e/opentelemetry/proto/resource/v1/resource.proto#L32
+    [key: string]?: string,
+}
+```
+
+The reason sdk and service are split are so they can be indexed as top level fields in the future for easier usage within Sentry.
```
vladanpaunovic, line 75: It would be good to capture how are we planning to fill in the service fields

danielkhan:
@AbhiPrasad 

> If there are two traces ongoing, do we create two transactions? If we do, which one becomes the active transaction? Do we store state about all ongoing transactions?

Can you outline a scenario that would cause two ongoing traces within one execution?
Technically it's possible for Node but this should not be in the same context.
Within the same context, I would assume that there is only one active trace with unique trace ID.

> My opinion: We track all transactions in a set, and remove them when we get finished. So instead of assigning to the active transaction on the scope, we assign to transactions the span processor is aware about.

Such operations often cause memory leaks, don't they?

> OpenTelemetry has a [list of semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/README.md) that identify the type of span based on span attributes, but this is annoying to keep duplicating in every single SDK. The easiest solution here is just to push all of that work onto Relay.

Yes to avoid redundancies by centralizing on Relay if needed.

src/docs/sdk/performance/opentelemetry/index.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+</Alert>
+
+This document details Sentry's work in integrating and supporting [OpenTelemetry](https://opentelemetry.io/), the open standard for metrics, traces and logs. In particular, it focuses on the integration between [Sentry's performance monitoring product](https://docs.sentry.io/product/performance/) and [OpenTelemetry's tracing spec](https://opentelemetry.io/docs/concepts/signals/traces/).
+
+## Background
+
+When Sentry performance monitoring was initially introduced, OpenTelemetry was in early stages. This lead to us adopt a slightly different model from OpenTelemetry, notably we have this concept of transactions that OpenTelemetry does not have. We've described this, and some more historical background, in our <Link to="/sdk/research/performance/">performance monitoring research document</Link>.
+
+TODO: Add history about OpenTelemetry Sentry Exporter: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/sentryexporter
```
danielkhan, line 19: Really needed in the docs? I would leave that to a README.

src/docs/sdk/performance/opentelemetry/index.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+"Proof is in the progress" - Drake
+
+</Alert>
+
+This document details Sentry's work in integrating and supporting [OpenTelemetry](https://opentelemetry.io/), the open standard for metrics, traces and logs. In particular, it focuses on the integration between [Sentry's performance monitoring product](https://docs.sentry.io/product/performance/) and [OpenTelemetry's tracing spec](https://opentelemetry.io/docs/concepts/signals/traces/).
+
+## Background
+
+When Sentry performance monitoring was initially introduced, OpenTelemetry was in early stages. This lead to us adopt a slightly different model from OpenTelemetry, notably we have this concept of transactions that OpenTelemetry does not have. We've described this, and some more historical background, in our <Link to="/sdk/research/performance/">performance monitoring research document</Link>.
```
danielkhan, line 17: I would just mention that Sentry and OpenTelemetry have conceptual differences. This is normal and true for each vendor with an SDK that predates OTel. It doesn't matter why it is like that today. It's more important to a reader what the differences are.

src/docs/sdk/performance/opentelemetry/index.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+TODO: Add history about OpenTelemetry Sentry Exporter: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/sentryexporter
+
+## Approach
+
+TODO: Talk about the approach we are using, based on Matt's hackweek project - https://github.com/getsentry/sentry-ruby/pull/1876
```
danielkhan, line 23: Does this really need to go into the docs?
Should this document our way to a solution or the solution?

src/docs/sdk/performance/opentelemetry/index.mdx
danielkhan: +1 - maybe an illustration. I can help if needed.

src/docs/sdk/performance/opentelemetry/index.mdx
danielkhan: +1

src/docs/sdk/performance/opentelemetry/span-protocol.mdx
antonpirker: I suggest to add them as both, breadcrumbs and point-in-time-spans.
The Ruby docs of oTel span events [1] state that `characteristic of events is that their timestamps are displayed as offsets from the beginning of the span`. This suggested we display them in our performance waterfall diagram. 
And adding them to the breadcrumbs makes sense, because this is a fundamental Sentry feature and existing users are used to it. 

1: https://opentelemetry.io/docs/instrumentation/ruby/events/

src/docs/sdk/performance/opentelemetry/index.mdx
```diff
@@ -0,0 +1,80 @@
...
+        // from: service.namespace
+        namespace?: string;
+        // from: service.instance.id
+        instance_id?: string;
+        // from: service.version
+        version?: string;
+    };
+
+    // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#telemetry-sdk
+    sdk?: {
```
antonpirker, line 58: Maybe call this "otel_sdk" (or something similar) so we do not confuse it with sentry sdk versions, making just everything just more explicit?

src/docs/sdk/performance/opentelemetry/span-protocol.mdx
AbhiPrasad: Let's table this discussion - we can leave this as a TODO item here and evaluate this once we only really need to (like when the first Ruby SDK part is ready to review).

src/docs/sdk/performance/opentelemetry/index.mdx
AbhiPrasad: I'll work on clarifying the wording, let me see what I can do about the illustration.

src/docs/sdk/performance/opentelemetry/index.mdx
AbhiPrasad: I'm actually going to leave this in here so we don't confuse others reading the develop docs (new hires, other teams, etc.). once we are comfortable with this new otel context schema, let's move this outside into the main event payload page.

src/docs/sdk/performance/opentelemetry/index.mdx
AbhiPrasad: This is done using by grabbing them off the OpenTelemetry resource, but I can do this!

src/docs/sdk/performance/opentelemetry/span-protocol.mdx
```diff
@@ -0,0 +1,342 @@
...
+
+In the special case that the span event is an exception span, [where the `name` of the span event is `exception`](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/exceptions/), we also have the possibility of generating a Sentry error from an exception. In this case, we can create this [exception based on the attributes of an event](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/exceptions/#attributes), which include the error message and stacktrace. This exception can also inherit all other attributes of the span event + span as tags on the event.
```
untitaker, line 340: which and how many attributes do we expect there? if there are a ton of default attributes attached to an exception span, can we define which of them actually end up in event tags?

there is a storage concern about adding too many default tags to the event payload, and a product one as well considering user-defined tags and tags inherited from otel count against the same size limit

untitaker (commented):
I think this spec has some potentially big implications for storage so I would like somebody from S&S to review this as well.

fpacifici:
> My opinion: We track all transactions in a set, and remove them when we get finished. So instead of assigning to the active transaction on the scope, we assign to transactions the span processor is aware about. Pseudo-algo:

If I get this correctly, in this system, a sequence of spans at the root level of a trace would create one new transaction per span.
In the current Sentry protocol you could fit all of them in the same transaction.
Switching from the old model to the new model can multiply the number of produced transactions. Do we have a sense of the potential impact this would have on customer's quotas and ingestion capacity ?

fpacifici:
Since this could have very large cross system implications have you considered discussing proposing an rfc ?
https://github.com/getsentry/rfcs/

AbhiPrasad:
@fpacifici @untitaker Thanks for taking a look! I think we need to do some more testing before we can better understand storage and ingestion ramifications. For example, theoretically this change should send transactions at the same rate fir both an OpenTelemetry and Sentry SDK (otel sdk behaves === sentry sdk), but until we test various scenarios we cannot confirm this.

The primary purpose of this document was to outline a high level approach so we could start experimenting with the SDKs. Once we get to a comfortable spot and have an SDK prototype, we will open a more formal RFC so we can discuss infrastructure ramifications. There we can also bring up changes we'll need to make to Relay/ingestion to support this.

src/docs/sdk/performance/opentelemetry/index.mdx
AbhiPrasad: It should document the solution, I'll work with the team to get this in there.

untitaker:
>There we can also bring up changes we'll need to make to Relay/ingestion to support this.

What if SDK changes are needed? Are those then still possible? If the answer is no, this is IMO not a sustainable approach to prototyping.

For example I see a 50:50 chance that we cannot send and store this data as tags because there are too many. Would your proposal be to later detect those payloads in Relay and selectively move tags elsewhere? I'm sure that's convenient for SDK authors but painful to maintain for ingest.

danielkhan:
@untitaker 
>  I'm sure that's convenient for SDK authors but painful to maintain for ingest.

Yet, if we need to do a translation of values and tags, it would be redundant to do that in the individual SDKs. Instead, I would rather - if possible - vow for a dedicated mapping component that does this. This component could maybe also be maintained by the SDK team.

untitaker:
>Yet, if we need to do a translation of values and tags, it would be redundant to do that in the individual SDKs.

I don't have an issue with putting more things in Relay. I mainly take issue with the idea that questions concerning infrastructure and software architecture are entirely left unanswered until we have shipped a prototype potentially already used by customers. I don't have a clear suggestion for what we should do (because we don't have an answer to those questions), and I suspect the concern about tags is not the biggest concern with this proposal anyway. **If our answer is that SDK changes are still possible after prototyping (incl breakign changes) that's sufficient to me.**

----

**Regarding tags specifically**

So I think this is going too much into detail, but some concerns with tags off the top of my head:

1. generically named otel attributes can conflict with user-defined tags
2. tags have a single static limit applied to them. if there are more than `n` tags, we have no way of knowing which tags to trim first unless we build a static map of tag names we consider otel-specific
3. tags can only be strings (so greater-than or smaller-than queries are not possible)

then there are combinations of those problems. for example, let's say the UI is built to interpret `foo.bar` tag a certain way (because `foo.bar` has meaning in otel), eg assume an integer value. What happens if the user sends a custom tag of the same name?

you can fix all of those problems by making relay convert tags into some other structure in the payload (which is something we currently don't do). There are two downsides to that:

* the event json visible in the UI further diverges from what the user sends
* tag limits are still applied beforehand depending on customer relay version

one other option would be to dump otel attributes into a new custom context, and then figure out which of those attributes can be indexed and which limits to apply. More time would be needed (and an answer to the above questions) to determine whether that's a good idea.

untitaker:
>Instead, I would rather - if possible - vow for a dedicated mapping component that does this.

I think what you're suggesting here is that we should maintain two separate schemas: ingestion schema maintained by SDKs, and storage schema maintained by infra teams. I think that idea has potential but it's not where we're at right now, and it would break some assumptions both customers and our internals have.

danielkhan:
>  I mainly take issue with the idea that questions concerning infrastructure and software architecture are entirely left unanswered until we have shipped a prototype potentially already used by customers.

We won't release a prototype to customers, @untitaker.
This right here is the right place to raise all the concerns and discuss solutions and once this is all figured out, we will go ahead and productize this. Nothing will be released that isn't production ready and signed off by all teams this topic touches.

>  I don't think it's feasible to have them be maintained by two separate teams
That's fine for me. It was just an itch but if we aren't there yet, it's just that.

AbhiPrasad:
> one other option would be to dump otel attributes into a new custom context, and then let relay figure out which of those attributes can be indexed and which limits to apply. More time would be needed (and an answer to the above questions) to determine whether that's a good idea.

So for now, we are going to go with this approach - and then we can evaluate what to do as we see data coming in. So nothing stored in tags for now for transactions.

I'm going to revamps this based on some convos we had, and merge this in as a starting point. The next step we need to take here is to document the SDK API so that we have a baseline to work off of.

Once we're comfortable with that, we can start RFCs / wider convos re: logic in Relay and product implications.

untitaker:
Sounds good, thanks @AbhiPrasad!

PR #687, opened 2022-09-14
Add blurb about self-hosted monitoring

ethanhs:


src/docs/self-hosted/index.mdx
```diff
@@ -8,6 +8,10 @@ In addition to making its source code available publicly, Sentry offers and main
 
 Our recommendation is to download the [latest release of the self-hosted repository](https://github.com/getsentry/self-hosted/releases/latest), and then run `./install.sh` inside this directory. This script will take care of all the things you need to get started, including a base-line configuration, and then will tell you to run `docker-compose up -d` to start Sentry. Sentry binds to port `9000` by default. You should be able to reach the login page at [http://127.0.0.1:9000](http://127.0.0.1:9000/).
 
+### Self-Hosted Monitoring
+
+When you run `./install.sh`, you have a choice to opt in or out of our monitoring. This monitoring is used for development and debugging purposes. For more details please see the [section in the self-hosted README](https://github.com/getsentry/self-hosted#self-hosted-monitoring). Note that choosing whether to send errors or not will become mandatory beginning with the 22.10.0 release.
```
hubertdeng123, line 13: I feel like `Note that choosing whether to send errors or not will become mandatory beginning with the 22.10.0 release.` might be better on a new line, thoughts?

hubertdeng123, line 13: I feel like we could add some more info on how the monitoring will be used just to encourage people to enable it.

`This monitoring is used for development and debugging purposes so that we're on top of issues you're facing, allowing us to provide a more seamless installation process`

ethanhs: Yeah I originally had something but I didn't want to say its currently only for the installer (we may expand things in the future). I like your wording.

hubertdeng123 (approved):

PR #700, opened 2022-09-30
ref: Add Angular SDK span ops

Lms24:
The Angular SDK adds a few spans when using [component tracking](https://docs.sentry.io/platforms/javascript/guides/angular/components/tracehelpers/) features as well as when the [Angular routing instrumentation](https://docs.sentry.io/platforms/javascript/guides/angular/components/tracehelpers/) is used.

This PR adds these span operations to the JS Frameworks table, analogously to the span ops of our other Framework SDKs.

Open question: In [the docs (last code example on this page)](https://docs.sentry.io/platforms/javascript/guides/angular/components/tracehelpers/), we give an example on how to add custom spans to the Angular bootstrapping process. The span op we suggest for this span is `ui.angular.bootstrap`. Do we want to mention this in the dev spec? I didn't yet add it because technically speaking, our SDK doesn't do this automatically. Probably fine this way but I just wanted to mention it anyway.

### DIFF
--- a/src/docs/sdk/performance/span-operations.mdx
+++ b/src/docs/sdk/performance/span-operations.mdx
@@ -74,7 +74,7 @@ JS Frameworks should be prepended with the `ui` category for operations related 
 | ui.angular |                                 | Spans related to [Angular](https://angular.io/) components           |
 |            | ui.angular.init                 |                                                                      |
 |            | ui.angular.routing              |                                                                      |
-|            | ui.angular.<methodName>         | Created when a method wrapped with `TraceMethodDecorator` was called |
+|            | ui.angular.[methodName]         | Created when a method wrapped with `TraceMethodDecorator` was called |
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):
I don’t think we need to include the manual instrumentation op, but you can include it if you feel it will be useful.

The purpose of this page was more to be an example reference rather than an exhaustive list of all possible ops, the only thing that needs to be documented if you are adding a new category, as per the table at the bottom of the page (`category.X`)

Lms24:
Ahh I see. I thought we should add the span ops because we wanted to have an overview on which ops we currently have (I've seen that there were some efforts in this regard while I was on PTO). Happy to close this PR if it doesn't add value

AbhiPrasad:
No, I think we can merge it in now that it’s here. The more examples, the better it is for folks to understand.

PR #708, opened 2022-10-12
Polish Github installation instructions

armenzg:
Various fixes:
- It is unclear the create app button needs the URL to be adjusted [1]
- The rendered page hides where to place the webhook secret. [2]

[1]
```diff
- https://github.com/organizations/your-organization/settings/apps/new? [trimmed parameters]
+ https://github.com/organizations/armenzg-dev/settings/apps/new? [trimmed parameters]
```

[2]
image


src/docs/integrations/github.mdx
```diff
@@ -12,9 +12,9 @@ To configure the GitHub integration you'll need to create a GitHub app and obtai
 
 <CreateGitHubAppForm url="https://github.com/organizations/:org/settings/apps/new?name=:org-Sentry-Integration&public=false&members=read&emails=read&administration=read&contents=read&issues=write&pull_requests=write&repository_hooks=write&url=:url-prefix&callback_url=:url-prefix%2Fauth%2Fsso%2F&setup_url=:url-prefix%2Fextensions%2Fgithub%2Fsetup%2F&webhook_url=:url-prefix%2Fextensions%2Fgithub%2Fwebhook%2F&events[]=push&events[]=pull_request&webhook_active=true" defaultOrg="your-organization" defaultUrlPrefix="https://your-sentry-url-prefix.com" />
 
-Start by following GitHub's [official guide on creating a GitHub App](https://developer.github.com/apps/building-github-apps/creating-a-github-app/).
+NOTE: The button above will likely give you a 404. Adjust the url by changing `your-organization` for your Github org and try again.
```
armenzg, line 15: Ideally we should fix [this component](https://github.com/getsentry/develop/blob/master/src/components/createGitHubAppForm.tsx) instead of adding this warning.

ethanhs: I think fixing the component is a much better solution. What is the issue with it currently? Does it not replace the orgSlug correctly?

PR #711, opened 2022-10-13
Spec HTTP client errors

marandaneto:
Draft PR for this feature on Android - OkHttp https://github.com/getsentry/sentry-java/pull/2287


src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
+
+* SDKs won't capture error events if `captureFailedRequests` is disabled.
+* SDKs won't capture error events if the HTTP Request url isn't a match for any of the `failedRequestsTargets`.
+* SDKs won't capture error events if the HTTP Response status code isn't within the defined ranges in `failedRequestStatusCode`.
+
+The SDKs should capture error events with the following properties:
+
+The Request interface, See the <Link to="/sdk/event-payloads/request/">Spec</Link> for details.
+
+The Response context (Not yet under the Spec):
```
marandaneto, line 254: Linking to the Spec when Relay is done, See [RFC](https://github.com/getsentry/rfcs/blob/main/text/0022-response-context.md).

brustolin: I believe the response content would also be something nice to have.

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
+
+```json
+{
+  "contexts": {
+    "response": {
+      "type": "response",
+      "cookies": "PHPSESSID=298zf09hf012fh2; csrftoken=u32t4o3tb3gg43; _gat=1;",
+      "headers": {
```
brustolin, line 262: Cookies are also in the headers, should we duplicate this information?

marandaneto: This protocol already exists, send headers as it is, and cookies also in the specific field.

src/docs/sdk/features.mdx
marandaneto: This is the response content.
If you mean the request, it's linked in the original spec (line above), where there is the JSON content, I don't want to duplicate content, one more thing to maintain and to keep up to date.
This JSON example is going to be removed once this comment https://github.com/getsentry/develop/pull/711/files#r994681788 is addressed but it's not dependent on me.

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
+
+HTTP Client errors are automatically captured and sent to `https://sentry.io`.
+
+The HTTP Client integration should have 3 configuration options:
+
+* `captureFailedRequests` defaults to `false` due to PII reasons.
+* `failedRequestStatusCode` defaults to `500` to `599`, this configuration option accepts a `List` of `HttpStatusCodeRange` which is a range of HTTP status code -> `min` to `max` or a single `status_code`.
+* `failedRequestsTargets` defaults to `.*`, this configuration option accepts a `List` of `String` that may be Regular expressions as well, similar to <Link to="/sdk/performance/#tracepropagationtargets">tracePropagationTargets</Link>.
+
+Then the HTTP Client Integration should not capture the error:
```
romtsn, line 244:
```suggestion
Then the HTTP Client integration should not capture the error:
```
to align with the sentence above

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
+
+* SDKs won't capture error events if `captureFailedRequests` is disabled.
```
romtsn, line 246:
```suggestion
* If `captureFailedRequests` is disabled.
```
Since we say above "the Http Client integration should not capture the error" already, we probably don't need to add this part in the beginning

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
+* SDKs won't capture error events if the HTTP Request url isn't a match for any of the `failedRequestsTargets`.
+* SDKs won't capture error events if the HTTP Response status code isn't within the defined ranges in `failedRequestStatusCode`.
+
+The SDKs should capture error events with the following properties:
```
romtsn, line 250:
```suggestion
The HTTP Client integration should capture error events with the following properties:
```

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
 - when network error occurs, span status must be set to `internal_error`
 
+### HTTP Client errors
+
+HTTP Client errors are automatically captured and sent to `https://sentry.io`.
+
+The HTTP Client integration should have 3 configuration options:
+
+* `captureFailedRequests` defaults to `false` due to PII reasons.
+* `failedRequestStatusCode` defaults to `500` to `599`, this configuration option accepts a `List` of `HttpStatusCodeRange` which is a range of HTTP status code -> `min` to `max` or a single `status_code`.
```
romtsn, line 241:
```suggestion
* `failedRequestStatusCodes` defaults to `500` to `599`, this configuration option accepts a `List` of `HttpStatusCodeRange` which is a range of HTTP status code -> `min` to `max` or a single `status_code`.
```
Should this be code"s" since it's a list of ranges?

romtsn:
I noticed in the Android PR when dealing with headers you're also checking for sendDefaultPii + filtering out sensitive headers, should we mention it here in the spec as well? Otherwise LGTM, some minor wording things 👍

marandaneto:
> I noticed in the Android PR when dealing with headers you're also checking for sendDefaultPii + filtering out sensitive headers, should we mention it here in the spec as well? Otherwise LGTM, some minor wording things 👍

Nope, reason is here https://github.com/getsentry/sentry-java/pull/2287?notification_referrer_id=NT_kwDOAFd1vLI0NTc1NjU2MTY0OjU3MzE3NzI#discussion_r993200008
I just aligned the behavior with the Java repo, but other SDKs don't guard the `headers` with `defaultPii`.

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
 - span status must match HTTP response status code ([see Span status to HTTP status code mapping](/sdk/event-payloads/span/))
 - when network error occurs, span status must be set to `internal_error`
 
+### HTTP Client errors
+
+HTTP Client errors are automatically captured and sent to `https://sentry.io`.
+
+The HTTP Client integration should have 3 configuration options:
```
philipphofmann, line 238: `m`: I think it would be great if the description for each of the options below would not only point out the default values but also their functionality.

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
 - span status must match HTTP response status code ([see Span status to HTTP status code mapping](/sdk/event-payloads/span/))
 - when network error occurs, span status must be set to `internal_error`
 
+### HTTP Client errors
```
philipphofmann, line 234: `l`: Capital letters for titles.
```suggestion
### HTTP Client Errors
```

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
+
+HTTP Client errors are automatically captured and sent to `https://sentry.io`.
+
+The HTTP Client integration should have 3 configuration options:
+
+* `captureFailedRequests` defaults to `false` due to PII reasons.
+* `failedRequestStatusCodes` defaults to `500` to `599`, this configuration option accepts a `List` of `HttpStatusCodeRange` which is a range of HTTP status code -> `min` to `max` or a single `status_code`.
+* `failedRequestTargets` defaults to `.*`, this configuration option accepts a `List` of `String` that may be Regular expressions as well, similar to <Link to="/sdk/performance/#tracepropagationtargets">tracePropagationTargets</Link>.
+
+Then the HTTP Client integration should not capture the error:
```
philipphofmann, line 244: `l`: I would replace this everywhere.
```suggestion
Then the HTTP client integration should not capture the error:
```

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
+
+* If `captureFailedRequests` is disabled.
+* If the HTTP Request url isn't a match for any of the `failedRequestsTargets`.
```
philipphofmann, line 247: `l`: Typos
```suggestion
* If the HTTP request URL isn't a match for any of the `failedRequestTargets`.
```

src/docs/sdk/features.mdx
philipphofmann, line 238: `m`: On Cocoa, we don't expose integration options. We use the global SentryOptions for everything. Is this a typical pattern more SDKs use? If yes, we should also add it to Cocoa then. If not, I think we should change to option names, so they reflect they relate to the HTTP client error integration.

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
+* If the HTTP Response status code isn't within the defined ranges in `failedRequestStatusCodes`.
+
+The HTTP Client integration should capture error events with the following properties:
+
+The Request interface, See the <Link to="/sdk/event-payloads/request/">Spec</Link> for details.
```
philipphofmann, line 252: `l`: I think there is no reason to use capitals for these. The same applies to the next couple of lines.
```suggestion
The request interface, see the <Link to="/sdk/event-payloads/request/">spec</Link> for details.
```

src/docs/sdk/features.mdx
philipphofmann, line 244: `m`: Formulating this when the integration should capture an error would avoid double negative statements below and improve the readability.

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
+```
+
+The Exception Interface, See the <Link to="/sdk/event-payloads/exception/">Spec</Link> for details.
+
+If the HTTP Client integration does not throw an error for unsuccessful requests, you can create a synthetic exception following this Spec:
+
+* Set the <Link to="/sdk/event-payloads/exception/#exception-mechanism">Exception Mechanism</Link> with a proper `type` such as `SentryOkHttpInterceptor`.
+* Set the <Link to="/sdk/event-payloads/stacktrace/">Stack Trace Interface</Link> with `snapshot=true`.
+
+When capturing error events, pass the original `Request` and `Response` objects from the HTTP Client as `hints`, so the users may filter out events with the full context.
```
philipphofmann, line 280: `l`: I would mention `beforeSend`. What SDKs not having hints? Is it fine to not do this?
```suggestion
When capturing error events, pass the original `Request` and `Response` objects from the HTTP Client as `hints`, so the users may filter out events in `beforeSend` with the full context.
```

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
+  }
+}
+```
+
+The Exception Interface, See the <Link to="/sdk/event-payloads/exception/">Spec</Link> for details.
+
+If the HTTP Client integration does not throw an error for unsuccessful requests, you can create a synthetic exception following this Spec:
+
+* Set the <Link to="/sdk/event-payloads/exception/#exception-mechanism">Exception Mechanism</Link> with a proper `type` such as `SentryOkHttpInterceptor`.
+* Set the <Link to="/sdk/event-payloads/stacktrace/">Stack Trace Interface</Link> with `snapshot=true`.
```
philipphofmann, line 278: I don't really understand this. Why should the integration throw errors?

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
+
+When capturing error events, pass the original `Request` and `Response` objects from the HTTP Client as `hints`, so the users may filter out events with the full context.
+
+Those events can be searchable and alertable with the `http.url` and `http.status_code` properties, learn more about it in the [Searchable Properties](https://docs.sentry.io/product/sentry-basics/search/searchable-properties/) docs.
```
philipphofmann, line 282: `m`: Which events are those events?

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
+* `failedRequestStatusCodes` defaults to `500` to `599`, this configuration option accepts a `List` of `HttpStatusCodeRange` which is a range of HTTP status code -> `min` to `max` or a single `status_code`.
+* `failedRequestTargets` defaults to `.*`, this configuration option accepts a `List` of `String` that may be Regular expressions as well, similar to <Link to="/sdk/performance/#tracepropagationtargets">tracePropagationTargets</Link>.
+
+Then the HTTP Client integration should not capture the error:
+
+* If `captureFailedRequests` is disabled.
+* If the HTTP Request url isn't a match for any of the `failedRequestsTargets`.
+* If the HTTP Response status code isn't within the defined ranges in `failedRequestStatusCodes`.
+
+The HTTP Client integration should capture error events with the following properties:
```
philipphofmann, line 250: `m`: What is the title and content of the error event? Is it the exception itself?

philipphofmann (commented):
Thanks for writing down the spec and driving this new feature, @marandaneto. I have a couple of NITs and a few questions.

src/docs/sdk/features.mdx
marandaneto: That depends on the original or synthetic exception/stack trace, I will add a note about the title I've used.

src/docs/sdk/features.mdx
marandaneto: Other places use `Client` as uppercase and have been reviewed by the docs team, so I will keep it as it is.

src/docs/sdk/features.mdx
marandaneto: The difference is that Cocoa that thru Swizzling so no need of adding interceptors manually, here it is needed so it makes more sense to be per integration.
I can have an HTTP Client that captures events and the other one don't because I don't want to, a global config in the SentryOptions would not allow me to do that.

src/docs/sdk/features.mdx
marandaneto: Changing `See` to `see`, but will keep `Request`, docs team always keep them as Uppercase.

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,58 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
 - when network error occurs, span status must be set to `internal_error`
 
+### HTTP Client errors
+
+HTTP Client errors are automatically captured and sent to `https://sentry.io`.
+
+The HTTP Client integration should have 3 configuration options:
+
+* `captureFailedRequests` defaults to `false` due to PII reasons.
+* `failedRequestStatusCodes` defaults to `500` to `599`, this configuration option accepts a `List` of `HttpStatusCodeRange` which is a range of HTTP status code -> `min` to `max` or a single `status_code`.
```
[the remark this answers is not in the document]
marandaneto: I added as an option, so if the language allows, its a good idea.

romtsn (approved):

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,57 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
 
+### HTTP Client Errors
+
+The SDK automatically captures HTTP Client errors and sends them to [sentry.io](https://sentry.io).
+
+The HTTP Client integration should have 3 configuration options:
+
+- `captureFailedRequests` defaults to `false` due to PII reasons.
+  - The SDK will only capture HTTP Client errors if it is enabled.
+- `failedRequestStatusCodes` defaults to `500 - 599`, this configuration option accepts a `List` of `HttpStatusCodeRange` which is a range of HTTP status code -> `min` to `max` or a single `status_code`.
```
philipphofmann, line 242: `m`: A range can also be from 501 to 501. It would simplify the `HttpStatusCodeRange` class. Also noted that here https://github.com/getsentry/sentry-java/pull/2287/files#r998089883.
```suggestion
- `failedRequestStatusCodes` defaults to `500 - 599`, this configuration option accepts a `List` of `HttpStatusCodeRange` which is a range of HTTP status codes from `min` to `max`.
```

src/docs/sdk/features.mdx
philipphofmann: We will use swizzling, so most likely, we will add this to the global options. But I think the names should also be fine on the global options. They are kind of clear. We only need proper code docs.

src/docs/sdk/features.mdx
philipphofmann: I still would mention `beforeSend`.

src/docs/sdk/features.mdx
philipphofmann: I don't see the note, or am I missing something?

philipphofmann (commented):
One note about the range, apart from that great job @marandaneto 👏

src/docs/sdk/features.mdx
marandaneto: Its there, https://develop-git-chore-http-client-errors.sentry.dev/sdk/features/#http-client-errors
As a subitem of the stack trace interface, the title example.

src/docs/sdk/features.mdx
marandaneto: I disagree, just more boilerplate/confusing for the user, let's wait for the opinion of others.

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,57 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
 - when network error occurs, span status must be set to `internal_error`
 
+### HTTP Client Errors
+
+The SDK automatically captures HTTP Client errors and sends them to [sentry.io](https://sentry.io).
+
+The HTTP Client integration should have 3 configuration options:
+
+- `captureFailedRequests` defaults to `false` due to PII reasons.
+  - The SDK will only capture HTTP Client errors if it is enabled.
```
philipphofmann, line 241: `l`: Having a list with only one item kinda looks weird. I think we could just merge them into one bullet point. This applies to all items in this list.
```suggestion
- `captureFailedRequests` defaults to `false` due to PII reasons. The SDK will only capture HTTP Client errors if it is enabled.
```

src/docs/sdk/features.mdx
```diff
@@ -231,6 +231,57 @@ If Performance Monitoring is both supported by the SDK and enabled in the client
...
+}
+```
+
+The Exception Interface, see the <Link to="/sdk/event-payloads/exception/">Spec</Link> for details.
+
+If the HTTP Client integration does not throw an exception for unsuccessful requests, you can create a synthetic exception following this Spec:
+
+- Set the <Link to="/sdk/event-payloads/exception/#exception-mechanism">Exception Mechanism</Link> with a proper `type` such as `SentryOkHttpInterceptor`.
+- Set the <Link to="/sdk/event-payloads/stacktrace/">Stack Trace Interface</Link> with `snapshot=true`.
+  - The Exception message can be `Event was captured because the request status code was $code`.
```
philipphofmann, line 277: `l`: Again only one list item looks a bit weird. I would also change the `can` to a `should`
```suggestion
- Set the <Link to="/sdk/event-payloads/stacktrace/">Stack Trace Interface</Link> with `snapshot=true`. The Exception message should be `Event was captured because the request status code was $code`.
```

philipphofmann, line 277: `h`: I'm not convinced by the exception message. I think we should instead go for a short title, like `HTTP Client Error with status code: $code`

philipphofmann (commented):
Sorry I completely forgot to ask: How should these errors be grouped together?

src/docs/sdk/features.mdx
marandaneto: Since this is `l` and it's the develop docs and not user-facing docs, I'd keep it as it is.

src/docs/sdk/features.mdx
marandaneto: Yeah, I agree, it's better, will change it, thanks.

marandaneto:
> Sorry I completely forgot to ask: How should these errors be grouped together?

Stack trace normally, unless we see that this is not a good idea and can be improved later, but that's how Sentry works today, there's no special strategy unless you set fingerprints which can be done on eg `beforeSend`.

philipphofmann:
Thanks, @marandaneto, for being patient with all my questions.

PR #723, opened 2022-10-18
ref: use sentry's setup-volta for better caching

asottile-sentry:
this saves about ~30 seconds from the build time by caching yarn packages


AbhiPrasad (approved):
TIL, thanks for the patch!

PR #728, opened 2022-10-20
feat: app.in_foreground and thread.main context flags

markushi:
Relevant rendered parts:
https://develop-git-feat-app-in-foreground-and-threading.sentry.dev/sdk/event-payloads/contexts/#app-context
https://develop-git-feat-app-in-foreground-and-threading.sentry.dev/sdk/event-payloads/threads/

Adds `in_foreground` to the App Context event payload, as well as `main` to the Thread Context. 
As these are all similar but not the same and given that we already have some thread flags this PR tries to clear up their meanings and make their purpose hopefully more understandable.

@marandaneto I'm adding you as an early reviewer as I suspect you have some strong opinions about this 😊 

Related mobile issue: https://github.com/getsentry/team-mobile/issues/36

### DIFF
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -360,6 +360,10 @@ The `type` and default key is `"app"`.
 
 : _Optional_. Amount of memory used by the application in bytes.
 
+`in_foreground`
+
+: _Optional_. A flag indicating whether the app is in foreground or not. An app is in foreground when it's visible to the user.
+
 ## Browser Context
 
 Browser context carries information about the browser or user agent for
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -360,7 +360,7 @@ The `type` and default key is `"app"`.
 
 : _Optional_. Amount of memory used by the application in bytes.
 
-`app_in_foreground`
+`in_foreground`
 
 : _Optional_. A flag indicating whether the app is in foreground or not. An app is in foreground when it's visible to the user.
 
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -360,7 +360,7 @@ The `type` and default key is `"app"`.
 
 : _Optional_. Amount of memory used by the application in bytes.
 
-`in_foreground`
+`app_in_foreground`
 
 : _Optional_. A flag indicating whether the app is in foreground or not. An app is in foreground when it's visible to the user.
 
--- a/src/docs/sdk/event-payloads/threads.mdx
+++ b/src/docs/sdk/event-payloads/threads.mdx
@@ -26,7 +26,8 @@ attribute to cross-reference this thread.
 
 `crashed`
 
-: _Optional_. A flag indicating whether the thread was the cause for the event being sent, e.g. due to a crash. Defaults to `false`.
+: _Optional_. A flag indicating whether the thread was the cause for the event being sent, e.g. due to a crash.
+Defaults to `false`.
 
 `current`
 
@@ -36,7 +37,7 @@ Defaults to `false`.
 `ui`
 
 : _Optional_. If applicable, a flag indicating whether the thread was responsible for rendering the user interface.
-Defaults to `false`. On mobile platforms this is oftentimes referred to as the "main thread" or "ui thread".
+On mobile platforms this is oftentimes referred to as the "main thread" or "ui thread".
 
 `name`
 
--- a/src/docs/sdk/event-payloads/threads.mdx
+++ b/src/docs/sdk/event-payloads/threads.mdx
@@ -33,6 +33,11 @@ attribute to cross-reference this thread.
 : _Optional_. A flag indicating whether the thread was in the foreground. A thread is in foreground when it's executing.
 Defaults to `false`.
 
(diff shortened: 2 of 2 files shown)


jjbayer:
> This cannot go to the spec alone, this has to be implemented on Relay first. AFAIK, Relay drops unknown fields in some parts of the protocol, and it has to be explicitly typed out.

Yes, feel free to open a Relay PR with fields added in these two places:

https://github.com/getsentry/relay/blob/c5ff3c349f3ade03f03e582ab3061e9e205d9610/relay-general/src/protocol/contexts/app.rs#L10

https://github.com/getsentry/relay/blob/c5ff3c349f3ade03f03e582ab3061e9e205d9610/relay-general/src/protocol/thread.rs#L98

src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -360,6 +360,10 @@ The `type` and default key is `"app"`.
 
 : _Optional_. Amount of memory used by the application in bytes.
 
+`in_foreground`
```
markushi, line 363:
```suggestion
`app_in_foreground`
```

markushi:
> > This cannot go to the spec alone, this has to be implemented on Relay first. AFAIK, Relay drops unknown fields in some parts of the protocol, and it has to be explicitly typed out.
> 
> Yes, feel free to open a Relay PR with fields added in these two places:
> 
> [getsentry/relay@`c5ff3c3`/relay-general/src/protocol/contexts/app.rs#L10](https://github.com/getsentry/relay/blob/c5ff3c349f3ade03f03e582ab3061e9e205d9610/relay-general/src/protocol/contexts/app.rs#L10)
> 
> [getsentry/relay@`c5ff3c3`/relay-general/src/protocol/thread.rs#L98](https://github.com/getsentry/relay/blob/c5ff3c349f3ade03f03e582ab3061e9e205d9610/relay-general/src/protocol/thread.rs#L98)

Relay PR is now open -> https://github.com/getsentry/relay/pull/1578

marandaneto:
> > > This cannot go to the spec alone, this has to be implemented on Relay first. AFAIK, Relay drops unknown fields in some parts of the protocol, and it has to be explicitly typed out.
> > 
> > 
> > Yes, feel free to open a Relay PR with fields added in these two places:
> > [getsentry/relay@`c5ff3c3`/relay-general/src/protocol/contexts/app.rs#L10](https://github.com/getsentry/relay/blob/c5ff3c349f3ade03f03e582ab3061e9e205d9610/relay-general/src/protocol/contexts/app.rs#L10)
> > [getsentry/relay@`c5ff3c3`/relay-general/src/protocol/thread.rs#L98](https://github.com/getsentry/relay/blob/c5ff3c349f3ade03f03e582ab3061e9e205d9610/relay-general/src/protocol/thread.rs#L98)
> 
> Relay PR is now open -> [getsentry/relay#1578](https://github.com/getsentry/relay/pull/1578)

Nice, thanks for doing this.
Btw, Not a blocker for this PR, but I believe you should sync with Search and Storage team to make it searchable (`in_foreground`) on the Issues page, maybe even Discover and Alerts as well.
Also, The `ui` thread info won't be visible in the UI, so we'd need to sync with the UI team as well, otherwise, this alone won't deliver customer value.

marandaneto (approved):

src/docs/sdk/event-payloads/threads.mdx
```diff
@@ -26,13 +26,19 @@ attribute to cross-reference this thread.
...
+: _Optional_. A flag indicating whether the thread was the cause for the event being sent, e.g. due to a crash.
+Defaults to `false`.
 
 `current`
 
-: _Optional_. A flag indicating whether the thread was in the foreground.
+: _Optional_. A flag indicating whether the thread was in the foreground. A thread is in foreground when it's executing.
 Defaults to `false`.
 
+`ui`
```
marandaneto, line 37: Before merging this, should we call it `main`?
Reason is here https://github.com/getsentry/rfcs/pull/36#discussion_r1024461690

philipphofmann: If I'm not mistaken, you could have multiple UI threads if your application supports multiple windows or screens. For that reason, I think keeping it as UI thread makes sense.

src/docs/sdk/event-payloads/threads.mdx
```diff
@@ -26,13 +26,19 @@ attribute to cross-reference this thread.
...
+
+: _Optional_. If applicable, a flag indicating whether the thread was responsible for rendering the user interface.
+On mobile platforms this is oftentimes referred to as the "main thread" or "ui thread".
```
philipphofmann, line 40: Should we set a default of `false`, or should the default be `null`?

philipphofmann (approved):
One small question about the default of the UI property. Apart from that LGTM 🙏

src/docs/sdk/event-payloads/threads.mdx
philipphofmann: I agree with @marandaneto, we should rename this to main thread. Android, iOS, .NET MAUI, etc. call this main thread. We could also have multiple main threads for applications with multiple windows.

src/docs/sdk/event-payloads/threads.mdx
markushi: Good point, IMO I would default to `null` in this case, as a `false` default value may is not meaningful on some platforms.

PR #731, opened 2022-10-25
chore(mobile): Update Start-Up crashes docs

romtsn:


src/docs/sdk/features.mdx
```diff
@@ -224,6 +224,10 @@ If the platform allows it, the SDK may call flush directly after the detected st
 application terminates. If the SDK can guarantee transmission to Sentry while crashing, it can skip creating a marker file
 and making a blocking flush call on the next initialization.
 
+If the marker file exists upon the next SDK initialization, the SDK should clear the marker and block the `init` execution up
```
romtsn, line 227: I noticed this was actually missing, so decided to add it, wdyt @philipphofmann ? Also, is this true for iOS, that the SDK will continue flushing on a background thread when the timeout is exceeded?

philipphofmann (approved):
Thanks for the extra context.

src/docs/sdk/features.mdx
philipphofmann: yes it is.

PR #732, opened 2022-10-25
Update otel algo to read map

AbhiPrasad:
We need to read from a map instead from scope in case we mistakenly create the wrong parent-child relationship.

### DIFF
--- a/src/docs/sdk/performance/opentelemetry.mdx
+++ b/src/docs/sdk/performance/opentelemetry.mdx
@@ -63,7 +63,7 @@ class SentrySpanProcessor implements SpanProcessor {
 
     const otelSpanId = otelSpan.spanContext().spanId;
 
-    const parentSpan = scope.getSpan();
+    const parentSpan = this._map[otelSpanId];
     if (parentSpan) {
       // create a new span as a child of the parentSpan
       const childSpan = parentSpan.startChild({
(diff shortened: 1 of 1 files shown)


src/docs/sdk/performance/opentelemetry.mdx
```diff
@@ -63,7 +63,7 @@ class SentrySpanProcessor implements SpanProcessor {
 
     const otelSpanId = otelSpan.spanContext().spanId;
 
-    const parentSpan = scope.getSpan();
+    const parentSpan = this._map[otelSpanId];
     if (parentSpan) {
```
AbhiPrasad, line 67:
```suggestion
    const mapVal = this._map[otelSpanId];
    if (mapVal) {
    const [parentSpan] = mapVal;
```

src/docs/sdk/performance/opentelemetry.mdx
```diff
@@ -63,8 +63,9 @@ class SentrySpanProcessor implements SpanProcessor {
 
     const otelSpanId = otelSpan.spanContext().spanId;
 
-    const parentSpan = scope.getSpan();
-    if (parentSpan) {
+    const mapVal = this._map[otelSpanId];
+    if (mapVal) {
+    const [parentSpan] = mapVal;
```
AbhiPrasad, line 68:
```suggestion
      const [parentSpan] = mapVal;
```

PR #734, opened 2022-10-25
Add info on self-hosted beacon

hubertdeng123:
Adds some info on the self-hosted beacon from old docs that seem to no longer exist

### DIFF
--- a/src/docs/self-hosted/index.mdx
+++ b/src/docs/self-hosted/index.mdx
@@ -12,7 +12,27 @@ Our recommendation is to download the [latest release of the self-hosted reposi
 
 When you run `./install.sh`, you have a choice to opt in or out of our monitoring. This monitoring is used for development and debugging purposes so that we're on top of issues you're facing, allowing us to provide a more seamless installation process. For more details please see the [section in the self-hosted README](https://github.com/getsentry/self-hosted#self-hosted-monitoring).
 
-Note that choosing whether to send errors or not will become mandatory beginning with the 22.10.0 release.
+Note that choosing whether to send errors or not will become mandatory beginning with the 22.11.0 release.
+
+### Self-hosted Beacon
+
+Self-hosted will periodically communicate with a remote beacon server. This is utilized for a couple of things, primarily:
+- Getting information about the current version of Sentry
+- Retrieving important system notices
+The remote server is operated by the Sentry team (sentry.io), and the information reported follows the company's [privacy policy](https://sentry.io/privacy/).
+
+The following information is reported:
+1. A unique installation ID
+2. The version of Sentry
+3. A technical contact email (system.admin-email)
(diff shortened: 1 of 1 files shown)


src/docs/self-hosted/index.mdx
```diff
@@ -12,7 +12,27 @@ Our recommendation is to download the [latest release of the self-hosted reposi
 
 When you run `./install.sh`, you have a choice to opt in or out of our monitoring. This monitoring is used for development and debugging purposes so that we're on top of issues you're facing, allowing us to provide a more seamless installation process. For more details please see the [section in the self-hosted README](https://github.com/getsentry/self-hosted#self-hosted-monitoring).
 
-Note that choosing whether to send errors or not will become mandatory beginning with the 22.10.0 release.
+Note that choosing whether to send errors or not will become mandatory beginning with the 22.11.0 release.
+
+### Self-hosted Beacon
+
+Self-hosted will periodically communicate with a remote beacon server. This is utilized for a couple of things, primarily:
```
ethanhs, line 19: This is a configured option at install time, right? Right now this reads as a always-on feature whereas I think it is important to be clear it is opt-in.

Maybe something like 
```suggestion
If you opt-in to it, self-hosted Sentry will periodically communicate with a remote beacon server. This is utilized for a couple of things, primarily:
```

src/docs/self-hosted/index.mdx
```diff
@@ -12,7 +12,27 @@ Our recommendation is to download the [latest release of the self-hosted reposi
...
+
+If you opt-in to it, self-hosted Sentry will periodically communicate with a remote beacon server. This is utilized for a couple of things, primarily:
+- Getting information about the current version of Sentry
+- Retrieving important system notices
+The remote server is operated by the Sentry team (sentry.io), and the information reported follows the company's [privacy policy](https://sentry.io/privacy/).
+
+The following information is reported:
+1. A unique installation ID
+2. The version of Sentry
+3. A technical contact email (system.admin-email)
```
ethanhs, line 27: This is also opt in right? Based on your other PR that seems to be the case?

hubertdeng123: actually yes, the anonymity of this is opt-in I'll also include that. Good catch

hubertdeng123: added the note that sending contact info is opt in and it can always be changed in the settings of the admin panel

ethanhs: Did you push that? I don't think I see that change

hubertdeng123: oh 🤦

hubertdeng123: now it's pushed haha

ethanhs (approved):

PR #737, opened 2022-10-27
Update otel specs for span attribute mapping

mydea:
Small update to the otel specs. Otel span attributes should be mapped to sentry span data, as attributes can be arbitrary - so mapping them to tags (which have to be strings) is tricky.


sl0thentr0py (approved):

mydea:
After checking with @AbhiPrasad I also removed the `status` -> `tags` mapping - the status should only be mapped to the status (which is already defined anyhow).

AbhiPrasad (approved):

PR #743, opened 2022-11-02
feat(otel): Make status mapping example use TS for consistency

mydea:
This one example was written in go, I think it makes sense for all examples to be written in the same language for consistency. Since I've implemented the same logic in TS in the meanwhile, we can update this here as well.


AbhiPrasad (approved):
I had copied the approach from https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/13407 in golang, but converting to JS sounds good to me!

PR #746, opened 2022-11-02
Added configuration for not tracing HTTP OPTION requests by default

antonpirker:
SDKs should not create transactions for HTTP `OPTION` requests by default, but the user should have the possibility to opt into to creating transactions for `OPTION` requests.


cleptric (approved):
LGTM

PR #759, opened 2022-11-14
Add another way to clean up disc space for postgres

hubertdeng123:
As per discussion on discord, this adds documentation for another way to clean up postgres for self-hosted

### DIFF
--- a/src/docs/self-hosted/troubleshooting.mdx
+++ b/src/docs/self-hosted/troubleshooting.mdx
@@ -114,6 +114,14 @@ docker-compose run -T web cleanup --days 7 -m nodestore -l debug
 docker-compose run -T postgres bash -c "apt update && apt install -y --no-install-recommends postgresql-9.6-repack && su postgres -c 'pg_repack -E info -t nodestore_node'"
 ```
 
+If the script above still does not clear up enough disc space, you can try the following in postgres. This will lead to data loss in event details. _SENTRY_RETENTION_DAYS_ will need to be filled in manually.
+```shell
+DELETE FROM public.nodestore_node WHERE “timestamp” < NOW() - INTERVAL ‘SENTRY_RETENTION_DAYS’;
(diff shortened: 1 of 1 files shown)


src/docs/self-hosted/troubleshooting.mdx
```diff
@@ -114,6 +114,14 @@ docker-compose run -T web cleanup --days 7 -m nodestore -l debug
 docker-compose run -T postgres bash -c "apt update && apt install -y --no-install-recommends postgresql-9.6-repack && su postgres -c 'pg_repack -E info -t nodestore_node'"
 ```
 
+If the script above still does not clear up enough disc space, you can try the following in postgres. This will lead to data loss in event details. _SENTRY_RETENTION_DAYS_ will need to be filled in manually.
```
chadwhitacre, line 117:
```suggestion
If the script above still does not clear up enough disk space, you can try the following in Postgres. This will lead to data loss in event details. _SENTRY_RETENTION_DAYS_ will need to be filled in manually.
```

src/docs/self-hosted/troubleshooting.mdx
```diff
@@ -114,6 +114,14 @@ docker-compose run -T web cleanup --days 7 -m nodestore -l debug
...
+```shell
+DELETE FROM public.nodestore_node WHERE “timestamp” < NOW() - INTERVAL ‘SENTRY_RETENTION_DAYS’;
```
chadwhitacre, line 119: Make it a little clearer that this is a value the user needs to provide?

```suggestion
DELETE FROM public.nodestore_node WHERE “timestamp” < NOW() - INTERVAL ‘[SENTRY_RETENTION_DAYS]’;
```

chadwhitacre:
A couple small suggestions, otherwise looks good.

chadwhitacre (approved):

Issue #647, opened 2022-07-15
Document `UI Event Transaction`


bruno-garcia:
This feature was introduced to some Mobile SDKs. It's also known as _User Interaction Tracking".

The goal is for the SDK to create a transaction when the user interacts with the app. These are based on 'idle timeout' since there's no clear indication of when the transaction ended.

The general tracking issue for it:
* https://github.com/getsentry/team-mobile/issues/4

Docs:
* [Android docs](https://docs.sentry.io/platforms/android/performance/instrumentation/automatic-instrumentation/#user-interaction-instrumentation)
* [iOS docs](https://docs.sentry.io/platforms/apple/guides/ios/performance/instrumentation/automatic-instrumentation/#user-interaction-instrumentation)

We need to write the guideline for SDK engineers. Some of the operations used are already documented, and we could link to it: https://develop.sentry.dev/sdk/performance/span-operations/#mobile

![image](https://user-images.githubusercontent.com/1633368/179299479-b07c92f7-dd49-4ec7-81fa-cc3be1110be4.png)

PR #766, opened 2022-11-17
Add UI event transactions

philipphofmann:
Add specification for UI event transactions with Gherkin syntax.

Fixes GH-647

### DIFF
--- a/src/docs/sdk/performance/ui-event-transactions.mdx
+++ b/src/docs/sdk/performance/ui-event-transactions.mdx
@@ -48,24 +48,17 @@ Scenario: Wait for children when starting span
     Then the SDK cancels the idle timeout
     And waits for the child to finish
 
-Scenario: Schedule timeout when the last span finishes
+Scenario: Schedule idle timeout when the last span finishes
     Given a UI event transaction
     And the transaction has one or multiple running child spans
     And the transaction is waiting for its children to finish
     When the SDK finishes the last child span
-    Then the SDK starts the idle timeout
-
-Scenario: Don't reset timeout when the second last span finishes
-    Given a UI event transaction
-    And the transaction has two running child spans
-    When the SDK finishes the first child span
-    Then the SDK doesn't reset the idle timeout
-    And waits for the children to finish
+    Then the SDK schedules the idle timeout 
 
 Scenario: Discard UI event transactions without child spans
     Given a UI event transaction
     And the transaction has no child spans
-    When the idleTimeout times out
+    When the idle timeout times out
     Then the SDK discards the transaction 
 
 Scenario: Set time to last finished child span
@@ -88,7 +81,7 @@ Scenario: Same UI element with same event
     Given an ongoing UI event transaction
     When the user triggers the same UI event with the same type for the same
         UI element
-    Then the SDK resets the timeout
+    Then the SDK reschedule the idle timeout
     And doesn't create a new transaction
 
 # If your SDK binds auto-generated transactions to the scope see binding
@@ -100,7 +93,7 @@ Scenario: Same UI element with different event
     Then the SDK finishes the ongoing transaction
     And sets the status to OK
     And waits for the children to finish
-    And cancels the timeout
+    And cancels the idle timeout
     And starts a new transaction
 ```
 
@@ -119,7 +112,7 @@ Scenario: Same UI element with different event
     Then the SDK finishes the ongoing transaction
     And sets the status to OK
     And waits for the children to finish
-    And cancels the timeout
+    And cancels the idle timeout
     And removes the ongoing transaction from the scope
     And starts a new transaction
     And puts the new transaction on the scope
--- a/src/docs/sdk/performance/ui-event-transactions.mdx
+++ b/src/docs/sdk/performance/ui-event-transactions.mdx
@@ -1,0 +1,173 @@
+---
+title: "UI Event Transactions"
+---
+
+We recommend implementing this feature for mobile and desktop SDKs.
+
+UI event transactions aim to capture transactions based on user interactions, such as clicks,
+scroll events, pinches, etc. The UI events the SDK can hook into might vary depending on the
+platform. UI event transactions are an expansion to transactions generated automatically by
+the SDK, which are referred to as auto-generated transactions in this document. 
+
+Creating transactions for a single UI event without any spans would be fruitless. Instead, the
+SDK shall use a UI event as the entry point to a possibly meaningful transaction. It should
+wait to see if it can add any auto-generated spans to the transaction. If it can, it shall keep
+and send the transaction. If not, the SDK should discard the empty transaction. A combination of
+two concepts is needed to implement the wait-and-see logic: __idle transactions__ and
+__wait-for-children__. Check out the specification below to see how those two concepts work
+together with covering multiple edge cases. The following description is drastically simplified:
+
+* __Idle-transactions__: The SDK starts an idle timeout for the transaction when starting it.
+When the transaction or any of its spans starts a new span, it resets the timeout. The SDK finishes
+the transaction when the timeout succeeds.
+* __Wait-for-children__: A transaction waits for all its child spans to finish before finishing itself.
+
+Before diving into the specification with all the edge cases, let's look at two simple examples:
+
+1. The user clicks a button that triggers some requests to the backend and stores data in the
+local database. The SDK can create a meaningful transaction in that case.
+2. The user clicks a button that solely validates some form data and triggers nothing that the
+SDK can automatically instrument. The SDK could only create a transaction without any spans, which would be valueless.
+
+# Specification
+
+Users can change the `idleTimeout` via the SDK config options. The default is 3.0 seconds. 
+
+The specification is written in the [Gherkin syntax](https://cucumber.io/docs/gherkin/reference/).
+
+**1. Starting UI Event transactions** 
+```
+Given an instrumentable UI event
+Then the SDK starts a UI event transaction 
+And starts the idle timeout with idleTimeout of the options
+```
+
+**5. Wait for children when starting span**
+```
+Given a UI event transaction
+When the SDK starts a child span
+Then the SDK cancels the idle timeout
+And waits for the child to finish
+```
+
+**4. Start timeout when the last span finishes**
+```
+Given a UI event transaction
+And the transaction has one or multiple running child spans
+And the transaction is waiting for its children to finish
+When the SDK finishes the last child span
+Then the SDK starts the idle timeout
+```
+
+**3. Don't reset timeout when the second last span finishes**
+```
+Given a UI event transaction
+And the transaction has two running child spans
+When the SDK finishes the first child span
+Then the SDK doesn't reset the idle timeout
+And waits for the children to finish
+```
+
+**2. Discard UI event transactions without child spans**
+```
+Given a UI event transaction
+And the transaction has no child spans
+When the idleTimeout times out
+Then the SDK discards the transaction 
+```
+
+**3. Set time to last finished child span**
+```
+Given a UI event transaction
+And the transaction has one finished child span
+And the transaction has one running child span
+When the running child span finishes
+Then the SDK finishes the transaction
+And trims the end time of the transaction to the one of the last finished child span
+```
+
+**6. Same UI element with same event**
+```
+Given an ongoing UI event transaction
+When the user triggers the same UI event with the same type for the same UI element
+Then the SDK resets the timeout
+And doesn't create a new transaction
+```
+
+**10. Same UI element with different event**
+If your SDK binds transactions to the scope see 11.
+```
+Given an ongoing UI event transaction
+When the user triggers the same UI element with a different event
+Or the user triggers a different UI element
+Then the SDK finishes the ongoing transaction
+And sets the status to OK
+And waits for the children to finish
+And cancels the timeout
+And starts a new transaction
+```
+
+## Binding to scope
+
+**11. Same UI element with different event**
+```
+Given an ongoing UI event transaction
+When the user triggers the same UI element with a different event
+Or the user triggers a different UI element
+Then the SDK finishes the ongoing transaction
+And sets the status to OK
+And waits for the children to finish
+And cancels the timeout
+And removes the ongoing transaction from the scope
+And starts a new transaction
+And puts the new transaction on the scope
+```
+
+**11. UI event triggered but transaction ended**
+```
+Given an auto-generated transaction from a any event
+And the transaction finished
+When the user triggers the same event
+Then the SDK starts a new transaction
+```
+
+**6. Ongoing screen load transaction**
+```
+Given an ongoing screen load transaction
+When the SDK starts a new UI event transaction
+Then the SDK doesn't bind the new UI event transaction to the scope
+```
+
+**7. Ongoing UI event transaction**
+```
+Given an ongoing UI event transaction
+When the SDK creates a new screen load transaction
+Then the SDK finishes the ongoing UI event transaction
+And removes it from the scope
+And sets the status to canceled
+And waits for its children to finish
+```
+
+**8. Manually created transaction on the scope**
+```
+Given an ongoing manually transaction bound to the scope by the user
+When the SDK creates a new auto-generated transaction
+Then the SDK doesn't add the new auto-generated transaction to the scope 
+```
+
+## Transaction name
+
+**12. Transaction name**
+```
+When the user clicks a button
+Then the SDK starts a transaction
+And the transaction name is the [[screen.name](http://screen.name/)](http://screen.name) + [[view.id](http://view.id/)](http://view.id) or the accessibility identifier of the view (e.g. `LoginActivity.login_button`)
+```
+
+**13. Transaction name [[view.id](http://view.id/)](http://view.id) or accessibility identifier not specified**
+```
+When the user clicks a button
+And the button has no [[view.id](http://view.id/)](http://view.id) or accessibility identifier
+Then the SDK doesn't start a transaction
+And prints a warning to the console
+```
--- a/src/docs/sdk/performance/ui-event-transactions.mdx
+++ b/src/docs/sdk/performance/ui-event-transactions.mdx
@@ -132,7 +132,7 @@ Scenario: UI event triggered but transaction ended
 Scenario: Manually created transaction bound to the scope
     Given an ongoing manually created transaction by the user bound to the
         scope
-    When the user triggers the a UI event
+    When the user triggers a UI event
     Then the SDK doesn't start a UI event transaction
 ```
 
--- a/src/docs/sdk/performance/ui-event-transactions.mdx
+++ b/src/docs/sdk/performance/ui-event-transactions.mdx
@@ -17,7 +17,7 @@ two concepts is needed to implement the wait-and-see logic: __idle transactions_
 __wait-for-children__. Check out the specification below to see how those two concepts work
 together with covering multiple edge cases. The following description is drastically simplified:
 
-* __Idle-transactions__: The SDK starts an idle timeout for the transaction when starting it.
+* __Idle-transactions__: The SDK schedule an idle timeout for the transaction when starting it.
 When the transaction or any of its spans starts a new span, it resets the timeout. The SDK finishes
 the transaction when the timeout succeeds.
 * __Wait-for-children__: A transaction waits for all its child spans to finish before finishing itself.
@@ -40,7 +40,7 @@ The specification is written in the [Gherkin syntax](https://cucumber.io/docs/gh
 Scenario: Starting UI Event transactions
     Given an instrumentable UI event
     Then the SDK starts a UI event transaction 
-    And starts the idle timeout with idleTimeout of the options
+    And schedule the idle timeout with idleTimeout of the options
 
 Scenario: Wait for children when starting span
     Given a UI event transaction
@@ -48,7 +48,7 @@ Scenario: Wait for children when starting span
     Then the SDK cancels the idle timeout
     And waits for the child to finish
 
-Scenario: Start timeout when the last span finishes
+Scenario: Schedule timeout when the last span finishes
     Given a UI event transaction
     And the transaction has one or multiple running child spans
     And the transaction is waiting for its children to finish
--- a/src/docs/sdk/performance/ui-event-transactions.mdx
+++ b/src/docs/sdk/performance/ui-event-transactions.mdx
@@ -119,7 +119,7 @@ Scenario: Same UI element with different event
     And puts the new transaction on the scope
 
 Scenario: UI event triggered but transaction ended
-    Given an auto-generated transaction from a any UI event
+    Given an auto-generated transaction from any UI event
     And the transaction already  is finished
     When the user triggers the same UI event
     Then the SDK starts a new UI event transaction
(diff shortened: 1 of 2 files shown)


src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,173 @@
...
+Then the SDK starts the idle timeout
+```
+
+**3. Don't reset timeout when the second last span finishes**
+```
+Given a UI event transaction
+And the transaction has two running child spans
+When the SDK finishes the first child span
+Then the SDK doesn't reset the idle timeout
+And waits for the children to finish
```
marandaneto, line 68: H: This is different than Android and Flutter, actually, when a new span is created, the timer is reset only.
@romtsn to confirm.

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+**3. Set time to last finished child span**
+```
+Given a UI event transaction
+And the transaction has one finished child span
+And the transaction has one running child span
+When the running child span finishes
+Then the SDK finishes the transaction
+And trims the end time of the transaction to the one of the last finished child span
+```
```
marandaneto, line 87: Android and Flutter have an option for that called `trimEnd` and it is set by default for idle transactions.

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+**10. Same UI element with different event**
+If your SDK binds transactions to the scope see 11.
+```
+Given an ongoing UI event transaction
+When the user triggers the same UI element with a different event
+Or the user triggers a different UI element
+Then the SDK finishes the ongoing transaction
+And sets the status to OK
```
marandaneto, line 104: I know this is already implemented, but it seems like the status could be `canceled` or something?
Also, the status is `ok` only if the user has not set a status manually.

src/docs/sdk/performance/ui-event-transactions.mdx
philipphofmann: Thanks for the info, but I guess it's OK to keep this? If you have `trimEnd` you are safe.

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,160 @@
...
+    When the SDK starts a new UI event transaction
+    Then the SDK doesn't bind the new UI event transaction to the scope
+```
+
+## Transaction name
+
+The user should be able to identify to which UI element the UI event transaction belongs on which screen
+by only looking at the transaction name. Choose whatever works best for your specific platform. Be
+attentive not to use any PII in the transaction name. One recommendation is 
+`screen name + view identifier || accessibility identifier`. On Android, this would map to
```
marandaneto, line 159: `accessibility identifier` may have PII because of screen readers, do we guard this against `sendDefaultPii`?

src/docs/sdk/performance/ui-event-transactions.mdx
philipphofmann: On Cocoa, we have `ok`, but we can change it.
```Swift
func test_DifferentUIElement_DifferentAction_FinishesTransaction() {
    let view1 = fixture.button
    callExecuteAction(action: "otherAction", target: fixture.target, sender: view1, event: TestUIEvent())
    
    let firstTransaction = try! XCTUnwrap(SentrySDK.span as? SentryTracer)
    
    let view2 = UIButton()
    callExecuteAction(action: action, target: fixture.target, sender: view2, event: TestUIEvent())
    
    assertFinishesTransaction(firstTransaction, operationClick)
}

private func assertFinishesTransaction(_ transaction: SentryTracer, _ operation: String) {
    XCTAssertTrue(transaction.isFinished)
    XCTAssertEqual(.ok, transaction.context.status)
    assertTransaction(name: "SentryTests.FirstViewController.\(expectedAction)", operation: operation)
    
    let transactions = getInternalTransactions()
    XCTAssertEqual(1, transactions.count)
}
```

https://github.com/getsentry/sentry-cocoa/blob/944bebeafd548a2cb391f8e6b2047ed37eec3755/Tests/SentryTests/Integrations/UIEvents/SentryUIEventTrackerTests.swift#L186-L196

https://github.com/getsentry/sentry-cocoa/blob/944bebeafd548a2cb391f8e6b2047ed37eec3755/Tests/SentryTests/Integrations/UIEvents/SentryUIEventTrackerTests.swift#L274

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,160 @@
...
+    Given an auto-generated transaction from a any UI event
+    And the transaction already  is finished
+    When the user triggers the same UI event
+    Then the SDK starts a new UI event transaction
+
+Scenario: Manually created transaction bound to the scope
+    Given an ongoing manually created transaction by the user bound to the
+        scope
+    When the SDK creates a new UI event transaction
+    Then the SDK doesn't add the UI event transaction to the scope
```
marandaneto, line 130: In this case, if the transaction is not in the scope, there will be no spans for it since the user can't access it nor the auto instrumentation, doesn't that mean that the transaction won't have spans and won't be sent?

src/docs/sdk/performance/ui-event-transactions.mdx
marandaneto: For context, accessibility id was a problem for https://docs.sentry.io/platforms/react-native/touchevents/ and we needed to provide a way out, so it can't be used by default.

src/docs/sdk/performance/ui-event-transactions.mdx
philipphofmann: Basically, yes, we could drop it and not start it. That's an improvement. Good catch, thanks 🙏 . @romtsn, I think we could change that in our SDKs. WDYT?

src/docs/sdk/performance/ui-event-transactions.mdx
philipphofmann: Doesn't Android also use `waitForChildren` @romtsn ?

src/docs/sdk/performance/ui-event-transactions.mdx
marandaneto: I guess it makes sense if it's not an Idle transaction (for platforms that you know that the loading is finished), but not with the idle transaction concept.

src/docs/sdk/performance/ui-event-transactions.mdx
romtsn: we only use the view identifier

src/docs/sdk/performance/ui-event-transactions.mdx
marandaneto: There's `waitForChildren`, but the approach here is different if I understood it correctly, it does reset the timer every time that a new child is created and so on, at the end, the result is the same, I guess.

src/docs/sdk/performance/ui-event-transactions.mdx
marandaneto: Maybe mention that some SDKs have the `trimEnd`, but you can also skip it and use the timestamp of the last finished child span as well.

src/docs/sdk/performance/ui-event-transactions.mdx
marandaneto: On Android, if I understood it correctly, it forces to `ok`.
@romtsn to confirm.

src/docs/sdk/performance/ui-event-transactions.mdx
marandaneto: @philipphofmann so I'd rather remove the `accessibility identifier` recommendation or guard it against `sendDefaultPii ` in the spec.
Also, you don't always have an identifier, in this case, we should write in the spec that you don't start a transaction at all.

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+Users can change the `idleTimeout` via the SDK config options. The default is 3.0 seconds. 
+
+The specification is written in the [Gherkin syntax](https://cucumber.io/docs/gherkin/reference/).
+
+```Gherkin
+Scenario: Starting UI Event transactions
+    Given an instrumentable UI event
+    Then the SDK starts a UI event transaction 
+    And starts the idle timeout with idleTimeout of the options
```
romtsn, line 43:
```suggestion
    And schedules to finish with idleTimeout of the options
```
`l`: maybe this is clearer a bit?

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+Scenario: Wait for children when starting span
+    Given a UI event transaction
+    When the SDK starts a child span
+    Then the SDK cancels the idle timeout
+    And waits for the child to finish
+
+Scenario: Start timeout when the last span finishes
```
romtsn, line 51: `l`: Maybe we should use `schedule` everywhere instead of `start`?

src/docs/sdk/performance/ui-event-transactions.mdx
romtsn: I guess I'm confused a bit by the word `reset` - does it mean `cancel` or does it mean `reschedule`? If we mean `reschedule` here, I think we can remove this scenario, because it should be covered by the one above ("when the _last_ span finishes")

That's what we do on Android (if all children are finished - we reschedule the finish timer):
https://github.com/getsentry/sentry-java/blob/84f0ef1abb55bccd90315665ff4e307c5f71886c/sentry/src/main/java/io/sentry/SentryTracer.java#L234-L245

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,160 @@
...
+    And the transaction has no child spans
+    When the idleTimeout times out
+    Then the SDK discards the transaction 
+
+Scenario: Set time to last finished child span
+    Given a UI event transaction
+    And the transaction has one finished child span
+    And the transaction has one running child span
+    When the running child span finishes
+    Then the SDK finishes the transaction
```
romtsn, line 76:
```suggestion
    Then the SDK schedules the idle timeout
    And the SDK finishes the transaction after timeout
```
`m`: maybe this will help better understand why we need to trim the end time - because transaction is not finished immediately with the last child span

src/docs/sdk/performance/ui-event-transactions.mdx
romtsn: yes we use `ok`. I'm not sure about `canceled` because it doesn't sound like the user is cancelling/aborting their action - e.g. if they scrolled through twitter feed and found some interesting tweet and clicked to expand the thread - is it really cancelling the scrolling gesture, or just finishing the gesture logically, because there was an interesting thing requiring another gesture type?

`canceled` would make sense to me, if e.g. user clicked a button which triggered some network request, and this request was very slow to display the content, and the user pressed back/home button because they don't wanna wait anymore. Then it's for sure `canceled`, but we don't track system buttons yet.

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,160 @@
...
+    Then the SDK resets the timeout
+    And doesn't create a new transaction
+
+# If your SDK binds auto-genrated transactions to the scope see binding
+# to scope
+Scenario: Same UI element with different event
+    Given an ongoing UI event transaction
+    When the user triggers the same UI element with a different event
+    Or the user triggers a different UI element
+    Then the SDK finishes the ongoing transaction
```
romtsn, line 93: `h`: I think it's important to mention here that we also remove it from scope (on Android we do), because calling `finish` doesn't necessarily trigger the immediate finish due to `waitForChildren`

```suggestion
    Then the SDK finishes the ongoing transaction
    And removes the transaction from scope
```

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,160 @@
...
+        child span
+
+Scenario: Same UI element with same event
+    Given an ongoing UI event transaction
+    When the user triggers the same UI event with the same type for the same
+        UI element
+    Then the SDK resets the timeout
+    And doesn't create a new transaction
+
+# If your SDK binds auto-genrated transactions to the scope see binding
```
romtsn, line 87:
```suggestion
# If your SDK binds auto-generated transactions to the scope see binding
```

src/docs/sdk/performance/ui-event-transactions.mdx
romtsn: ahh, just noticed it's in the different section, sorry my bad!

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,160 @@
...
+    Then the SDK finishes the ongoing transaction
+    And sets the status to OK
+    And waits for the children to finish
+    And cancels the timeout
+    And removes the ongoing transaction from the scope
+    And starts a new transaction
+    And puts the new transaction on the scope
+
+Scenario: UI event triggered but transaction ended
+    Given an auto-generated transaction from a any UI event
```
romtsn, line 121:
```suggestion
    Given an auto-generated transaction from any UI event
```
is it a typo or my english is not that advanced? 🤣

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,160 @@
...
+    And the transaction already  is finished
```
romtsn, line 122:
```suggestion
    And the transaction is already finished
```

src/docs/sdk/performance/ui-event-transactions.mdx
romtsn: yep, I agree we should mention that - we don't start any transactions if we're unable to find the view id. Can be improved, but usually the views that trigger action always have some ids on them to attach listeners to them.

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,160 @@
...
+    Then the SDK finishes the ongoing UI event transaction
+    And removes it from the scope
+    And sets the status to canceled
+    And waits for its children to finish
+
+Scenario: Ongoing screen load transaction
+    Given an ongoing screen load transaction
+    When the SDK starts a new UI event transaction
+    Then the SDK doesn't bind the new UI event transaction to the scope
+```
```
romtsn, line 152: @marandaneto I guess we can also add a navigation transaction here (we have it for compose, and you for flutter). Or is it considered a screen load transaction?

src/docs/sdk/performance/ui-event-transactions.mdx
romtsn: yeah we could do that, I guess that would work for any kind of transaction on the scope - if there's something, we just don't start the UI transaction at all

romtsn (commented):
We also have a boolean option in SentryOptions (enableUserInteractionTracing), should we mentioned it here?

src/docs/sdk/performance/ui-event-transactions.mdx
marandaneto: If the agreement was `ok`, let's keep it as it is, but Android should only set a status if there's no status, right now its overwriting the user's status in case it's set, @romtsn mind filling a bug for this?

src/docs/sdk/performance/ui-event-transactions.mdx
marandaneto: @romtsn depends on the platform, it's not that common having an `id` (key) in Flutter unless people use it for UI tests or something because you set the listeners directly in the declaration.

src/docs/sdk/performance/ui-event-transactions.mdx
marandaneto: Yeah, a Timer is scheduled.

src/docs/sdk/performance/ui-event-transactions.mdx
marandaneto: Should we raise a meta issue for that?

src/docs/sdk/performance/ui-event-transactions.mdx
marandaneto: I guess it can be both, it's just a concept but they behave in the same way.
`ui.load` when you know when to start/likely when to finish.
`navigation` when it's a router/navigation API.

src/docs/sdk/performance/ui-event-transactions.mdx
marandaneto: Btw just a small info, status `canceled` does not degrade the failure rate of transactions.
https://docs.sentry.io/product/performance/metrics/#failure-rate

romtsn: https://github.com/getsentry/sentry-java/issues/2378 filed

src/docs/sdk/performance/ui-event-transactions.mdx
romtsn: https://github.com/getsentry/team-mobile/issues/71

src/docs/sdk/performance/ui-event-transactions.mdx
philipphofmann: `trimEnd` what happens on Android if you disable it for that scenario, @romtsn?

src/docs/sdk/performance/ui-event-transactions.mdx
philipphofmann: It's `reschedule`. 

>  I think we can remove this scenario, because it should be covered by the one above

Yes, it's covered. I removed it.

src/docs/sdk/performance/ui-event-transactions.mdx
philipphofmann: @marandaneto, on Cocoa, we use the method the UI element calls. I added this now. I also mention that `Choose whatever works best for your specific platform.`. Do you think that's fine or should I add more information, @marandaneto?

marandaneto: @philipphofmann https://github.com/getsentry/sentry-cocoa/blob/efeb69e9c50860c600d68640b1f465cc83a1a7f7/Sources/Sentry/SentryUIEventTracker.m#L130 and https://github.com/getsentry/sentry-cocoa/blob/d9757459a4a288ad57b2a8348def9eb839cbe72b/Sources/Sentry/SentryBreadcrumbTracker.m#L219
This can be PII and isn't guarded with `defaultPii`

src/docs/sdk/performance/ui-event-transactions.mdx
romtsn: It won't set the transaction end time to the latest child end time, so it will be set to `latest child end timestamp + idleTimeout`

romtsn: but worth noting, it's not configurable from within SentryOptions, so it's internal on Android and is more of a convenience flag for us if we decide not to trim

marandaneto: Yeah, tbh `trimEnd` might be just overengineering, I guess we can skip this.

src/docs/sdk/performance/ui-event-transactions.mdx
marandaneto: > @marandaneto, on Cocoa, we use the method the UI element calls. I added this now. I also mention that `Choose whatever works best for your specific platform.`. Do you think that's fine or should I add more information, @marandaneto?

Yes, all good for the spec, but I guess we should raise an issue to guard the linked code against `sendDefaultPii` or remove it altogether, what do you think?

marandaneto (approved):
LGTM @philipphofmann and once more, thank you for doing this.

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,166 @@
...
+    Given an ongoing UI event transaction
+    When the SDK creates a new screen load transaction
+    Then the SDK finishes the ongoing UI event transaction
+    And removes it from the scope
+    And sets the status to canceled
+    And waits for its children to finish
+
+Scenario: Ongoing screen load transaction
+    Given an ongoing screen load transaction
+    When the SDK starts a new UI event transaction
```
romtsn, line 154: @philipphofmann I guess we should also change it to `does not start a new UI event transaction` similar to Manually create transactions

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,166 @@
...
+```Gherkin
+Scenario: Ongoing UI event transaction
+    Given an ongoing UI event transaction
+    When the SDK creates a new screen load transaction
+    Then the SDK finishes the ongoing UI event transaction
+    And removes it from the scope
+    And sets the status to canceled
+    And waits for its children to finish
+
+Scenario: Ongoing screen load transaction
```
romtsn, line 152:
```suggestion
Scenario: Ongoing screen load/navigation transaction
```

src/docs/sdk/performance/ui-event-transactions.mdx
philipphofmann: I will raise issues, yes. 🥇 for finding them.

philipphofmann: https://github.com/getsentry/sentry-cocoa/issues/2406

src/docs/sdk/performance/ui-event-transactions.mdx
philipphofmann: Typo thanks 😁🙏

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,168 @@
...
+    When the SDK creates a new screen load transaction
+    Then the SDK finishes the ongoing UI event transaction
+    And removes it from the scope
+    And sets the status to canceled
+    And waits for its children to finish
+
+Scenario: Ongoing screen load/navigation transaction
+    Given an ongoing screen load/navigation transaction
+    When the SDK starts a new UI event transaction
+    Then the SDK doesn't start a UI event transaction
```
romtsn, line 157:
```suggestion
    Given an ongoing screen load/navigation transaction
    When the user triggers the a UI event
    Then the SDK doesn't start a UI event transaction
```

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -0,0 +1,168 @@
...
+Scenario: UI event triggered but transaction ended
+    Given an auto-generated transaction from any UI event
+    And the transaction is already finished
+    When the user triggers the same UI event
+    Then the SDK starts a new UI event transaction
+
+Scenario: Manually created transaction bound to the scope
+    Given an ongoing manually created transaction by the user bound to the
+        scope
+    When the user triggers the a UI event
```
romtsn, line 135:
```suggestion
    When the user triggers a UI event
```

romtsn (approved):
2 minor fixes + still missing `enableUserInteractionTracing` option. Otherwise, LGTM 👍 thanks for doing that

PR #776, opened 2022-12-06
add ui.ttid to span operations

stefanosiano:
We are going to add a ttid (time to initial display) span to the automatic activity transactions.
The use case is mainly for mobile developers.
The current activity transaction duration is not reliable to have an understanding of how much time it took to actually load an activity, due to possible other spans added during the transaction.

### DIFF
--- a/src/docs/sdk/performance/span-operations.mdx
+++ b/src/docs/sdk/performance/span-operations.mdx
@@ -181,6 +181,7 @@ Serverless related spans are expected to follow OpenTelemetry's [Function as a S
 |            | app.start.cold            |                               |
 | ui         |                           | An operation on a mobile UI   |
 |            | ui.load                   |                               |
+|            | ui.ttid                   |                               |
 |            | ui.action                 |                               |
 |            | ui.action.click           |                               |
 |            | ui.action.swipe           |                               |
--- a/src/docs/sdk/performance/span-operations.mdx
+++ b/src/docs/sdk/performance/span-operations.mdx
@@ -182,6 +182,7 @@ Serverless related spans are expected to follow OpenTelemetry's [Function as a S
 | ui         |                           | An operation on a mobile UI   |
 |            | ui.load                   |                               |
 |            | ui.load.initial_display   |                               |
+|            | ui.load.full_display      |                               |
 |            | ui.action                 |                               |
 |            | ui.action.click           |                               |
 |            | ui.action.swipe           |                               |
--- a/src/docs/sdk/performance/span-operations.mdx
+++ b/src/docs/sdk/performance/span-operations.mdx
@@ -181,7 +181,7 @@ Serverless related spans are expected to follow OpenTelemetry's [Function as a S
 |            | app.start.cold            |                               |
 | ui         |                           | An operation on a mobile UI   |
 |            | ui.load                   |                               |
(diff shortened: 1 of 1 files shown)


stefanosiano:
@AbhiPrasad @marandaneto @philipphofmann 
Android's implementation is in [this pr](https://github.com/getsentry/sentry-java/pull/2369/)
Is there anything to consider to add this new span?

src/docs/sdk/performance/span-operations.mdx
```diff
@@ -181,6 +181,7 @@ Serverless related spans are expected to follow OpenTelemetry's [Function as a S
 |            | app.start.cold            |                               |
 | ui         |                           | An operation on a mobile UI   |
 |            | ui.load                   |                               |
+|            | ui.ttid                   |                               |
```
marandaneto, line 184: Should that be part of `app.start`? or every screen would have its own `ttid`?

stefanosiano: every screen it's going to have its own `ttid`
we'll also add the `ttfd` (time to full display) to every screen, too, but that's for another pr

AbhiPrasad (approved):
I think the operation makes sense to me

src/docs/sdk/performance/span-operations.mdx
philipphofmann: I think we should mention what ttid stands for = time to initial display. Maybe we could also call it
`ui.load.inital` `ui.load.full`? Most developers won't understand what ttid stands for and they have to go to our docs.

romtsn: could be even `ui.load.initial_draw` and `ui.load.full_draw`. But I agree `ttid` is probably not well-known, so I'd also vote for changing it

marandaneto: That would be semantically wrong I guess, `initial_draw` would mean it's drawn but that refers only to the very first frame only.
`full_draw` would mean it's fully drawn, but it's not only about drawing but rather waiting for async content and redrawing it.
so maybe `initial_display` and `full_dislay`? it's pretty much the wording for both acronyms

stefanosiano: so let's go with `ui.load.initial_display` and `ui.load.full_display`? These make sense to me

stefanosiano: @philipphofmann does it make sense to you on ios?

philipphofmann: Yes, that makes sense. I'm happier now with this name as it's self-explanatory.

philipphofmann:
@stefanosiano, please also add `ui.load.full_display`.

stefanosiano:
> @stefanosiano, please also add `ui.load.full_display`.

In this same pr? Shouldn't we wait after implementing it?
Or are we going to keep this pr unmerged until the sdk implement it?

philipphofmann:
I think it's OK to merge it already, even if we don't have the implementation. We have already decided on the name. We can always change the docs again.

philipphofmann (commented):
LGTM

philipphofmann (approved):
LGTM

PR #786, opened 2022-12-19
chore(js): Remove unused packages

evanpurkhiser:
These were not being used 🤷 

Notably removing `gatsby-plugin-mermaid` since it pulls in a TON of stuff, including pupeteer, which requires chromium. It's just a LOT for something that we can just embed as a SVG with a link to the source.


evanpurkhiser:
uhg. I lied, we are using mermaid

priscilawebdev (approved):

PR #790, opened 2022-12-19
build: Bump to node 16

evanpurkhiser:
This is required since volta does not provide m1 builds of node 14


package.json
```diff
@@ -77,7 +77,7 @@
     "react-test-renderer": "^16.13.1"
   },
   "volta": {
-    "node": "14.20.1",
-    "yarn": "1.22.17"
+    "node": "16.16.0",
+    "yarn": "1.22.5"
```
evanpurkhiser, line 81: Matches sentry yarn 🤷

joshuarli (approved):

PR #798, opened 2022-12-30
docs(application): Add options documentation

gggritso:
There's appetite for updated `options` documentation. There's a smorgasbord of documents about `options` in Notion, but a lot of them are stale and/or confusing. This adds a dedicated `options` page, and adds it to the sidebar.

### DIFF
--- a/src/docs/feature-flags.mdx
+++ b/src/docs/feature-flags.mdx
@@ -76,7 +76,7 @@ default_manager.add('organizations:test-feature', OrganizationFeature)
 
 Flagr has significant latency, and is somewhat flakey. If you're working in
 high throughput areas like Ingest or Relay, Flagr is not fast enough or
-reliable enough. In these cases, you should use [options](/options) instead.
+reliable enough. In these cases, you should use [options](/options/) instead.
 
 ### Add it to the Organization Model Serializer
 
--- a/src/docs/feature-flags.mdx
+++ b/src/docs/feature-flags.mdx
@@ -68,7 +68,7 @@ add a third optional boolean parameter when adding the feature, for example:
 default_manager.add('organizations:test-feature', OrganizationFeature, True)
 ```
 
-If you don't plan to use flagr don't pass this third parameter, for example: 
+If you don't plan to use flagr don't pass this third parameter, for example:
 
 ```python
 default_manager.add('organizations:test-feature', OrganizationFeature)
@@ -192,7 +192,7 @@ flags are then configured via `sentry.conf.py`. For Sentry's SaaS deployment,
 Flagr is used to configure flags in production.
 
 If you want to enable your feature for a subset of production users, you will
-need to set up your feature in Flagr. If you haven't already make sure that 
+need to set up your feature in Flagr. If you haven't already make sure that
 when you add your flag in sentry that you passed the third option so that Flagr
 knows to check this feature in production.
 
@@ -404,7 +404,7 @@ clicking **Delete Flag** at the bottom of the page.
 ## Getsentry feature handlers
 
 Getsentry contains a variety of feature handlers that override the
-`SENTRY_FEATURES` map. 
+`SENTRY_FEATURES` map.
 
 ### Plan specific features
 
@@ -418,7 +418,7 @@ you need to:
 ### Custom handlers
 
 If your feature requires additional logic to become active, you can also
-implement a feature handler in getsentry, follow these steps:
+implement a feature handler in getsentry. For example, you can create a feature flag that is backed by [options](/options/). follow these steps:
 
 1. Disable the feature in `getsentry/conf/settings/defaults.py` by updating `SENTRY_FEATURES`.
 2. Add a new feature handler class in `getsentry/features.py` that determines availability of the feature based on the organization or actor.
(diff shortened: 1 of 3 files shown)


src/docs/feature-flags.mdx
```diff
@@ -68,7 +68,7 @@ add a third optional boolean parameter when adding the feature, for example:
 default_manager.add('organizations:test-feature', OrganizationFeature, True)
 ```
 
-If you don't plan to use flagr don't pass this third parameter, for example: 
+If you don't plan to use flagr don't pass this third parameter, for example:
```
wmak, line 71: Could we expand this or the previous paragraph to explain when not to use flagr (could take it from the notion) and to use options instead, then link to the new section?

src/docs/options.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+## Adding New Options
+
+To add a new system option, add it to [`sentry/options/defaults.py`](https://github.com/getsentry/sentry/blob/master/src/sentry/options/defaults.py) by registering it. e.g.,
+
+```python
+register("performance.some-feature-rate", default=0.0)
+```
+
+### Tips
```
wmak, line 15: Should we rephrase this from tips to rules on making an option? I'm thinking if I saw a PR with an option name that wasn't namespaced I would definitely comment on it

src/docs/options.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+- namespace your option (e.g., `feature-scope.feature-name`)
+- colocate your new option with related existing options
+- use dashes and not underscores
+- always set a default value with `default=`
+- the value of an option can be any pickleable value
+
+<Alert level="warning">
+  Always split up registering the new option and the code that relies on the new option. Register new options in their own pull requests, ahead of the code changes.
```
wmak, line 24: Could we link to two PRs that show this being done here?

src/docs/options.mdx
```diff
@@ -0,0 +1,126 @@
...
+</Alert>
+
+## Using Options
+
+To check the value of an option, import the `options` module and use the `get` method. e.g.,
```
wmak, line 29: Is it worth mentioning that `options.get` is cached already so there's no need to do it again?

src/docs/options.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+```python
+from sentry import options
+feature_rate = options.get("performance.some-feature-rate")
+```
+
+## Setting Options
+
+Options are stored in the database and are cached. You can change the value of an option using `sentry shell`, or by using the management UI.
```
wmak, line 38: Could we have `management UI` here link to the section below?

src/docs/options.mdx
```diff
@@ -0,0 +1,126 @@
...
+options.set("performance.some-feature-rate", 0.01)
+```
+
+<Alert level="info">
+If you do not have access to the shell, you'll need to contact OPS to set the option value for you.
+</Alert>
+
+### Management UI
+
+If you expect to frequently update your option, you can make it editable in the management UI. The `/admin/settings` page in the Sentry app has inputs for a small set of the available options. You can submit a pull request to add your option to the UI by [updating the corresponding React view](https://github.com/getsentry/sentry/blob/master/static/app/views/admin/options.tsx).
```
wmak, line 61: Can we mention that this is only available to superusers too?

src/docs/options.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+## Using Options for Feature Rollout
+
+If you're working on a system-wide feature, you may choose to use options for your rollout instead of feature flags. Unlike feature flags, options don't allow for easy segmentation, but they are performance, stable, and simple. e.g.,
```
wmak, line 65:
```suggestion
If you're working on a system-wide feature, you may choose to use options for your rollout instead of feature flags. Unlike feature flags, options don't allow for easy segmentation, but they are performant, stable, and simple to implement. e.g.,
```

wmak (commented):
Looks good to me, thanks again for adding this

src/docs/feature-flags.mdx
gggritso: Done! Added a note below

src/docs/options.mdx
gggritso: Hmm, what do you mean? Do what again?

src/docs/options.mdx
gggritso: Done!

src/docs/options.mdx
gggritso: Done!

src/docs/options.mdx
gggritso: Yep, done! And added some info about the required permissions

src/docs/options.mdx
gggritso: Thank you!

src/docs/options.mdx
wmak: To roll your own caching someone thinking they have to do

```
cache.set(some_key, option.get(...))
```

gggritso: Ah, okay! I'll move the caching/database bit from "Setting Options" a little higher!

PR #813, opened 2023-01-18
feat(unified-api): Define `includeLocalVariables` option

AbhiPrasad:
Currently the PHP, Python, Ruby and NodeJS SDKs support local variables, but they all use inconsistent names for the option that gates this functionality.

- Ruby: [`capture_exception_frame_locals`](https://docs.sentry.io/platforms/ruby/configuration/options)
- Python: [`with_locals`](https://docs.sentry.io/platforms/python/configuration/options/#with-locals)
- PHP no option
- NodeJS [`_experiments.includeStackLocals`](https://docs.sentry.io/platforms/node/configuration/options/#_experiments.include-stack-locals)

I propose we rename this to `includeLocalVariables` / `include_local_variables`, unify these across the SDKs and add them when we get time

We also should decide if `sendDefaultPii` should have an affect on this.

cc @timfish

### DIFF
--- a/src/docs/sdk/features.mdx
+++ b/src/docs/sdk/features.mdx
@@ -73,6 +73,8 @@ Lines of source code to provide context in stack traces. This is easier in inter
 
 Local variable names and values for each stack frame, where possible. Restrictions apply on some platforms, for example it’s may only be possible to collect the values of parameters passed into each function, or it may be completely impossible to collect this information at all.
 
+This functionality should be gated behind the `includeLocalVariables` option, which is `true` by default.
+
 ## Desymbolication
 
 Turn compiled or obfuscated code/method names in stack traces back into the original. Desymbolication always requires Sentry backend support. Not necessary for many languages.


antonpirker:
I like consistent naming, so I am in.

As this is only for errors should `error` be in the variable name? (bikesheding alarm)

sl0thentr0py:
let's keep the PII stuff separate, I'm working on a comprehensive proposal to solve this better.
If that gets rejected, we can gate locals behind `sendDefaultPII` later.

sl0thentr0py (approved):
fine with me, but for existing implementations, we'll have to alias to existing names, deprecate them and remove them later in a major.

AbhiPrasad:
> As this is only for errors should error be in the variable name

This is a good suggestion, but since I think local variables are a stacktrace concept, it should be clear to users that this only applies to errors.

antonpirker:
Python issue for this: https://github.com/getsentry/sentry-python/issues/1854

PR #814, opened 2023-01-18
Don't discourage declaring and using options immediately

gggritso:
This used to be a problem, but it no longer is. From https://github.com/getsentry/sentry/pull/43335#discussion_r1073800593


wedamija (approved):
Just for reference, the pr that fixed the old caching issue https://github.com/getsentry/sentry/pull/39799

PR #815, opened 2023-01-18
feat(event): Add source to SDK payload

AbhiPrasad:
This change adds the `source` field to the SDK interface that is part of the event payload.

We can use this source field to list the installation source of the SDK. For most cases this will be a package repository like `npm` or `pypi`, but it can also be things like `cdn`.

Current uses cases:

Browser Web:
- `npm`
- `cdn`
- `loader`

Apple:
- `spm`
- `carthage`
- `cocoapods`

### DIFF
--- a/src/docs/sdk/event-payloads/sdk.mdx
+++ b/src/docs/sdk/event-payloads/sdk.mdx
@@ -11,39 +11,43 @@ and transmit an event.
 `name`
 
 : **Required**. The name of the SDK. The format is `entity.ecosystem[.flavor]`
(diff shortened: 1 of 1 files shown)


Lms24 (approved):
LGTM!

philipphofmann (approved):
LGTM

AbhiPrasad:
FYI folks I'm reverting this https://github.com/getsentry/develop/pull/819

This is because we can probably just the the `source` part of the `source:indentifier` schema in the packages field to accomplish this.

Issue #816, opened 2023-01-19
Add span semantic conventions for span data


AbhiPrasad:
Fields that the product uses under `span.data`.

PR #817, opened 2023-01-19
feat(perf): Add span data conventions

AbhiPrasad:
Fixes https://github.com/getsentry/develop/issues/816

I would like us to start tracking how we are setting/using keys and values on the span data field. Ideally we move toward using more of [OpenTelemetry's OOO semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/24de67b3827a4e3ab2515cd8ab62d5bcf837c586/specification/trace/semantic_conventions/README.md) for this, but add our own values as appropriate.

This is important, as we should make sure our spans behave consistently across different sdks and platforms. In addition, it helps us make sure that changing fields in the SDK won't break product usage.

We can also then look to use OpenTelemetry's span attributes for some of Sentry's generated spans. For example, we currently store the `url` as a string value on `http.client` spans that are generated by the Browser JS SDK. We could move to using `http.url` instead, [which matches OpenTelemetry](https://github.com/open-telemetry/opentelemetry-specification/blob/24de67b3827a4e3ab2515cd8ab62d5bcf837c586/specification/trace/semantic_conventions/http.md#common-attributes).

This will help us more easily create performance detectors for OpenTelemetry data (as we have access to many different kind of spans and attributes), and this is also important for us to document moving forward so that we can iterate more efficiently.

### DIFF
--- a/src/docs/sdk/event-payloads/span.mdx
+++ b/src/docs/sdk/event-payloads/span.mdx
@@ -32,6 +32,8 @@ import "./properties/tags.mdx";
 
 : _Optional_. Arbitrary data associated with this Span.
 
+The semantic conventions for the `data` field are described in <Link to="/sdk/performance/span-data-conventions/">Sentry's Span Convention Documentation</Link>.
+
 ```json
 {
   "data": {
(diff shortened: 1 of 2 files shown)


AbhiPrasad:
Gonna go ahead and merge this in - @getsentry/performance would be great if we could keep this up to date with all the new performance issues coming out!

PR #827, opened 2023-02-01
feat(heroku): Add heroku plugin develop docs

ceorourke:
Add develop docs for the Heroku plugin.

### DIFF
--- a/src/docs/integrations/heroku.mdx
+++ b/src/docs/integrations/heroku.mdx
@@ -1,0 +1,22 @@
+---
+title: "Heroku Plugin"
+---
+
+## Create a Heroku application
+
+Create an account on [Heroku](http://www.heroku.com) and create an app.
+
+Follow the [instructions](https://devcenter.heroku.com/articles/git) to set up a Heroku app using the command line tool and deploy process.
+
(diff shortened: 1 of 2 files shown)


src/docs/integrations/heroku.mdx
```diff
@@ -0,0 +1,22 @@
+---
+title: "Heroku Plugin"
+---
+
+## Create a Heroku application
+
+Create an account on [Heroku](http://www.heroku.com) and create an app.
+
+Follow the [instructions](https://devcenter.heroku.com/articles/git) to set up a Heroku app using the command line tool and deploy process.
```
barkbarkimashark, line 9: small thing. maybe add some clarification about what setting up a Heroku app means. i.e. 'to set up a Heroku app(doesn't have to be a working webapp, a random text file would work)`

barkbarkimashark (approved):
looks good.

src/docs/integrations/heroku.mdx
ceorourke: Oh huh, when I was trying to deploy mine I actually had an issue getting it to work if it _wasn't_ a working webapp. I'm not sure I see the usefulness of deploying a text file, would that be solely for development purposes?

PR #829, opened 2023-02-06
beforeSend clarification

smeubank:
is this in dev docs somewhere already?


AbhiPrasad (approved):
No I think we don't have this explicitly written out like this. Let's add this!

PR #832, opened 2023-02-09
sdk.name for hybrid sdks

bruno-garcia:
This addresses the `sdk.name` and allows us to find out, for example: 

1. Events from React Native customers
2. Events from the Cocoa SDK (regardless of hybrid SDK that bundled it)
3. By looking at an event know exactly what SDK that is: It's cocoa version X but it was part of the Unity SDK

More importantly, this is simple, solves an actual problem we have right now. 
And we use this already on Unreal and Unity. Works as intended.

### DIFF
--- a/src/docs/sdk/event-payloads/sdk.mdx
+++ b/src/docs/sdk/event-payloads/sdk.mdx
@@ -19,6 +19,24 @@ ecosystem. Official Sentry SDKs use the _entity_ `sentry`. Examples:
 - `sentry.python`
 - `sentry.javascript.react-native`
 
+For SDKs that are composed of more than one Sentry SDK. For example, [the Unity SDK](https://github.com/getsentry/sentry-unity/issues/616)
+which includes a .NET layer as well as different native layers. When events come out
+of the native layer, it's important to distinguish from the stand-alone native SDK.
+We do that by appending the top SDK to the end of the native SDK name. Eamples:
+
+Unity SDK
+
+- `sentry.dotnet.unity` on events coming from C#. As based on the base spec above.
+  - `sentry.java.android.unity` on for events of the Java layer on Android
+  - `sentry.native.android.unity` on for events of the Native layer on Android
+  - `sentry.cocoa.unity` on for events from iOS/macOS layer
+  - `sentry.native.unity` on for events coming from `sentry.native` directly. Such as on Windows and Linux.
+
+Android SDK
+
+- `sentry.java.android` on events from the Java layer. Since the Android SDK is based on the Java SDK.
+- `sentry.native.android` on events from NDK. It's `sentry.native` but it's bundled in the Android SDK with some customization.
+
 `version`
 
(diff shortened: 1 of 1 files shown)


src/docs/sdk/event-payloads/sdk.mdx
```diff
@@ -19,6 +19,24 @@ ecosystem. Official Sentry SDKs use the _entity_ `sentry`. Examples:
 - `sentry.python`
 - `sentry.javascript.react-native`
 
+For SDKs that are composed of more than one Sentry SDK. For example, [the Unity SDK](https://github.com/getsentry/sentry-unity/issues/616)
+which includes a .NET layer as well as different native layers. When events come out
+of the native layer, it's important to distinguish from the stand-alone native SDK.
+We do that by appending the top SDK to the end of the native SDK name. Eamples:
```
bruno-garcia, line 25:
```suggestion
We do that by appending the top SDK to the end of the native SDK name. Examples:
```

src/docs/sdk/event-payloads/sdk.mdx
```diff
@@ -19,6 +19,24 @@ ecosystem. Official Sentry SDKs use the _entity_ `sentry`. Examples:
...
+
+Unity SDK
```
bruno-garcia, line 27:
```suggestion
Unity SDK:
```

src/docs/sdk/event-payloads/sdk.mdx
```diff
@@ -19,6 +19,24 @@ ecosystem. Official Sentry SDKs use the _entity_ `sentry`. Examples:
...
+
+- `sentry.dotnet.unity` on events coming from C#. As based on the base spec above.
```
bruno-garcia, line 29:
```suggestion
- `sentry.dotnet.unity` on events coming from C#. As based on the spec above.
```

bitsandfoxes (approved):
This worked really well for us so far.

mattjohnsonpint:
I'm good with this, but have a question:

How far along the stack should this extend?  Should we be appending `ecosystem.flavor`, or just `ecosystem`?

For example, if calling the Cocoa SDK from various .NET SDKs, appending `ecosystem.flavor` would result in:

- `sentry.dotnet` => `sentry.cocoa.dotnet`
- `sentry.dotnet.maui` => `sentry.cocoa.dotnet.maui`
- `sentry.dotnet.extensions.logging` => `sentry.cocoa.dotnet.extensions.logging`
- etc...

Whereas appending only `ecosystem would have them _all_ use `sentry.cocoa.dotnet`.

It would be great if the doc could clarify this point.  Thanks.

mattjohnsonpint:
... but now that I think about it, `unity` is a *flavor*, not an ecosystem.  Right?  So maybe there's no hard rule to follow?

bruno-garcia:
> ... but now that I think about it, unity is a flavor, not an ecosystem. Right? So maybe there's no hard rule to follow?

yeah I think this is accurate

PR #833, opened 2023-02-09
Update body/transfer size unit from string -> number

AbhiPrasad:
I made a mistake when initially adding this, amending!


k-fish (approved):
😄

PR #834, opened 2023-02-09
Add `debug_id` to stacktrace protocol

lforst:
Ref: https://github.com/getsentry/team-webplatform-meta/issues/17

This PR adds the `debug_id` field to the stacktrace protocol as implemented in https://github.com/getsentry/relay/pull/1832.


AbhiPrasad (commented):
Should we write the cardinality/type of this value also? Like we expect it to be a uuid?

lforst:
Be free

PR #836, opened 2023-02-13
Add Source Map Images to Debug Meta Interface

lforst:
Ref: https://www.notion.so/sentry/Debug-IDs-for-Source-Maps-Brainstorming-66601f1cd5f646a399f9b46fa0dfa3a5
Ref: https://github.com/getsentry/team-webplatform-meta/issues/17

This PR adds a new image type to the Debug Meta interface. The new `sourcemap` image type will be used to map filenames of JavaScript resources to a particular source map / source file on Sentry via a `debug_id`.

I am not really opinionated about what the fields should be called _exactly_ so if you have suggestions lmk.

### DIFF
--- a/src/docs/sdk/event-payloads/debugmeta.mdx
+++ b/src/docs/sdk/event-payloads/debugmeta.mdx
@@ -333,8 +333,8 @@ URL pulled from the `external_debug_info` custom section.
 
 `code_id`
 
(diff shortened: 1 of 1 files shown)


lforst:
Maybe `source_filename` is not a good name. Idk what else to call it though. `source_abs_path`?

Swatinem (approved):

PR #841, opened 2023-02-15
feat(profiling): Document the Sample format

phacops:
We defined a format to upload a profile. This is the document explaining how to build and format it.

### DIFF
--- a/src/docs/sdk/sample-format.mdx
+++ b/src/docs/sdk/sample-format.mdx
@@ -12,8 +12,6 @@ collected by a profiler.
 
 A profile should always be associated with a transaction.
 
-Here is a JSON Schema definition of the format.
-
 ## Data Model
 
 ### Example
@@ -293,10 +291,23 @@ in the `timestamp` field when the sample was captured.
 
 : *List, required*. Contains a list of frame indices forming a stack trace.
 
+We expect the stack to be from outermost frame to innermost (which means the
+root frame is at the bottom of the stack).
+
 `frames`
 
 : *List, required*. Contains a list of frame objects (see [Frame
 Attributes](/sdk/event-payloads/stacktrace/#frame-attributes)).
+
+Technically, all attributes are optional so we won't reject the payload if some
+are missing.
+
+Now, for native platforms (`cocoa` or `rust` for example),
+`instruction_addr` is required in order to be able to symbolicate and get more
+info about the frame.
+
+For the rest of the platforms, whatever is passed here will travel to the
+frontend and whatever is available can be used to show more or less information.
 
 `thread_metadata`
 
--- a/src/docs/sdk/sample-format.mdx
+++ b/src/docs/sdk/sample-format.mdx
@@ -313,7 +313,7 @@ detected during runtime.
 ## Validation
 
 We will reject a profile in Relay for several reasons:
-- profile data is missing
+- profile data is missing (no frame, no sample, no stack)
 - not enough samples (we need at least 2 samples)
 - no transaction associated with the profile
 - any of the metadata required is missing
--- a/src/docs/sdk/sample-format.mdx
+++ b/src/docs/sdk/sample-format.mdx
@@ -1,0 +1,286 @@
+---
+title: Sample format
+sidebar_order: 20
+---
+
+This document defines the Sample format used by Sentry to ingest profiles. This
+document is meant for Sentry SDK developers and maintainers of the profile
+ingestion pipeline.
+
+This format is meant to be cross platform and allow to represent samples
+collected by a profiler.
+
+A profile should always be associated with a transaction.
+
+## Data Model
+
+### Example
+
+```json
+{
+  "debug_meta": { ... },
+  "device": {
+    "architecture": "arm64e",
+    "is_emulator": true,
+    "locale": "en_US",
+    "manufacturer": "Apple",
+    "model": "iPhone14,8"
+  },
+  "environment": "development",
+  "event_id": "",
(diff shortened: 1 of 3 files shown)


src/docs/sdk/sample-format.mdx
```diff
@@ -0,0 +1,286 @@
...
+  },
+  "timestamp": "2023-01-01T00:00:00.000Z",
+  "transaction": {
+    "id": "",
+    "name": "",
+    "trace_id": "",
+    "active_thread_id": "",
+  },
+  "version": "1",
+  "profile": { ... },
```
JonasBa, line 51: Wdyt about adding one example of sample, weight etc here? Would be nice to be able to see the whole format in a single place.

could be like

```
"profile": {
  "samples": [{relative_ts_ns: 0, stack_id: 0, thread_id: 0}...]
},
```
and same for weights and stacks obviously

phacops: I describe the profiling data separately below https://github.com/getsentry/develop/pull/841/files#diff-f87ed348469343306eb1312b9ba48345f7e7657ba2ae3e62c4d4da89eb0a2b82R199.

I think separating the metadata from the raw data would help to make it easier to read through and reference.

JonasBa: Yeah, I understand - I like it when everything in the same place (even if it's not explained) so you minimize any user error from going back/forth between definitions (I find that useful when trying to validate the format). In any case, it's a short doc so not a big concern

phacops: Alright, I can expand the example to include those fields as well, it doesn't need to be separate.

Also, I'm planning to ship a JSON Schema definition to be able to validate the profile, I think that'd be helpful.

JonasBa: ➕  for JSON schema. Btw, I do wonder if it would make sense to document the validation constraints or reasons why a "valid" schema may still get dropped, e.g profiles with <= 1 samples.

phacops: Yes, very good point, that's something I need to add to this document.

src/docs/sdk/sample-format.mdx
```diff
@@ -0,0 +1,331 @@
...
+```
+
+`stacks`
+
+: *List, required*. Contains a list of frame indices forming a stack trace.
+
+`frames`
+
+: *List, required*. Contains a list of frame objects (see [Frame
+Attributes](/sdk/event-payloads/stacktrace/#frame-attributes)).
```
JonasBa, line 299: What does this reference btw? Would be nice to have contents of frames documented in this doc too. I only see instruction address in the sample above

phacops: For native platforms, the bare minimum is to have an instruction address. For platforms like Python or Node, I need a package, function, file names and line info. Now, we support all the attributes listed in the section "Frame Attributes".

I'll expand the explanation about this.

bruno-garcia:
@phacops can we rely on the details already provided here or is it pending significant changes?

src/docs/sdk/sample-format.mdx
```diff
@@ -0,0 +1,341 @@
...
+5. Check if that stack exists in the list of stacks and if it doesn't, add it to the list of stacks
+6. Create a sample containing the thread ID, the relative timestamp and the
+stack ID.
+
+`samples`
+
+: *List, required* Contains a list of sample object captured during execution.
+Each sample can contain the following attributes:
+
+- `elapsed_since_start_ns`: *String, required.* Contains the relative timestamp in nanoseconds since the value contained
```
Zylphrex, line 277: This should mention that it should be a `uint64` formatted as a string. `float` is not accepted.

Zylphrex:
Should this doc document the 101hz sampling frequency we expect by default?

src/docs/sdk/sample-format.mdx
```diff
@@ -0,0 +1,341 @@
...
+: *String, required.* Represents the version of the Sample format.
+
+Acceptable values are:
+- `"1"`
+
+### Profile data
+
+The profile data consists of a list of samples referring to stacks which in
+terms refers to frames as well as some metadata about threads. This is to keep the amount of data we transfer to the
+server as limited as possible.
```
vaind, line 235:
```suggestion
The profile data consists of 
* sample, which specifies the start time, their corresponding thread and a stack,
* stacks, which are a list of frames,
* (stack) frames, which identify the function and/or line of code for each item in a stacktrace,
* thread metadata, with info about threads.

The selected format is trying to keep the amount of data we transfer to the server as limited as possible.
```

src/docs/sdk/sample-format.mdx
```diff
@@ -0,0 +1,341 @@
...
+
+Stacks should be ordered from outermost to innermost frame.
+
+`frames`
+
+: *List, required*. Contains a list of frame objects (see [Frame
+Attributes](/sdk/event-payloads/stacktrace/#frame-attributes)).
+
+Technically, all attributes are optional so we won't reject the payload if some
+are missing.
```
vaind, line 302: Is that true, though? If there is absolutely no frame information, what is shown then?

vaind: Maybe just say the rules are the same as for standard event stack frames & symbolication?

vaind (commented):
Besides the specific comments below, I'm also missing the info that this is a sampling profiler, the `samples` section should contain start times for each frame currently running on a given thread (or at least that's my assumption)

PR #844, opened 2023-02-17
Split contexts into separate files and order them alphabetically.

antonpirker:
Make it easier to search for a context on that page and also making it easier to edit the contexts.


antonpirker:
Not needed anymore. Will close this

PR #848, opened 2023-02-23
chore(dev env): Add troubleshooting entry for devservices being down

lobsterkatie:
This adds a troubleshooting entry to the dev env setup docs, explaining that a connection error when running the dev server may be because devservices haven't yet been started.

Note: When DEVINFRA-30 is addressed, this will likely need to be modified or possibly even deleted.


asottile-sentry (approved):
I wonder if we can't auto fix this

lobsterkatie:
> I wonder if we can't auto fix this

That would be even better!

PR #856, opened 2023-03-02
Update UI Event Transaction the same event same ui element scenario

krystofwoldrich:
Based on the discussion [here](https://github.com/getsentry/sentry-javascript/pull/7236):
- https://github.com/getsentry/sentry-javascript/pull/7236

I think we should update the develop docs and open issues for SDKs that already implemented the current scenarios to update it.

### DIFF
--- a/src/docs/sdk/performance/ui-event-transactions.mdx
+++ b/src/docs/sdk/performance/ui-event-transactions.mdx
@@ -85,23 +85,23 @@ Scenario: Don't overwrite existing status of UI event transactions
     Then it keeps the status 
     And doesn't overwrite it
 
-Scenario: Same UI element with same event
-    Given an ongoing UI event transaction
-    When the user triggers the same UI event with the same type for the same
-        UI element
-    Then the SDK reschedule the idle timeout
-    And doesn't create a new transaction
-
 # If your SDK binds auto-generated transactions to the scope see binding
 # to scope
-Scenario: Same UI element with different event
-    Given an ongoing UI event transaction
-    When the user triggers the same UI element with a different event
+Scenario: New event on same UI element
+    Given an ongoing UI event transaction with child spans
+    When the user triggers the same UI element with a new event (different or same)
     Or the user triggers a different UI element
     Then the SDK finishes the ongoing transaction
     And sets the status to OK
     And waits for the children to finish
     And cancels the idle timeout
+    And starts a new transaction
+
+Scenario: New event on same UI element
+    Given an ongoing UI event transaction without child spans
+    When the user triggers the same UI element with a new event (different or same)
+    Or the user triggers a different UI element
+    Then the SDK discards the transaction 
     And starts a new transaction
 ```
 
--- a/src/docs/sdk/performance/ui-event-transactions.mdx
+++ b/src/docs/sdk/performance/ui-event-transactions.mdx
@@ -88,7 +88,7 @@ Scenario: Don't overwrite existing status of UI event transactions
 # If your SDK binds auto-generated transactions to the scope see binding
 # to scope
 Scenario: Event on same UI element with child spans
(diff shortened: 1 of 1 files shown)


0Calories (approved):
These scenarios look good to me, thanks @krystofwoldrich !

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -85,24 +85,24 @@ Scenario: Don't overwrite existing status of UI event transactions
...
-
 # If your SDK binds auto-generated transactions to the scope see binding
 # to scope
-Scenario: Same UI element with different event
-    Given an ongoing UI event transaction
-    When the user triggers the same UI element with a different event
+Scenario: New event on same UI element
+    Given an ongoing UI event transaction with child spans
+    When the user triggers the same UI element with a new event (different or same)
     Or the user triggers a different UI element
```
philipphofmann, line 93: `l`: There is already `Or the user triggers a different UI element` below so we don't need to mention it twice.
```suggestion
    When the user triggers the same UI element with a new event
    Or the user triggers a different UI element
```

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -85,24 +85,24 @@ Scenario: Don't overwrite existing status of UI event transactions
...
-        UI element
-    Then the SDK reschedule the idle timeout
-    And doesn't create a new transaction
-
 # If your SDK binds auto-generated transactions to the scope see binding
 # to scope
-Scenario: Same UI element with different event
-    Given an ongoing UI event transaction
-    When the user triggers the same UI element with a different event
+Scenario: New event on same UI element
```
philipphofmann, line 90: `m`: We have now two scenarios with the same name. I would rename this to 
```suggestion
Scenario: Event on same UI element with child spans
```
and rename the scenario below to
```
Scenario: Event on same UI element without child spans
```

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -85,24 +85,24 @@ Scenario: Don't overwrite existing status of UI event transactions
...
     Then the SDK finishes the ongoing transaction
     And sets the status to OK
     And waits for the children to finish
     And cancels the idle timeout
     And starts a new transaction
+
+Scenario: New event on same UI element
+    Given an ongoing UI event transaction without child spans
+    When the user triggers the same UI element with a new event (different or same)
+    Or the user triggers a different UI element
```
philipphofmann, line 103: `l`:
```suggestion
    When the user triggers the same UI element with a new event
    Or the user triggers a different UI element
```

philipphofmann (commented):
Thanks for updating the develop docs accordingly, @krystofwoldrich 👏 . A few issues to address.

krystofwoldrich:
@philipphofmann Thanks for the suggestions and [the meta issue](https://github.com/getsentry/team-mobile/issues/85), does it looks good now?

src/docs/sdk/performance/ui-event-transactions.mdx
```diff
@@ -85,24 +85,24 @@ Scenario: Don't overwrite existing status of UI event transactions
...
     And sets the status to OK
     And waits for the children to finish
     And cancels the idle timeout
     And starts a new transaction
+
+Scenario: Event on same UI element without child spans
+    Given an ongoing UI event transaction
+    When the user triggers the same UI element with a new event
+    Or the user triggers a different UI element
+    Then the SDK discards the transaction
```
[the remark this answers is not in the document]
krystofwoldrich: I agree this should be more explicit, so there can't be multiple ways of interpreting it. 
The way I understood it was not to sample/drop the transaction.

PR #880, opened 2023-03-21
Document Check-In envelope items

cleptric:


src/docs/sdk/check-ins.mdx
```diff
@@ -0,0 +1,49 @@
...
+`status`
+
+: **String, required**. The status of the check-in. Can be one of the following:
+  - `in_progress`: The check-in has started.
+  - `ok`: The check-in has completed successfully.
+  - `error`: The check-in has failed.
+
+`duration`
+
+: _Number, optional_. The duration of the check-in. Will only take effect if the
```
antonpirker, line 40:
```suggestion
: _Number, optional_. The duration of the check-in in milliseconds. Will only take effect if the
```

antonpirker: so even more important to add the resolution :-) 
Is it only ints or can it be a float?

PR #884, opened 2023-03-22
Define ignore_errors & ignore_transactions

smeubank:
### Background
With improved server side sampling Sentry attempts to already remove some transactions which do not provide value to the user. However this is a known case for many users for a long time. There is the traces sampler function, but the UX for this is bulky when you have a simple use case.

**Use case:** There transcations and errors which a users never wants sent to sentry, to appear in a feed or count against their quota.

**Proposal:** provide simple options which is a list of errors/exceptions and transactions, respectively, which the SDK will then always discard and never send to Sentry

**Preview:** [Event Sampling 
](https://develop-git-smeubank-define-ignore-options.sentry.dev/sdk/features/#event-sampling)
Similar implementations:
https://github.com/getsentry/sentry-php/pull/1503

JS `ignoreErrors`: [Manage Your Error Quota](https://docs.sentry.io/product/accounts/quotas/manage-event-stream-guide/#inboundfilters)

### DIFF
--- a/src/docs/sdk/features.mdx
+++ b/src/docs/sdk/features.mdx
@@ -55,7 +55,7 @@ if random() < sample_rate:
     transport.capture_event(event)
 ```
 
-To further simplfiy ignoring certain events from being sent to sentry, it is also suggested to provide `ignoreTransactions` and `ignoreErrors` (or exception, choose terminology which is best for the platform). This option should provide a simple way to allow users to discard events (ignore) from before they are sent to sentry. Prevents sending data which is is undesired and may consume quota or resources on the Sentry server.
+To further simplfiy ignoring certain events from being sent to sentry, it is also suggested to provide `ignoreTransactions` and `ignoreErrors` (or exception, choose terminology which is best for the platform). The array for `ignoreTransactions` specifically should contain an array of transaction names, as is stored in the transcation event schema (ie `GET /info`). These options should provide a simple way to allow users to discard events (ignore) from before they are sent to sentry. Prevents sending data which is is undesired and may consume quota or resources on the Sentry server.
 
 ```python
 ignore_transactions = ['GET /api/health','/api/v1/*']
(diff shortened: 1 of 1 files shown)


src/docs/sdk/features.mdx
```diff
@@ -55,6 +55,12 @@ if random() < sample_rate:
     transport.capture_event(event)
 ```
 
+To further simplfiy ignoring certain events from being sent to sentry, it is also sugggested to provide `ignoreTransactions` and `ignoreErrors` (or exception, choose terminology which is best for the platform). This option should provide a simple way to allow users to discard events (ignore) from before they are sent to sentry. Prevents sending data which is is undesired and may consume quota or resources on the Sentry server.
```
philipphofmann, line 58:
```suggestion
To further simplfiy ignoring certain events from being sent to sentry, it is also suggested to provide `ignoreTransactions` and `ignoreErrors` (or exception, choose terminology which is best for the platform). This option should provide a simple way to allow users to discard events (ignore) from before they are sent to sentry. Prevents sending data which is is undesired and may consume quota or resources on the Sentry server.
```

src/docs/sdk/features.mdx
```diff
@@ -55,6 +55,12 @@ if random() < sample_rate:
...
+
+```python
+ignore_transactions = ['GET /api/health','/api/v1/*']
```
philipphofmann, line 61: Please specify that this list refers to transaction names.

philipphofmann (commented):
Awesome feature. I added two small notes.

marandaneto:
@smeubank might be worth pointing out to https://develop.sentry.dev/sdk/client-reports/ as well since events are dropped on the client.
We don't have a generic item for that, maybe let's add new ones? should be straightforward.

smeubank:
> @smeubank might be worth pointing out to https://develop.sentry.dev/sdk/client-reports/ as well since events are dropped on the client. We don't have a generic item for that, maybe let's add new ones? should be straightforward.

will merge this as is. @jjbayer is there anything we need to consider on relay to send a new item type client reports for these 2 new discard reasons?

PR #887, opened 2023-03-23
Add trace origin

philipphofmann:
Add docs for trace origin being used for trace context and spans. Trace origin indicates what created a trace or a span.
Every SDK will benefit from trace origin. Therefore, I tagged also @getsentry/team-web-sdk-backend and @getsentry/team-web-sdk-frontend.

Related [RFC](https://github.com/getsentry/rfcs/pull/73)

[Visit docs preview](https://develop-git-feat-trace-and-span-origin.sentry.dev/sdk/performance/trace-origin/)

### DIFF
--- a/src/docs/sdk/performance/trace-and-span-origin.mdx
+++ b/src/docs/sdk/performance/trace-and-span-origin.mdx
@@ -7,7 +7,7 @@ to tell whether the user or what precisely in the SDK created it. Origin solves 
 the <Link to="sdk/event-payloads/contexts/#trace-context">trace context</Link> or <Link to="sdk/event-payloads/span/">spans</Link>.
 
 
-The origin consists of four parts: `<type>.<category>.<integration-name>.<integration-part>`
+The origin consists of four parts: `<type>.<category>.<integration-name>.<integration-part>`.
 Only the first is mandatory. The parts build upon each other, meaning it is forbidden
 to skip one part. For example, you may send parts one and two but aren't allowed to send parts one and three without part two.
 
--- a/src/docs/sdk/performance/trace-and-span-origin.mdx
+++ b/src/docs/sdk/performance/trace-and-span-origin.mdx
@@ -55,6 +55,6 @@ Examples:
 
 ## See also:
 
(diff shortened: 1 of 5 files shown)


src/docs/sdk/performance/trace-and-span-origin.mdx
```diff
@@ -0,0 +1,60 @@
...
+- `auto.http.okhttp`
+- `auto.http.http_client_5`
+- `auto.http.client.stream`
+- `auto.db.core_data`
+- `auto.db.graphql`
+
+## See also:
+
+- <Link to="sdk/event-payloads/span/">Span Interface</Link>
+- <Link to="sdk/event-payloads/contexts/#trace-context">Trace Context</Link>
```
marandaneto, line 59:
```suggestion
- <Link to="/sdk/event-payloads/span/">Span Interface</Link>
- <Link to="/sdk/event-payloads/contexts/#trace-context">Trace Context</Link>
```

marandaneto:
We were not sure about the naming tho, `origin` is already used before, maybe we just double check this?

marandaneto:
> > We were not sure about the naming tho, `origin` is already used before, maybe we just double check this?
> 
> @marandaneto, I didn't find any usages of `origin` in the [develop docs](https://github.com/search?q=repo%3Agetsentry%2Fdevelop+origin+language%3AMarkdown&type=code&l=Markdown) or in [Relay](https://getsentry.github.io/relay/relay/?search=origin). Where do you think it is already used?
> 
> There is an [HTTP header origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin) whose concept is slightly similar to our `origin` in this PR.
> 
> > The Origin request header indicates the [origin](https://developer.mozilla.org/en-US/docs/Glossary/Origin) (scheme, hostname, and port) that caused the request.

You're right, I've mistaken with the `event.origin` tag, See https://docs.sentry.io/platforms/react-native/usage/event-information/
All good.

marandaneto (approved):

krystofwoldrich (approved):

PR #897, opened 2023-04-05
ref(integrations): Update instructions for installing Jira integration

lobsterkatie:
This clarifies and expands upon the instructions for installing our Jira integration in the following ways:
- Explain that the Atlassian cloud instance is development-specific.
- Add section about installing ngrok, for folks developing locally.
- Clarify that the domain to use for the descriptor is the sentry domain, not the jira domain.

Note that the ngrok section is incomplete, insofar as it doesn't mention any of the changes you need to make to `sentry.conf.py`, but really that should probably go on its own page, which will have to be a separate PR. For the moment, the info is available in Notion.

Ref: WOR-2912

### DIFF
--- a/src/docs/integrations/jira/index.mdx
+++ b/src/docs/integrations/jira/index.mdx
@@ -1,17 +1,26 @@
 ---
-title: "Jira Integration"
+title: 'Jira Integration'
 ---
 
-Create an [Atlassian Cloud Instance](http://go.atlassian.com/cloud-dev)
+Create an [Atlassian Cloud Developer Instance](http://go.atlassian.com/cloud-dev)
 
-Configure Jira settings.
-  1. Navigate to Jira settings (cog icon) > Apps > Manage apps.
-  2. Scroll to the bottom of the Manage apps page, and click Settings (1).
-  3. Select Enable development mode and Enable private listings (2), and click Apply.
+Atlassian offers a development-specific org option, different from simply having a personal Jira organization. Even if you already have a personal Jira org, it's worth creating a development org, because it will correctly set a number of necessary options for you automatically.
+
+Set Up ngrok
+
+(This is only necessary if you're developing locally.) [Download](https://ngrok.com/download) and [set up](https://ngrok.com/docs/getting-started/) ngrok, then start an ngrok server.
+
+Configure Jira Settings
+
+1. If you don't have any projects in your Jira org yet, create a new project.
+2. In the navigation bar at the top of the screen, go to Apps > Manage Your Apps.
+3. At the bottom of the Manage Apps page, click Settings (1).
+4. Select Enable Development Mode and Enable Private Listings (2), and click Apply.
 
 ![Jira Settings](./configure-jira.png)
 
-Upload the Atlassian Connect descriptor to Jira. It can be found at `{YOUR_DOMAIN}/extensions/jira/descriptor/`
+5. Reload the page, and Upload App should appear next to Build a New App.
+6. Click Upload App and upload the Atlassian Connect descriptor by using the URL  `https://{YOUR_SENTRY_DOMAIN}/extensions/jira/descriptor/`. Note that if you are running a local devserver, `YOUR_SENTRY_DOMAIN` will be your ngrok (or other tunneling service) domain.
 
 ![Upload Descriptor to Jira](./upload-descriptor.png)
 


src/docs/integrations/jira/index.mdx
```diff
@@ -1,18 +1,27 @@
...
 ---
 
-Create an [Atlassian Cloud Instance](http://go.atlassian.com/cloud-dev)
+Create an [Atlassian Cloud Developer Instance](http://go.atlassian.com/cloud-dev)
 
-Configure Jira settings.
-  1. Navigate to Jira settings (cog icon) > Apps > Manage apps.
-  2. Scroll to the bottom of the Manage apps page, and click Settings (1).
-  3. Select Enable development mode and Enable private listings (2), and click Apply.
+Atlassian offers a development-specific org option, different from simply having a personal Jira organization. Even if you already have a personal Jira org, it's worth creating a development org, because it will correctly set a number of necessary options for you automatically.
```
ceorourke, line 7: What are the options? I use a regular personal Jira organization.

src/docs/integrations/jira/index.mdx
```diff
@@ -1,18 +1,27 @@
...
+
+Set Up ngrok
+
+(This is only necessary if you're developing locally.) [Download](https://ngrok.com/download) and [set up](https://ngrok.com/docs/getting-started/) ngrok, then start an ngrok server.
```
ceorourke, line 11: I'm okay with leaving this in, but note that every integration that has instructions in the develop docs needs this to develop locally. It might be better served as it's own page as you mentioned, just feels a little strange to have it on one page and not the others.

src/docs/integrations/jira/index.mdx
```diff
@@ -1,18 +1,27 @@
...
+
+Configure Jira Settings
+
+1. Create a new project in Jira.
```
ceorourke, line 15: This is only necessary if you are using a brand new Jira organization, right? If you've already been using Jira and have a project/projects you can skip this step.

src/docs/integrations/jira/index.mdx
```diff
@@ -1,18 +1,27 @@
...
+2. In the navigation bar at the top of the screen, go to Apps > Manage Your Apps.
+3. At the bottom of the Manage Apps page, click Settings (1).
+4. Select Enable Development Mode and Enable Private Listings (2), and click Apply.
 
 ![Jira Settings](./configure-jira.png)
 
-Upload the Atlassian Connect descriptor to Jira. It can be found at `{YOUR_DOMAIN}/extensions/jira/descriptor/`
+5. Reload the page, and Upload App should appear next to Build a New App.
+6. Click Upload App and upload the Atlassian Connect descriptor by using the URL  `https://{YOUR_SENTRY_DOMAIN}/extensions/jira/descriptor/`. Note that if you are running a local devserver, `YOUR_SENTRY_DOMAIN` will be your ngrok domain.
```
ceorourke, line 23: It's possible to use other services / methods besides ngrok here which is why we've kept the language agnostic

src/docs/integrations/jira/index.mdx
lobsterkatie: Yeah, once I move ngrok stuff from Notion to a real dev docs page, this (and any other integration) can just have a one-liner, along the lines of "if you're developing locally, start up ngrok" and link to the setup page.

src/docs/integrations/jira/index.mdx
lobsterkatie: True. But we're also telling people to make a new cloud instance when they might already have one. That said, I'm fine to change this to, "If you don't have any projects in your Jira org yet, create a new project."

src/docs/integrations/jira/index.mdx
lobsterkatie: I can add "(or equivalent)" if that helps. More or less I'm trying to write down all of the things I had to figure out/realize so that they're just obvious for the next person.

src/docs/integrations/jira/index.mdx
lobsterkatie: A Jira org existing, users being in projects, and users being "assignable" (which also makes them eligible to be the reporter, which is a required field in our Create Issue modal). 

It's partially my fault, because - thinking I already had a personal Jira - I didn't click the Atlassian Cloud link. But trying to log into Jira with my personal email took me instead to BitBucket (because it turns out I didn't have a personal Jira account), and so first I wandered around the Atlassian admin figuring out how to add Jira. Then I spent more time than I care to admit Friday afternoon tearing my hair out, trying to figure out why on earth I wasn't showing up in the options for the Reporter field even though it was my personal org, I'm the one who created the linked project, and the basic member search API would return my record just fine. Turns out I had to both add myself to the project and go into project settings and change the permissions for "assignable users," but it took a lottttt of trial and error and print debugging and hitting various API endpoints before I finally figured that out.

So then Monday morning, I popped open this repo to add a note about the necessary steps, and noticed that the link I'd ignored on Friday said "cloud-dev" at the end, which clued me into the fact that this was probably different than just a personal Jira. So I started again, this time making a dev org (totally separate from my personal one), and a) Jira was already there and didn't need to be turned on, b) when I created a project I was automatically added to it, and c) the permissions by default were set correctly such that I was assignable and therefore showed up in the Reporter field in our modal. So instead of the literal hours it took me on Friday, I was instead able to set up the integration in under five minutes - everything "just worked" - and so I figured that's why we chose that option instead of just telling people "go sign up for Jira."

Given our conversation on Slack, though, I'm taking it you didn't need to go through all of that rigamarole when you used your personal Jira?

ceorourke: 😅 I don't remember a similar rigamarole, though I did set it up quite some time ago. I hadn't actually heard of the dev specific org option which is why I was curious. It may have not existed when we (I? did I write these?) wrote these docs.

src/docs/integrations/jira/index.mdx
lobsterkatie: Okay, I updated it to, "Note that if you are running a local devserver, `YOUR_SENTRY_DOMAIN` will be your ngrok (or other tunneling service) domain."

ceorourke (approved):

PR #903, opened 2023-04-10
ref(feature flags): Fix broken Flagr images and hover text

ceorourke:
Fix the image path for all the images on the Flagr docs, plus put hover text where it is supposed to be.


wmak (approved):
Huh when did these break? I know they were working before 🤔

PR #921, opened 2023-04-25
Send DSC on all envelopes

bruno-garcia:
Relates to: https://github.com/getsentry/team-webplatform-meta/issues/45

### DIFF
--- a/src/docs/sdk/performance/dynamic-sampling-context.mdx
+++ b/src/docs/sdk/performance/dynamic-sampling-context.mdx
@@ -117,7 +117,7 @@ After the DSC of a particular trace has been frozen, API calls like `set_user` s
 
 Dynamic Sampling Context is sent to Sentry via the `trace` envelope header and is propagated to downstream SDKs via a baggage header.
 
-All of the values in the payload schemas below are required (non-optional) in a sense, that when they are known to an SDK at the time a transaction envelope is sent to Sentry, or at the time a baggage header is propagated, they must also be included in said envelope or baggage.
+All of the values in the payload schemas below are required (non-optional) in a sense, that when they are known to an SDK at the time an envelope with an event (transaction or error) is sent to Sentry, or at the time a baggage header is propagated, they must also be included in said envelope or baggage.
 
 <Alert level="warning">
 


mattjohnsonpint:
Cool.

Also, is dsc really still "experimental"? 🤔

mattjohnsonpint (approved):

mattjohnsonpint:
Also:

- We don't (yet) have `trace_id` for errors, but it is still listed under ["Strictly Required Values"](https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/#strictly-required-values)

- `sample_rate` is also listed as strictly required, but also in the [envelope header section](https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/#envelope-header) it is described as being the traces sample rate.  For errors, should it be the _errors_ sample rate? That's usually 100%, but could possibly be set lower.  Does the back-end handle them interchangeably?  Or perhaps for errors, this should be omitted?

mattjohnsonpint:
Also, https://github.com/getsentry/sentry-javascript/pull/7820 doesn't actually put DSC on _all_ error events, only those error events that have an active transaction.  Is that the intent for all SDKs, or is that wrong?

mattjohnsonpint (changes requested):
Let's hold on this until we get clarification on the previously mentioned points.  I'll ask at TSC.  Thanks.

AbhiPrasad:
https://github.com/getsentry/sentry-javascript/pull/7820 was intentional - we don't have tracing without performance support in the JS SDK yet, so we can only extract DSC from from an active transaction (right now the transaction is the carrier of the DSC in all the SDKs)

When we add tracing without performance support, we'll start adding DSC to every single envelope, regardless of active transaction.

bruno-garcia:
> We don't (yet) have trace_id for errors, but it is still listed under ["Strictly Required Values"](https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/#strictly-required-values)

Then don't send it. This is only when a transaction is happening (performance is a requirement)

> sample_rate is also listed as strictly required, but also in the [envelope header section](https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/#envelope-header) it is described as being the traces sample rate. For errors, should it be the errors sample rate? That's usually 100%, but could possibly be set lower. Does the back-end handle them interchangeably? Or perhaps for errors, this should be omitted?

Send the transaction sample rate, nothing changes.

mattjohnsonpint:
Discussed at TSC and agreed.  See https://github.com/getsentry/team-webplatform-meta/issues/45#issuecomment-1523815710

So for now, this just needs to be updated to say that a transaction must be available.

bruno-garcia:
> So for now, this just needs to be updated to say that a transaction must be available.

Right now a transaction must be available in the current implementation. But the goal is for that requirement not to exist. So I'll avoid mentioning this requirement here and the need to remove it later

PR #924, opened 2023-04-27
feat: Add conventions for http.method and db.operation

AbhiPrasad:
- Add `http.method`
- Add `db.operation`
- Re-organize sections
- Explicitly add deprecated conventions (need to list since SDKs still send).

### DIFF
--- a/src/docs/sdk/performance/span-data-conventions.mdx
+++ b/src/docs/sdk/performance/span-data-conventions.mdx
@@ -12,11 +12,14 @@ Below describes the conventions for the Span interface for the `data` field on t
 
 ## HTTP
 
-| Attribute       | Type   | Description                          | Examples           |
-| --------------- | ------ | ------------------------------------ | ------------------ |
-| `http.query`    | string | The Query string present in the URL. | `?foo=bar&bar=baz` |
-| `http.fragment` | string | The Fragments present in the URL.    | `#foo=bar`         |
-| `http.method`   | string | The HTTP method used.                | `GET`              |
+| Attribute                          | Type   | Description                                                          | Examples           |
+| ---------------------------------- | ------ | -------------------------------------------------------------------- | ------------------ |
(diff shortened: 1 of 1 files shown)


src/docs/sdk/performance/span-data-conventions.mdx
```diff
@@ -12,10 +12,14 @@ Below describes the conventions for the Span interface for the `data` field on t
...
-| --------------- | ------ | ------------------------------------ | ------------------ |
-| `http.query`    | string | The Query string present in the URL. | `?foo=bar&bar=baz` |
-| `http.fragment` | string | The Fragments present in the URL.    | `#foo=bar`         |
+| Attribute                          | Type   | Description                           | Examples           |
+| ---------------------------------- | ------ | ------------------------------------- | ------------------ |
+| `http.query`                       | string | The Query string present in the URL.  | `?foo=bar&bar=baz` |
+| `http.fragment`                    | string | The Fragments present in the URL.     | `#foo=bar`         |
+| `http.method`                      | string | The HTTP method used.                 | `GET`              |
+| `http.request_content_length`      | number | The encoded body size of the request. | `123`              |
+| `http.decoded_request_body_length` | number | The decoded body size of the request. | `456`              |
```
narsaynorath, line 21: Ah, I realized there's a mistake here that I merged in my last PR. I think we should be using `response` and not `request`

AbhiPrasad: you're right! I can update.

AbhiPrasad:
@narsaynorath I updated, and I also switched it from `http.transfer_size` to `http.response_transfer_size` to be a bit more exact. Mind updating the SDK/Sentry PRs for this as well?

AbhiPrasad:
Merging since we're matching otel for the added attributes, and the rest is cleanup.

PR #936, opened 2023-05-17
Add SDK field to check-ins

AbhiPrasad:
Aligns docs with Relay.

See https://github.com/getsentry/relay/blob/dd93441bbb607dba806c6f1ca0c03783e348f048/relay-monitors/src/lib.rs#L129


HazAT:
We discussed it in Slack - we are closing this since this info is already on the envelope header

gaprl:
FYI @AbhiPrasad we'll remove the code from Relay that extracts this info from the payload to make sure we're consistent with the envelope header.

PR #940, opened 2023-05-22
Update span-data-conventions for HTTP.status_code

smeubank:
Per https://github.com/getsentry/team-webplatform-meta/issues/65 we want to define where HTTP status codes should be stored in events (200, 404, 500, etc)

Reasoning for why it is a problem, is that SDKs generally should not set tags "for" users on events.


src/docs/sdk/performance/span-data-conventions.mdx
```diff
@@ -17,6 +17,7 @@ Below describes the conventions for the Span interface for the `data` field on t
 | `http.query`                        | string | The Query string present in the URL.   | `?foo=bar&bar=baz` |
 | `http.fragment`                     | string | The Fragments present in the URL.      | `#foo=bar`         |
 | `http.method`                       | string | The HTTP method used.                  | `GET`              |
+| `http.status_code`                  | string | The status HTTP resposne.              | `404`              |
```
marandaneto, line 20:
```suggestion
| `http.status_code`                  | string | The status HTTP response.              | `404`              |
```

marandaneto (approved):

PR #944, opened 2023-05-24
SDK: Don’t forget the big picture

HazAT:
Let's never forget we are building a product not just a library

### DIFF
--- a/src/docs/sdk/philosophy.mdx
+++ b/src/docs/sdk/philosophy.mdx
@@ -53,3 +53,7 @@ Some types of errors cannot be resolved without the data that was given to the p
 But make sure that auto instrumentations doesn't attach PII without an explicit opt-in from the user.
 The server must be aware of parts of the protocol that include PII to scrub them by default.
 Please check <Link to="/sdk/data-handling">Data Handling</Link> for more detail.
+
+## Don’t forget the big picture
+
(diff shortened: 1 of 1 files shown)


marandaneto (approved):

src/docs/sdk/philosophy.mdx
```diff
@@ -52,4 +52,8 @@ While we generally should try to keep the API surfaces of SDKs reasonable small.
 Some types of errors cannot be resolved without the data that was given to the program. 
 But make sure that auto instrumentations doesn't attach PII without an explicit opt-in from the user.
 The server must be aware of parts of the protocol that include PII to scrub them by default.
-Please check <Link to="/sdk/data-handling">Data Handling</Link> for more detail.
\ No newline at end of file
+Please check <Link to="/sdk/data-handling">Data Handling</Link> for more detail.
+
+## Don’t forget the big picture
+
+Being correct is important but even more important is doing the right thing. Yes, we are building libraries and with that comes much responsibility. But in the end, we build a product that should solve a problem users have in their application. This principle relates to many other things mentioned on this page - but most importantly, it should act as a reminder that we only succeed if we manage to solve a problem end-to-end.
```
philipphofmann, line 59: Tiny nit picks:

```suggestion
Being correct is important, but doing the right thing is even more essential. Yes, we are building libraries, and with that comes much responsibility. But in the end, we build a product that should solve a problem users have in their application. This principle relates to many other things mentioned on this page - but most importantly, it should act as a reminder that we only succeed if we manage to solve a problem end-to-end.
```

philipphofmann (approved):
Thank you, @HazAT. Just merge it 🚀 ✔

PR #950, opened 2023-05-30
feat(crons): Update SDK docs for empty check-ins

evanpurkhiser:
See https://github.com/getsentry/sentry/issues/49664

### DIFF
--- a/src/docs/sdk/check-ins.mdx
+++ b/src/docs/sdk/check-ins.mdx
@@ -24,6 +24,11 @@ The following fields exist:
 
 : **String, required**. Check-In ID (unique and client generated).
 
+  This may be provided as a empty UUID (128 bit zero value) to indicate to
+  Sentry that the checkin should update the previous check-in that is not in
+  the terminal state. If the most recent check-in is already in a terminal
+  state a new one will be created instead.
+
 `monitor_slug`
 
 : **String, required**. The distinct slug of the monitor.


src/docs/sdk/check-ins.mdx
```diff
@@ -24,6 +24,11 @@ The following fields exist:
 
 : **String, required**. Check-In ID (unique and client generated).
 
+  This may be provided as a empty UUID (128 bit zero value) to indicate to
```
cleptric, line 27: Do you expect the SDKs to implement this feature?

evanpurkhiser: At the moment no I would expect almost all SDKs to maintain their checkin IDs for each run. This is to support the simplified CURL style API as outlined in https://github.com/getsentry/sentry/issues/49664

cleptric: The linked PR mentions `null`, do you suggest `00000...`?

src/docs/sdk/check-ins.mdx
```diff
@@ -24,6 +24,11 @@ The following fields exist:
...
+  Sentry that the checkin should update the previous check-in that is not in
+  the terminal state. If the most recent check-in is already in a terminal
+  state a new one will be created instead.
```
cleptric, line 30: It might be worth clarifying what terminal state means.

src/docs/sdk/check-ins.mdx
evanpurkhiser: I think @mitsuhiko's description is slightly misleading, it is not nullable, but it passes a empty value instead. You can see it in the tests here https://github.com/getsentry/relay/pull/2153/files#diff-2b7b3259fa82d6066fa1268aed58cf7a8bac4e9317e275b7601cc916b1201e4bR56

src/docs/sdk/check-ins.mdx
rjo100: yeah reverse the logic so it should "update the most recent check-in in an `in progress` state. If the most recent check-in is not `in progress`..."

PR #952, opened 2023-06-01
feat(dynamic-sampling): Add docs for new sliding window per org

iambriccardo:
This PR adds the new sliding window concepts to the dynamic sampling docs.


ale-cota (approved):
LGTM 😄

PR #953, opened 2023-06-01
fix(handbook): add type decl section

JonasBa:


src/docs/frontend/index.mdx
```diff
@@ -68,23 +68,28 @@ Both the `app/components/` and `app/views` folders contain React components.
...
-})
+interface B {
+  color: "blue";
+}
+// Interface 'C' incorrectly extends interface 'B'.
+//   Types of property 'color' are incompatible.
+interface C extends B {
+  color: "red";
+}
 ```
```
ryan953, line 91:
```suggestion
```typescript
// This type intersection has conflicts, `A` resolves to be the type `never`:
type A = { color: "blue" } & { color: "red" };

// By extending interface `B` TypeScript will detects conflicts:
interface B {
  color: "blue";
}

// TS2430: Interface 'C' incorrectly extends interface 'B'.
//   Types of property 'color' are incompatible.
interface C extends B {
  color: "red";
}
```
```

File: src/docs/frontend/index.mdx [hunk 2]
```diff
@@ -68,23 +68,28 @@ Both the `app/components/` and `app/views` folders contain React components.
 - Use a view for UI that will typically not be reused in other parts of the codebase
 - Use a component for UI that is designed to be highly reusable.
 
-### PropTypes
+### Type declarations
 
-Use them, be explicit, use the shared custom proptypes when possible.
+We do not follow any strict "prefer type over interface" rules for type declarations, so using either is at your discretion. That being said, there are [functional differences](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces) between the two that you should be aware of so you can use them appropriately. 
+A simple guiding principle is to prefer interfaces when you require declaration merging. The reason for that is that type intersections are poor at detecting conflicts, produce unexpected or unwanted results and when used in combination with large types [perform worse than interfaces](https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections).
```
ryan953, line 74:
```suggestion

A simple guiding principle is to prefer interfaces when you require declaration merging. The reason for that is that type intersections are poor at detecting conflicts, produce unexpected or unwanted results and when used in combination with large types [perform worse than interfaces](https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections).
```

src/docs/frontend/index.mdx
```diff
@@ -68,23 +68,28 @@ Both the `app/components/` and `app/views` folders contain React components.
 - Use a view for UI that will typically not be reused in other parts of the codebase
 - Use a component for UI that is designed to be highly reusable.
 
-### PropTypes
+### Type declarations
 
-Use them, be explicit, use the shared custom proptypes when possible.
+We do not follow any strict "prefer type over interface" rules for type declarations, so using either is at your discretion. That being said, there are [functional differences](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces) between the two that you should be aware of so you can use them appropriately.
```
ryan953, line 73:
```suggestion
We do not follow any strict "prefer interface over type" rules for type declarations, so using either is at your discretion. That being said, there are [functional differences](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces) between the two that you should be aware of so you can use them appropriately. 
```

ryan953 (approved):

src/docs/frontend/index.mdx
JonasBa: I would keep this as is now. Looking over the repo, it seems that prefer type was the unofficial guideline so I want to kind of debunk that with this line :)

src/docs/frontend/index.mdx
```diff
@@ -68,23 +68,29 @@ Both the `app/components/` and `app/views` folders contain React components.
...
 - Use a component for UI that is designed to be highly reusable.
 
-### PropTypes
+### Type declarations
 
-Use them, be explicit, use the shared custom proptypes when possible.
+We do not follow any strict "prefer type over interface" rules for type declarations, so using either is at your discretion. That being said, there are [functional differences](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces) between the two that you should be aware of so you can use them appropriately. 
 
-Prefer `Proptypes.arrayOf` to `PropTypes.array` and `PropTypes.shape` to `PropTypes.object`
+A simple guiding principle is to prefer interfaces when you require declaration merging. The reason for that is that type intersections are poor at detecting conflicts, produce unexpected or unwanted results and when used in combination with large types [perform worse than interfaces](https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections).
```
priscilawebdev, line 75: love this one! it is a good explanation ⭐️

src/docs/frontend/index.mdx
priscilawebdev: we should point out that these are examples of things we should not be doing. Maybe we can use a symbol or something ⛔️

priscilawebdev: can we also have examples of good types?

JonasBa: I cant really think of examples of "good" types per say, as long as you don't need declaration merging, using a type is fine.

ryan953 (approved):

PR #969, opened 2023-06-14
docs(api): Update api instructions

schew2381:
Update `Making an API Public` section with more recent `@extend_schema` example, new `OpenApiExample` that includes `request_only=True` addition, and warning to update all requests in one endpoint in the same PR if multiple requests in the endpoint previously used manual JSON build.


JoshFerge (approved):
great stuff!

PR #992, opened 2023-07-04
updated ttid/ttfd docs for app start timestamp

stefanosiano:
updated ttid/ttfd specs to explicitly align them to the app start timestamp

### DIFF
--- a/src/docs/sdk/performance/time-to-initial-full-display.mdx
+++ b/src/docs/sdk/performance/time-to-initial-full-display.mdx
@@ -11,6 +11,7 @@ which can be used to show more accurate information on the time spent in the scr
 Being spans, they don't impact user quota.
 
 The SDK starts them as soon as a new screen is created, as long as there is an active transaction for the screen load.
+Note: When the app is started, the SDK aligns the start timestamp of the Time to Initial Display and Time to Full Display to the app start timestamp.
 
 ## Time to Initial Display
 
@@ -46,6 +47,12 @@ Scenario: Option enabled
     Given the enableTimeToFullDisplay option is enabled
     When the SDK creates a new screen load transaction
     Then the Full Display span is started
+
+Scenario: App started
+    Given the enableTimeToFullDisplay option is enabled
+    When the app starts
+    And the SDK start the span
+    Then the SDK aligns the span start timestamp to the app start timestamp
 
 Scenario: Screen displayed
     Given an ongoing Full Display span


src/docs/sdk/performance/time-to-initial-full-display.mdx
```diff
@@ -47,6 +48,12 @@ Scenario: Option enabled
     When the SDK creates a new screen load transaction
     Then the Full Display span is started
 
+Scenario: App started
+    Given the enableTimeToFullDisplay option is enabled
+    When the app starts
+    And the SDK start the span
+    Then the SDK aligns the span start timestamp to the app start timestamp
```
philipphofmann, line 55: I think that's a bit cleaner
```suggestion
    The SDK sets the start timestamp of the Time to Initial Display and Time to Full Display span to the app start timestamp.
```

src/docs/sdk/performance/time-to-initial-full-display.mdx
```diff
@@ -47,6 +48,12 @@ Scenario: Option enabled
     When the SDK creates a new screen load transaction
     Then the Full Display span is started
 
+Scenario: App started
+    Given the enableTimeToFullDisplay option is enabled
+    When the app starts
+    And the SDK start the span
```
philipphofmann, line 54: Should the SDK also change the recorded measurements? If so, please add it to this scenario.

philipphofmann (commented):
Some minor comments.

src/docs/sdk/performance/time-to-initial-full-display.mdx
stefanosiano: The measurement is always the span duration. If the timestamp is changed, the span duration changes, so the SDK shouldn't do anything special regarding it.
At least this is how i see it. Would it make more sense to explicitly list it in the scenarios? (including all the other scenarios when the SDK modifies the start or end timestamp of the span)

philipphofmann: Yes, I would explicitly list that in this scenario.

philipphofmann (approved):
LGTM, just an update on the measurements, please. I'm approving to unblock this.

src/docs/sdk/performance/time-to-initial-full-display.mdx
stefanosiano: @philipphofmann is the wording fine?

src/docs/sdk/performance/time-to-initial-full-display.mdx
```diff
@@ -51,7 +51,8 @@ Scenario: Option enabled
 Scenario: App started
     Given the enableTimeToFullDisplay option is enabled
     When the app starts
-    The SDK sets the start timestamp of the Time to Initial Display and Time to Full Display span to the app start timestamp.
+    Then the SDK sets the start timestamp of the Time to Initial Display and Time to Full Display span to the app start timestamp
+    And the SDK will update the `time_to_initial_display` and `time_to_full_display` measurements accordingly
```
philipphofmann, line 55: `l`: we use the present tense above, so we can also use it here to align.
```suggestion
    And the SDK updates the `time_to_initial_display` and `time_to_full_display` measurements accordingly
```

philipphofmann (approved):
LGTM

PR #1015, opened 2023-08-07
fix: add line breaks in self-hosted overview

RafDevX:
Without line breaks, the following paragraph is merged into the last list item.


AbhiPrasad (approved):
Thanks!

PR #1025, opened 2023-08-17
fix: 1MB limit is after decompression

jredmond-sentry:
This clarifies the current payload size limit for the `/api/PROJECT_ID/store/` endpoint, and makes it more like how the limit is described at https://develop.sentry.dev/sdk/store/#size-limits.


antonpirker (approved):
lgtm

PR #1032, opened 2023-09-05
docs(api): Add notes on designing new APIs

sentaur-athena:
Adding more notes on writing a new API

### DIFF
--- a/src/docs/api/design.mdx
+++ b/src/docs/api/design.mdx
@@ -2,7 +2,7 @@
 title: "Designing a New API"
 sidebar_order: 2
 ---
-[Django REST framework](https://www.django-rest-framework.org/) is a powerful and flexible toolkit
+[Django REST framework](https://www.django-rest-framework.org/)(DRF) is a powerful and flexible toolkit
 for building Web APIs. Sentry APIs are built using DRF. Here are some considerations to make when designing APIs at Sentry. 
 
 ## URL Patterns
@@ -31,20 +31,26 @@ Here are some additional examples:
 - 👍 `/projects/{organization_slug}/{project_slug}/issues/`
 - 👎 `/projects/`
 
-**Note:** Hierarchy here does not necessarily mean that one collection belongs to a parent collection. For example `projects/{project_identifier}/teams/` and `teams/{team_identifier}/projects/` can co-exist. The first one means the teams that are working on a specific project and the latter means the projects one specific team is working on.
+<Alert title="Note" level="info">
+
+Hierarchy here does not necessarily mean that one collection belongs to a parent collection. For example:
+- `projects/{project_identifier}/teams/` refers to the **teams** that have been added to specific project
+- `teams/{team_identifier}/projects/` refers to the **projects** a specific team has been added to
+
+</Alert>
 
 ### Naming Guidance
-- Collection name should be hyphen-ated lower case, e.g. commit-files.
-- Collection names must be plural. Avoid using uncountable words because user can’t know if the GET returns one item or a list.
-- Query params and body params are CamelCased. 
-- For sorting and filtering try to stick with the common param names: sortBy (e.g. date_created), orderBy (either asc or desc) groupBy, limit 
-- Path params are snake_cased
+- Collection names should be lowercase and hyphenated, e.g. `commit-files`.
+- Collection names must be plural. Avoid using uncountable words because the user can’t know whether the GET returns one item or a list.
+- Query params and body params should be `CamelCase`.
+- For sorting and filtering, stick with the common param names: sortBy (e.g. date_created), orderBy (either asc or desc), groupBy, limit
+- Path params should be `snake_case`
 
 ## Functionality
-Each API should be **stateless**, have a clear purpose and do one clear thing. To achieve that try to stick with the standard methods below. If you have a need to a more complicated functionality reach out to [owners-api](https://github.com/orgs/getsentry/teams/owners-api) and we can help you achieve it.
+Each API should be **stateless**, have a clear purpose, and do one specific thing. To achieve that, stick with the standard methods listed below. If your API needs to be more complicated, work with [owners-api](https://github.com/orgs/getsentry/teams/owners-api) on how to create it.
 
-- 👍 An API that updates project settings: PATCH or PUT
-- 👎 An API that creates a project and creates a team and creates alerts for that team about that project
+- 👍 An API that updates project settings: PATCH for updating a field or PUT for updating settings
+- 👎 An API that creates a project, creates a team, and creates alerts for that team about that project
 
 | Functionality | HTTP Method | Response Object | Example |
 | --- | ---- | ---- | ---- |
@@ -58,7 +64,7 @@ Each API should be **stateless**, have a clear purpose and do one clear thing. T
 | Batch Update | PUT | List of serialized updated resources | [Update a list of issues](https://github.com/getsentry/sentry/blob/ea14f740c78b8df68281ffad6a3bf3709ed3d5b5/src/sentry/api/endpoints/organization_group_index.py#L379) |
 | Batch Delete | DELETE | None | [Delete a list of issues](https://github.com/getsentry/sentry/blob/ea14f740c78b8df68281ffad6a3bf3709ed3d5b5/src/sentry/api/endpoints/organization_group_index.py#L467)|
 
-Here are some examples where you might think standard methods are not enough but they actually are:
+Here are some examples of how to use standard methods to represent complex tasks:
 - Get count of a resource
     - Count is part of the `List` API and is provided in header X-Total-Count param
 - Get latest of a resource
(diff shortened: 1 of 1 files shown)


src/docs/api/design.mdx
```diff
@@ -24,3 +24,54 @@ In the above example we flattened `projects`. The table below shows the existing
 | projects | When the resource is under a specific project in the hierarchy but not under an issue | organizations | {organization_slug}/ {project_slug} | [Create a New Client Key](https://docs.sentry.io/api/projects/create-a-new-client-key/)|
 | issues | When the resource is under a specific issue in the hierarchy | projects | {issue_id} | [List an Issue's Events](https://docs.sentry.io/api/events/list-an-issues-events/)|
 | sentry-app-installations | When the resource is mapped to a specific integration | organizations | {integration_slug} | [Delete an External Issue](https://docs.sentry.io/api/integration/delete-an-external-issue/)|
+
+Here are some additional examples:
+- 👍 `/organizations/{organization_slug}/projects/`
+- 👍 `/projects/{organization_slug}/{project_slug}/issues/`
+- 👎 `/projects/`
+
+**Note:** Hierarchy here does not necessarily mean that one collection belongs to a parent collection. For example `projects/{project_identifier}/teams/` and `teams/{team_identifier}/projects/` can co-exist. The first one means the teams that are working on a specific project and the latter means the projects one specific team is working on.
```
sentaur-athena, line 33: Is there a nicer way to show a note?

src/docs/api/design.mdx
```diff
@@ -2,7 +2,8 @@
 title: "Designing a New API"
 sidebar_order: 2
 ---
-Here are some considerations to make when designing new Sentry APIs. 
+[Django REST framework](https://www.django-rest-framework.org/) is a powerful and flexible toolkit
```
EricHasegawa, line 5: Should probably call out the acronym like (DRF) directly here

src/docs/api/design.mdx
```diff
@@ -24,3 +25,54 @@ In the above example we flattened `projects`. The table below shows the existing
...
+- 👍 `/organizations/{organization_slug}/projects/`
+- 👍 `/projects/{organization_slug}/{project_slug}/issues/`
+- 👎 `/projects/`
+
+**Note:** Hierarchy here does not necessarily mean that one collection belongs to a parent collection. For example `projects/{project_identifier}/teams/` and `teams/{team_identifier}/projects/` can co-exist. The first one means the teams that are working on a specific project and the latter means the projects one specific team is working on.
+
+### Naming Guidance
+- Collection name should be hyphen-ated lower case, e.g. commit-files.
+- Collection names must be plural. Avoid using uncountable words because user can’t know if the GET returns one item or a list.
+- Query params and body params are CamelCased.
```
EricHasegawa, line 39: The example is pascal cased no? camel case has lowercase first word

src/docs/api/design.mdx
```diff
@@ -24,3 +25,54 @@ In the above example we flattened `projects`. The table below shows the existing
...
+
+### Naming Guidance
+- Collection name should be hyphen-ated lower case, e.g. commit-files.
+- Collection names must be plural. Avoid using uncountable words because user can’t know if the GET returns one item or a list.
+- Query params and body params are CamelCased. 
+- For sorting and filtering try to stick with the common param names: sortBy (e.g. date_created), orderBy (either asc or desc) groupBy, limit 
+- Path params are snake_cased
+
+## Functionality
+Each API should be **stateless**, have a clear purpose and do one clear thing. To achieve that try to stick with the standard methods below. If you have a need to a more complicated functionality reach out to [owners-api](https://github.com/orgs/getsentry/teams/owners-api) and we can help you achieve it.
```
EricHasegawa, line 44:
```suggestion
Each API should be **stateless**, have a clear purpose and do one clear thing. To achieve that try to stick with the standard methods below. If you require more complicated functionality reach out to [owners-api](https://github.com/orgs/getsentry/teams/owners-api) and we can help you achieve it.
```

src/docs/api/design.mdx
```diff
@@ -24,3 +25,54 @@ In the above example we flattened `projects`. The table below shows the existing
...
+
+- 👍 An API that updates project settings: PATCH or PUT
```
EricHasegawa, line 46: Do we want to distinguish between the two? I.e. `PATCH` is replacing select field, `PUT` replaces the whole object

src/docs/api/design.mdx
```diff
@@ -24,3 +25,54 @@ In the above example we flattened `projects`. The table below shows the existing
...
+- Get count of a resource
+    - Count is part of the `List` API and is provided in header X-Total-Count param
+- Get latest of a resource
+    - Order and filtering should happen as part of list api query parameters. Here’s a [good read](https://www.moesif.com/blog/technical/api-design/REST-API-Design-Filtering-Sorting-and-Pagination/).
+    - 👎 `/api/0/issues/{issue_id}/events/latest/`
+    - 👍 `/api/0/issues/{issue_id}/events?orderBy=-date,limit=1`, `-` for descending
+
+### Batch vs. Single Resource Methods
+Here are some notes that can help you decide between similar methods. We use *Get* here as an example but the same applies to all the other methods in the parenthesis.
+- **Get (Update, Delete)**: Use get on the `{resource}DetailsEndpoint` to retreive a resource. For example, `ProjectDetailsEndpoint`.
```
EricHasegawa, line 71:
```suggestion
- **Get (Update, Delete)**: Use get on the `{resource}DetailsEndpoint` to retrieve a resource. For example, `ProjectDetailsEndpoint`.
```

src/docs/api/design.mdx
```diff
@@ -24,3 +25,54 @@ In the above example we flattened `projects`. The table below shows the existing
...
+- **List (Create, Batch Create, Batch Update, Batch Delete)**: Use get on the `{resource-parent}{resource}Endpoint` to retreive all resources that belong to that parent. For example `TeamProjectsEndpoint`.
+- **Batch Get (Batch Create, Batch Update, Batch Delete)**: Use get on the `{root-parent}{resource}Endpoint`. The difference between `Batch` and `List` is that batch usually includes a list of `ids` as query parameter and retruns details about those ids. This list does not necessarily belong to one parent. For example, we can't retreive two projects that belong to two different teams in the above example and in that case we use the get method in the root resource, in this case `OrganizationProjectsEndpoint`.
```
EricHasegawa, line 73:
```suggestion
- **Batch Get (Batch Create, Batch Update, Batch Delete)**: Use get on the `{root-parent}{resource}Endpoint`. The difference between `Batch` and `List` is that batch usually includes a list of `ids` as query parameter and returns details about those ids. This list does not necessarily belong to one parent. For example, we can't retrieve two projects that belong to two different teams in the above example and in that case we use the get method in the root resource, in this case `OrganizationProjectsEndpoint`.
```

src/docs/api/design.mdx
shanamatthews: I just pushed a commit that turns this into a Note. It seems like the <Note> component is broken in develop, but this works lol

shanamatthews: @sentaur-athena - FYI :D

shanamatthews:
@sentaur-athena - I've got a review in progress, will finish up tomorrow!

src/docs/api/design.mdx
```diff
@@ -20,7 +21,62 @@ In the above example we flattened `projects`. The table below shows the existing
...
+
+## Response Object
+Each response object returned from an API should be a serialized version of the django model associated with the resource. You can see all the existing serializers [here](https://github.com/getsentry/sentry/tree/master/src/sentry/api/serializers/models).
```
shanamatthews, line 80:
```suggestion
Each response object returned from an API should be a serialized version of the Django model associated with the resource. You can see all the existing serializers [here](https://github.com/getsentry/sentry/tree/master/src/sentry/api/serializers/models).
```

src/docs/api/design.mdx
```diff
@@ -20,7 +21,62 @@ In the above example we flattened `projects`. The table below shows the existing
...
+
+**Note:** Some models might have different serializers based on usecase. For example `Project` can be serialized into `DetailedProjectSerializer` or `ProjectSerializer`. Decide which one to use based on your usecase and API scope but **DO NOT RETURN CUSTOM OBJECTS** like `{slug: project_slug, platform: project_platform}`. We want the API responses to be uniform and can be used in multiple automations without adding extra complication to the external developers' code.
```
shanamatthews, line 82:
```suggestion
**Note:** Some models might have different serializers based on use case. For example, `Project` can be serialized into `DetailedProjectSerializer` or `ProjectSerializer`. Decide which one to use based on your use case and API scope but **DO NOT RETURN CUSTOM OBJECTS** like `{slug: project_slug, platform: project_platform}`. We want the API responses to be uniform and useable in multiple automations without adding extra complication to the external developers' code.
```

shanamatthews (approved):

PR #1039, opened 2023-09-25
chore(docs): Update instructions with body param info

schew2381:
View updated instructions [here](https://develop-git-seiji-choreupdate-api-instructions-with-body-5c7898.sentry.dev/api/public/)

This PR updates instructions about how to use body params. It also clarifies how optional/nullable fields work in TypedDicts, and also cleans up some language to be more smooth.

### DIFF
--- a/src/docs/api/public.mdx
+++ b/src/docs/api/public.mdx
@@ -9,12 +9,13 @@ In this document, we explain when and how to publish API documentation to public
 
 ## API Versioning
 
-Currently, Sentry's API is `v0` and considered to be in draft phase. While we don't expect our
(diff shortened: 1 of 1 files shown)


src/docs/api/public.mdx
```diff
@@ -58,33 +68,35 @@ our APIs.
 
 **API Documentation consists of**:
 
-1. Declaring Owner for the Endpoint
-2. Setting Publish Status on Endpoint
+1. Owner
+2. Publish Status
```
schew2381, line 72: I shortened these to be more in line with the other steps

sentaur-athena (approved):

PR #1041, opened 2023-09-28
chore(docs): Clarify private fields on response schema

schew2381:
`here` links to [this example](https://github.com/getsentry/sentry/blob/4828c60fd7f9fce5e32769054b9a7b10dc795aaa/src/sentry/replays/post_process.py#L43-L44)

Screenshot 2023-09-28 at 1 16 45 PM


isabellaenriquez (approved):
lgtm 💅🏽

PR #1044, opened 2023-10-03
chore(docs): Say each serializer must have unique name

schew2381:


src/docs/api/public.mdx
```diff
@@ -295,9 +296,10 @@ validate any included example responses. There are three ways to do this:
   )
   ```
 
-  Note that we HIGHLY recommend using the first method when your endpoint returns a single
-  object and utilizes an existing serializer. The first method ensures the documentation stays 
-  up-to-date as the endpoint evolves, and provides a higher quality API experience for our users.
+  Note that we HIGHLY discourage returning a single TypedDict object if your
+  endpoint utilizes an existing serializer and returns a single object. Instead,
+  please use the first method as this ensures the documentation stays up-to-date as the
+  endpoint evolves.
```
schew2381, line 302: Reworded this to include a more in-depth explanation

cathteng (approved):

sentaur-athena (approved):

PR #1048, opened 2023-10-07
Fix typo in kafka-consumer-groups command in self-hosted troubleshooting docs

bonyuta0204:
…ing docs


hubertdeng123 (approved):
thanks for catching this!

PR #1051, opened 2023-10-13
feat: Simpler column add migration guide

markstory:
We don't need to add columns with dangerous migrations anymore. All column adds can be done as deploy time migrations.


wedamija (approved):
Damn, didn't realise this was in the docs. Thanks!

PR #1058, opened 2023-10-18
docs: Add note to restart snuba after enabling relay

Dav1dde:
After enabling Relay and Kafka, the snuba container configuration needs to be updated, this only happens when re-creating the container.

Container creation is conditional on `SENTRY_EVENTSTREAM`:
https://github.com/getsentry/sentry/blob/9c487382d459203ebe4109f6b677507f03ca0891/src/sentry/conf/server.py#L2747-L2778

### DIFF
--- a/src/docs/environment/index.mdx
+++ b/src/docs/environment/index.mdx
@@ -280,6 +280,13 @@ SENTRY_FEATURES['organizations:metrics-extraction'] = True  # Enables session me
 SENTRY_FEATURES['organizations:transaction-metrics-extraction'] = True  # Enables transaction metrics
 ```
 
+After enabling relay the `snuba` service has to be reset to pickup the new configuration:
+
+```shell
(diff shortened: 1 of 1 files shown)


src/docs/environment/index.mdx
```diff
@@ -280,6 +280,13 @@ SENTRY_FEATURES['organizations:metrics-extraction'] = True  # Enables session me
 SENTRY_FEATURES['organizations:transaction-metrics-extraction'] = True  # Enables transaction metrics
 ```
 
+After enabling relay the `snuba` service has to be reset to pickup the new configuration:
```
jjbayer, line 283:
```suggestion
After enabling `KafkaEventStream` the `snuba` service has to be reset to pick up the new configuration:
```

jjbayer (approved):

PR #1061, opened 2023-10-20
feat: Replace docker-compose with "docker compose"

GeekCornerGH:
Figured out install script showed "docker compose" but docs mentionned "docker-compose"


hubertdeng123 (approved):
thanks for submitting this!

PR #1072, opened 2023-11-03
ref(mail): Add mail preview docs

ceorourke:
Update a TODO with information about how to set up the local mail preview so when you're developing you can actually trigger emails.

### DIFF
--- a/src/docs/services/email.mdx
+++ b/src/docs/services/email.mdx
@@ -10,8 +10,6 @@ Inbound email is fairly limited in use, and currently it only supports processin
 
 You'll need to configure an SMTP provider for outbound email.
 
-TODO: Document mail preview backend.
-
 <dl>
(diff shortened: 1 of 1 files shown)


src/docs/services/email.mdx
```diff
@@ -21,6 +19,8 @@ The backend to be used for email delivery. Options are `smtp`, `console`, and `d
 
 Defaults to `smtp`. Use `dummy` if you’d like to disable email delivery.
 
+Use `sentry.utils.email.PreviewBackend` to receive mail locally while developing. By default they'll open in the native Mail app if using a Macbook.
```
ceorourke, line 22: I'm on the fence about adding a mention of having to select iCloud as the provider and sign in with my work email (and specifying this is for Sentry staff). It seems really specific but also is what I was having a hard time with.

scttcper (approved):

PR #1110, opened 2023-12-14
docs(stories): Add some initial docs for the /stories/ Component Library

ryan953:
You should be able to visit the Component Library on any sentry instance, like on sentry.io: https://sentry.io/orgredirect/organizations/:orgslug/stories/

This is the first pass at some docs about it. Why we have it, how to get there, how to contribute and some ideas for the future!

### DIFF
--- a/src/docs/frontend/component-library.mdx
+++ b/src/docs/frontend/component-library.mdx
@@ -1,0 +1,151 @@
+---
+title: UI Component Library
+---
+
+<Alert level="info">
+  To jump straight in visit <Link to="https://sentry.io/orgredirect/organizations/:orgslug/stories/"><code>/stories/</code></Link> on your existing Sentry Org/Install.
+</Alert>
+
+## What is a Component Library
+
+This component library is a (growing) collection of all the React components and hooks that are available within the getsentry/sentry codebase. 
+
+The library consists of the framework itself, which helps to make stories available & discoverable, as well as the story content. Each story is meant to describe how to use a React Component or hook; what are the properties, behaviors, and common or recommended combinations.
+
+## Accessing and Using the Library
+
+The Component Library is implemented as a webpage inside the main sentry application. Whenever Sentry is running, the Library is available.
+
+- **Prod Access**<br />
+    Log in and then visit [`/stories/`](https://sentry.io/orgredirect/organizations/:orgslug/stories/) on your existing Sentry Org. This works the same on sentry.io or for self-hosted installations.
+
+- **Dev Access**<br />
+    Whether you are using `yarn dev-ui` to start a development server, or `sentry devserver` to spin up all the services locally, you can access the Component Library by logging in and changing the path from `/issues/` (the default landing page) to `/stories/`. 
+
+## Creating new Stories
+
+### Using the `storyBook()` framework
+
+The easiest way to start documenting a single component is to import and use `storyBook()`.
+
+Here's a template to copy, edit, and get started with:
+
+```typescript
+// Filename: components/button.stories.tsx
+
+import {Button} from 'sentry/components/button'; // replace with your component
+import storyBook from 'sentry/stories/storyBook';
+
+export default storyBook(Button, story => {
+  story('Default', () => <Button>Default Button</Button>);
+});
+```
+
+This template includes all the basics:
+
+1. The filename is `button.stories.tsx` and sits directly beside `button.tsx` which is the component being described.
+2. The file uses a default export, there is no need to name things when there is only one export.
+3. To create the `storyBook()` we're passing in the `Button` component directly as the first argument. This sets the title at the top of the page, and avoids typos.
+4. We've implemented a function that accepts `story` as it's argument. This is inspired by `jest` and the [`describe()`](https://jestjs.io/docs/api#describename-fn) function.
+5. Each time we call `story()` we can render out a unit onto the page. This includes a title, and the rendered component example. This is similar to `jest` and the [`test()` or `it()`](https://jestjs.io/docs/api#testname-fn-timeout) function, with the exception that `story` is not a global.
+    
+Using this framework you get some nice things automatically:
+
+- Stories are printed in the same order they are written in the source file
+- Each `story()` callback can be async
+
+Some tips to remember:
+
+- Don't be afraid to import anything else you need to help illustrate use cases for your component. It's useful to leverage `<p>` and `<Fragment>` in most cases for example.
+- If you are demonstrating a component that relies on a global hook like `useOrganization()` or `usePageFilters()`, then you should wrap your story with those context providers manually.
+- Named exports, or a mix of named and default exports are all supported. This is more useful with [non-framework stories](#non-framework-stories).
+- If you are testing a hook then you must pass a string as the first argument to `storyBook()` instead of the hook itself because.
+
+### Non-framework stories
+
+It is possible to skip the `storyBook()` framework function and build stories from scratch. 
+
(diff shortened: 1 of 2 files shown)


src/docs/frontend/component-library.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+    // will render as <code>&lt;IconFire<br/>color="red"<br/>size="sm" \:gt;</code>
+    ```
+
+- [`<JSXProperty />`](https://github.com/getsentry/sentry/blob/master/static/app/components/stories/jsxProperty.tsx)
+    Render a formatted JSX property name & value.<br />
+    ```typescript
+    return <JSXProperty name="disabled" value={false} />;
+
+    // will render as `<code>size={false}</code>`
```
michellewzhang, line 95:
```suggestion
    // will render as `<code>disabled={false}</code>`
```

src/docs/frontend/component-library.mdx
```diff
@@ -0,0 +1,151 @@
...
+Helper Components:
+
+<IconFire color="gray500" size="sm" />
+
+- [`<JSXNode />`](https://github.com/getsentry/sentry/blob/master/static/app/components/stories/jsxNode.tsx)<br />
+    Render a formatted JSX component name and some properties.<br />
+    ```typescript
+    return <JSXNode name="IconFire" props={{color: 'red400', size: 'sm'}} />;
+
+    // will render as <code>&lt;IconFire<br/>color="red"<br/>size="sm" \:gt;</code>
```
michellewzhang, line 87: are the `&lt;` and `\:gt;` intentional?

src/docs/frontend/component-library.mdx
```diff
@@ -0,0 +1,151 @@
...
+    By default uses `display:flex; overflow: hidden;` which can test responsive components well. Can also be set to `display: block; overflow: auto;`.<br />
+    ```typescript
+    <SizingWindow style={{height: '100px'}}>
+      <LoadingTriangle />
+    </SizingWindow>`
+    ```
+
+- [`<Matrix />`](https://github.com/getsentry/sentry/blob/master/static/app/components/stories/matrix.tsx)<br />
+    A helper to render multiple pairs of properties and values in a grid.
+    For example, we can render all button `priority` values against all possible `size` values which results in a 4x5 matrix of outputs.
```
michellewzhang, line 121: might be more clear if you say something like "render 2 button `priority` values against 2 possible `size` values which results in a 2x2..." which matches the example below, or you could match the example to the text

scttcper:
document the documentation

src/docs/frontend/component-library.mdx
ryan953: Yes. Trying to illustrate that we will output visible angle-braces to the screen. `&gt;` is the html escape code for that.

src/docs/frontend/component-library.mdx
ryan953:
```suggestion
    For example, we can render N values for one props, against M values for another prop. The example below compares 4 values for button `size` against 5 values for `priority`, resulting in a 20 example buttons in a grid.
```

src/docs/frontend/component-library.mdx
```diff
@@ -0,0 +1,151 @@
...
+    ```typescript
+    <Matrix
+      render={() => Button}
+      propMatrix={{
+        priority: ['default', 'primary'],
+        size: ['md', 'sm'],
```
ryan953, line 127:
```suggestion
        priority: ['default', 'primary', 'danger', 'link', undefined],
        size: ['md', 'sm', 'xs', 'zero'],
```

src/docs/frontend/component-library.mdx
```diff
@@ -0,0 +1,151 @@
...
+  story('Default', () => <Button>Default Button</Button>);
+});
+```
+
+This template includes all the basics:
+
+1. The filename is `button.stories.tsx` and sits directly beside `button.tsx` which is the component being described.
+2. The file uses a default export, there is no need to name things when there is only one export.
+3. To create the `storyBook()` we're passing in the `Button` component directly as the first argument. This sets the title at the top of the page, and avoids typos.
+4. We've implemented a function that accepts `story` as it's argument. This is inspired by `jest` and the [`describe()`](https://jestjs.io/docs/api#describename-fn) function.
```
gggritso, line 49:
```suggestion
4. We've implemented a function that accepts `story` as its argument. This is inspired by `jest` and the [`describe()`](https://jestjs.io/docs/api#describename-fn) function.
```

gggritso (approved):
B-e-a utiful

src/docs/frontend/component-library.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+<Alert level="info">
+  To jump straight in visit <Link to="https://sentry.io/orgredirect/organizations/:orgslug/stories/"><code>/stories/</code></Link> on your existing Sentry Org/Install.
+</Alert>
+
+## What is a Component Library
+
+This component library is a (growing) collection of all the React components and hooks that are available within the getsentry/sentry codebase. 
+
+The library consists of the framework itself, which helps to make stories available & discoverable, as well as the story content. Each story is meant to describe how to use a React Component or hook; what are the properties, behaviors, and common or recommended combinations.
```
michellewzhang, line 13:
```suggestion
The library consists of the framework itself, which helps to make stories available & discoverable, as well as the story content. Each story is meant to describe how to use a React Component or hook, as well as the properties, behaviors, and common or recommended combinations.
```

src/docs/frontend/component-library.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+export default storyBook(Button, story => {
+  story('Default', () => <Button>Default Button</Button>);
+});
+```
+
+This template includes all the basics:
+
+1. The filename is `button.stories.tsx` and sits directly beside `button.tsx` which is the component being described.
+2. The file uses a default export, there is no need to name things when there is only one export.
```
michellewzhang, line 47:
```suggestion
2. The file uses a default export; there is no need to name things when there is only one export.
```

src/docs/frontend/component-library.mdx
```diff
@@ -0,0 +1,151 @@
...
+Using this framework you get some nice things automatically:
+
+- Stories are printed in the same order they are written in the source file
+- Each `story()` callback can be async
+
+Some tips to remember:
+
+- Don't be afraid to import anything else you need to help illustrate use cases for your component. It's useful to leverage `<p>` and `<Fragment>` in most cases for example.
+- If you are demonstrating a component that relies on a global hook like `useOrganization()` or `usePageFilters()`, then you should wrap your story with those context providers manually.
+- Named exports, or a mix of named and default exports are all supported. This is more useful with [non-framework stories](#non-framework-stories).
```
michellewzhang, line 61:
```suggestion
- Named exports, or a mix of named and default exports, are all supported. This is more useful with [non-framework stories](#non-framework-stories).
```

src/docs/frontend/component-library.mdx
```diff
@@ -0,0 +1,151 @@
...
+- If you are testing a hook then you must pass a string as the first argument to `storyBook()` instead of the hook itself because.
```
michellewzhang, line 62: because what? 😵

michellewzhang (approved):
gif

src/docs/frontend/component-library.mdx
ryan953: just because!

PR #1115, opened 2023-12-21
docs(stories): Add a note about using storyBook() with hooks and HoC wrappers

ryan953:
Related to https://github.com/getsentry/sentry/pull/62229#pullrequestreview-1793692363

### DIFF
--- a/src/docs/frontend/component-library.mdx
+++ b/src/docs/frontend/component-library.mdx
@@ -67,6 +67,7 @@ Some tips to remember:
 - Don't be afraid to import anything else you need to help illustrate use cases for your component. It's useful to leverage `<p>` and `<Fragment>` in most cases for example.
 - If you are demonstrating a component that relies on a global hook like `useOrganization()` or `usePageFilters()`, then you should wrap your story with those context providers manually.
 - Named exports, or a mix of named and default exports, are all supported. This is more useful with [non-framework stories](#non-framework-stories).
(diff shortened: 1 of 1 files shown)


michellewzhang (approved):

michellewzhang:
just ran into this today i was so confused! great note

src/docs/frontend/component-library.mdx
```diff
@@ -67,6 +67,7 @@ Some tips to remember:
 - Don't be afraid to import anything else you need to help illustrate use cases for your component. It's useful to leverage `<p>` and `<Fragment>` in most cases for example.
 - If you are demonstrating a component that relies on a global hook like `useOrganization()` or `usePageFilters()`, then you should wrap your story with those context providers manually.
 - Named exports, or a mix of named and default exports, are all supported. This is more useful with [non-framework stories](#non-framework-stories).
+- You can pass the component itself into `storyBook()` to get a nice title on the page. This avoids typos. But be careful of exceptions and use a string instead! For example, in a production build components that use `forwardRef` or HoC's can have minified names. Also hook names are always minified. Pass a string instead.
```
malwilley, line 70: Appreciate the note! Might be nice to have in the code too since I know I would look there first and not realize there are docs here about it.

And just a little wording tweak:

```suggestion
- You can pass the component itself into `storyBook()` to get a nice title on the page. This avoids typos. But be careful of scenarios where a string should be used instead! For example, in a production build components that use `forwardRef` or HoC's can have minified names. Also hook names are always minified. Pass a string instead.
```

malwilley (approved):

PR #1130, opened 2024-01-16
Update backup/restore docs to allow parameterization for scripts

hubertdeng123:


hubertdeng123:
@azaslavsky Looks like `sentry-admin` has been in place for a while now and does pretty much the same thing as `backup/restore`. Should we actually just remove the backup/restore scripts in favor of that?

azaslavsky:
I think we should be careful about removing those, since users may have automation etc that relies on them? I kind of like `backup`/`restore` as the easy option, while `sentry-admin` has the more full-featured UI. Wdyt?

PR #1149, opened 2024-02-09
docs: resolve typo in Breaking Changes part of self-hosted docs

pataar:
According to the example config file, it should be `PyMemcacheCache` not `PyMemcachedCache`


aldy505:
cc @hubertdeng123 @azaslavsky

azaslavsky (approved):

PR #1151, opened 2024-02-10
docs(grouping): Fix a few typos in grouping docs

mrduncan:


lobsterkatie (approved):
Man, we really should run a spell-checker over this whole thing. I get that there'd be a million false positives, but still...

PR #1153, opened 2024-02-12
Add frames delay

philipphofmann:
Add docs for the performance metric frames delay.

### DIFF
--- a/src/docs/sdk/performance/frames-delay.mdx
+++ b/src/docs/sdk/performance/frames-delay.mdx
@@ -1,0 +1,145 @@
+---
+title: "Frames Delay"
+---
+
+Frames Delay is the user-perceived delayed duration of rendered frames. SDKs can attach this
+information to spans. We recommend implementing this feature for mobile and desktop SDKs.
+
+## Background
+
+This section explains why frames delay is a better metric than slow and frozen frames. If you
+already know why, you can skip this section.
+
+On Mobile, slow and frozen frames are well-established metrics for tracking the responsiveness of
+applications. A phone or tablet typically renders 60 frames per second (fps). At 60 fps, every frame
+has 16 or 16.67 ms to render: 
+
+* __Slow Frames__: Using 60 fps, slow frames are frames that take more than 16 ms (Android) or 16.67 ms
+(iOS) to render.
+* __Frozen Frames__: Frozen frames are frames that take longer than 700 ms to render.
+
+The major downside is that they lack essential information on how long they lasted, which is a
+significant problem when prioritizing which unresponsive application parts need improvement. The following
+scenarios highlight the problem when prioritizing by looking at slow and frozen frames.
+
+Scenario | Slow Frames | Frozen Frames |
+-- | -- | -- | --
+1 |  5 all 500ms | 0 | 
+2 | 10 all 20ms | 0 | 
+3 |  0 | 1 with 800ms |
+
+When comparing the different scenarios, a user might want to fix scenario 3 first because it contains one frozen
+frame. When comparing scenarios 1 and 2, they would pick scenario 2 cause it has more slow frames. While scenario
+2 looks worse, the user experience is way worse for scenario 1 because the slow frames last longer.
+
+That's where a new metric called __frames delay__ comes to the rescue. Frames delay represents the delayed frame
+render duration of all frames. The frame delay for one frame is `actual frame duration - expected frame duration`.
+For example, with 60 fps the expected frame duration is 16.67 ms. When a frame takes 20 ms to render,  the frame
+delay is `20ms - 16.67ms = 3.33ms`. 
+
+Let's calculate the frames delay for scenarios 1 - 3.
+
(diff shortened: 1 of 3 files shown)


src/docs/sdk/performance/frames-delay.mdx
```diff
@@ -0,0 +1,145 @@
...
+    And 2 frame with 500 ms with 120 fps
+    And 10 frames with 8.33 ms with 120 fps
+    When calculating the frames delay
+    Then the frames delay is 200 - 16.67 + (500 - 8.33) * 2 = 1166.67 ms
+
+Scenario: Missing delayed frame information
+    """
+    [nf][ ------- ?? ------ ]   |  nf = normal frame,  ?? = No frame information
+    [---- time interval ----]
+    """
```
brustolin, line 87: l: If we could change this type of text graph into an actual graph image, it would be so much faster to understand.

src/docs/sdk/performance/frames-delay.mdx
```diff
@@ -0,0 +1,145 @@
...
+    Given a frame rate of 60 fps
+    And 1 frame with 16.67 ms
+    And 200 ms pass after without any frame being drawn
```
kahest, line 90: when would that happen? does it make sense to provide an example?

kahest (commented):
LGTM, just left a clarifying question. I'm pretty familiar with the concept already, so it'll be good to have some more eyes on this

src/docs/sdk/performance/frames-delay.mdx
```diff
@@ -0,0 +1,145 @@
...
+Frames delay represents the delayed frame render duration of all frames. The frame delay for one frame is
+`actual frame duration - expected frame duration`. For example, with 60 fps the expected frame duration is 16.67 ms.
+When a frame takes 20 ms to render,  the frame delay is `20ms - 16.67ms = 3.33ms`. As the expected frame rate can
+change at any time, SDKs must check the expected frame rate for every frame. The frame delay for a span is the sum
+of frame delays for all slow and frozen frames during the start and end time of the span.
+
+SDKs should attach frames delay in seconds to spans via <Link to="/sdk/performance/span-data-conventions/#mobile">span data</Link>.
+For transactions SDKs should send frames delay via the root span with span data.
+While it makes sense to calculate frames delay in milliseconds, __SDKs must send frames delay in seconds__, as SDKs already
+use seconds for most timestamps. For the specification here, we use milliseconds cause they are easier to read.
```
brustolin, line 62: m: Although it says "frame delays should be added to spans and transactions", I'm missing an explanation about when to start and end the measurement of the frame delay. We take it for granted that it is the duration of the span/transaction itself, but it's not so obvious.
Later in the document, in the scenarios, we have these graphs where it says `[---time interval---]`, but nowhere in the document does it specify what this "time interval" refers to.

brustolin (approved):
It looks good.
Left a couple of non blocking comments

src/docs/sdk/performance/frames-delay.mdx
philipphofmann: Thanks for pointing that out. I updated the scenario.

src/docs/sdk/performance/frames-delay.mdx
philipphofmann: Sadly, I can't put images directly into the code block. While I could split the scenarios into multiple blocks, maintaining the images is harder. Therefore, I want to keep the textural description.

src/docs/sdk/performance/frames-delay.mdx
philipphofmann: Good catch, thanks 👏

stefanosiano:
looks good. 
I would just clarify explicitly the case of concurrent spans: what happens when 2 spans are running? 
should the frame delay be added to both (each with its own value) or only to one of the two?
e.g.
span 1 -> start
span 2 -> start
span 1 -> finish with fdd 10ms
span2 -> finish with fdd 50ms
should we send 10 as frame delay for span1 and 50 for span2?

philipphofmann:
@stefanosiano, I updated the docs. Is it clear now how the logic should handle concurrent spans? If not, please tell me how to make it clearer.

src/docs/sdk/performance/frames-delay.mdx
```diff
@@ -0,0 +1,160 @@
...
+## Background
+
+This section explains why frames delay is a better metric than slow and frozen frames. If you
+already know why, you can skip this section.
+
+On Mobile, slow and frozen frames are well-established metrics for tracking the responsiveness of
+applications. A phone or tablet typically renders 60 frames per second (fps). At 60 fps, every frame
+has 16 or 16.67 ms to render: 
+
+* __Slow Frames__: Using 60 fps, slow frames are frames that take more than 16 ms (Android) or 16.67 ms
```
markushi, line 17: On Android frame metrics are reported as nano durations, I think we can unify this to `16.67ms`

markushi (approved):
LGTM!

PR #1166, opened 2024-02-21
feat: Document new multi craft config feature

Lms24:
This PR adds documentation to our Craft/publishing setup guide in the develop docs: Since we completed the work on https://github.com/getsentry/publish/issues/3441, we now document how to configure a repo's publishing process to use multiple craft configs, depending on the selected merge target branch.

### DIFF
--- a/src/docs/sdk/craft-quick-start.mdx
+++ b/src/docs/sdk/craft-quick-start.mdx
@@ -1,5 +1,5 @@
 ---
-title: 'Craft Quick Start'
+title: "Craft Quick Start"
 ---
 
 This guide goes over the minimal setup needed to publish something via
@@ -45,8 +45,8 @@ We're just interested in making a GitHub release and a PyPI release so our
 ```yaml
 minVersion: 0.28.1
 targets:
-  - name: pypi
-  - name: github
+- name: pypi
+- name: github
 ```
 
 ### `scripts/bump-version.sh`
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):

asottile-sentry:
do we want to encourage this by documenting it?  I was hoping this is a one-off special case and not something we'll need to repeat

Lms24:
> do we want to encourage this by documenting it? I was hoping this is a one-off special case and not something we'll need to repeat

So, I want to document this somewhere because
- we might need it in the future and right now theres maybe 2-3 people who know what we changed. Not ideal.
- Sentry CLI will probably want to opt into this procedure but ideally it's the repo maintainers who take care of this (again that they know how their publishing config is set up). 

I also thought about adding it to the readme of publish if you prefer this. However, given that this page in the develop docs is the only one that talks about craft config _and_ prepare-release/publish stuff I figured it'd make more sense here.

asottile-sentry:
> > do we want to encourage this by documenting it? I was hoping this is a one-off special case and not something we'll need to repeat
> 
> So, I want to document this somewhere because
> 
> * we might need it in the future and right now theres maybe 2-3 people who know what we changed. Not ideal.
> * Sentry CLI will probably want to opt into this procedure but ideally it's the repo maintainers who take care of this (again that they know how their publishing config is set up).
> 
> I also thought about adding it to the readme of publish if you prefer this. However, given that this page in the develop docs is the only one that talks about craft config _and_ prepare-release/publish stuff I figured it'd make more sense here.

I was thinking the PRs themselves should be enough -- they all link to each other and it's pretty clear how the things tie together right?

Lms24:
As discussed internally, I merged this PR, given I see the value in having this documented outside of a PR/issue. We can always revert or refactor at a later time once we have more permanent solution.

PR #1168, opened 2024-02-26
Removed obsolete warning

antonpirker:


antonpirker:
Because this is just something only concerning a Python update that has not yet been shipped, I am just merging it.

PR #1173, opened 2024-02-28
Add pagination section for apis

ykamo001:
Adding a section that describes what pagination is, and how to implement it at Sentry

### DIFF
--- a/src/docs/api/concepts.mdx
+++ b/src/docs/api/concepts.mdx
@@ -121,6 +121,7 @@ GET /api/0/projects/{project_slug}/teams
 2. The endpoint can sometimes end in a plural (s), but more importantly, it does __not__ end in an identifier (`*_slug`, or `*_id`).
 
 To paginate a response at Sentry, you can leverage the [`self.paginate`](https://github.com/getsentry/sentry/blob/24.2.0/src/sentry/api/base.py#L463-L476) method as part of your endpoint.
+`self.paginate` is the standardized way we paginate at Sentry, and it helps us with unification of logging and monitoring.
 You can find multiple [examples of this](https://github.com/getsentry/sentry/blob/24.2.0/src/sentry/api/endpoints/api_applications.py#L22-L33) in the code base. They'll look something like:
 ```python
 def get(self, request: Request) -> Response:
--- a/src/docs/api/concepts.mdx
+++ b/src/docs/api/concepts.mdx
@@ -97,12 +97,10 @@ GET /api/0/projects/{project_slug}/?collapse=stats
 This is typically only needed if the endpoint is already public and we do not want to introduce a breaking change. Remember, if the endpoint is public and we remove an attribute, it is a breaking change. If you are iterating on an undocumented endpoint, return the minimal set of attributes and rely on the `expand` parameter to get more detailed information.
 
 ## Paginating responses
->> APIs often need to provide collections of data, most commonly in the `List` standard method.
-However, collections can often be arbitrarily sized, and also often grow over time, increasing lookup time as well as the size of the responses being sent over the wire.
-Therefore, it is important that collections be paginated.
+>> APIs often need to provide collections of data, most commonly in the `List` standard method. However, collections can be arbitrarily sized, and tend to grow over time, increasing lookup time as well as the size of the responses being sent over the wire. This is why it's important for collections to be paginated.
 
-Paginating responses is a [standard practice for APIs](https://google.aip.dev/158), and we follow them at Sentry as well.
-We've seen an example of a `List` endpoint above, and these endpoints have two tell-tale signs:
+Paginating responses is a [standard practice for APIs](https://google.aip.dev/158), which Sentry follows.
+We've seen an example of a `List` endpoint above; these endpoints have two tell-tale signs:
 ```json
 GET /api/0/projects/{project_slug}/teams
 [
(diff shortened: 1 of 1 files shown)


src/docs/api/concepts.mdx
```diff
@@ -95,3 +95,49 @@ GET /api/0/projects/{project_slug}/?collapse=stats
...
+		"name": "Team 2",
+		"slug": "team2",
+  }
+]
+
+```
+1. The endpoint returns an array, or multiple, objects instead of just one.
+2. The endpoint can sometimes end in a plural (s), but more importantly, it does __not__ end in an identifier (`*_slug`, or `*_id`).
+
+To paginate a response at Sentry, you can leverage the [`self.paginate`](https://github.com/getsentry/sentry/blob/24.2.0/src/sentry/api/base.py#L463-L476) method as part of your endpoint.
```
sentaur-athena, line 125: Can you add something around the line that self.paginate is the standardized way we paginate at sentry which helps us with unification of logging and monitoring?

sentaur-athena (approved):
Looks good. Just one comment.

src/docs/api/concepts.mdx
ykamo001: yeah sure!

src/docs/api/concepts.mdx
```diff
@@ -95,3 +95,49 @@ GET /api/0/projects/{project_slug}/?collapse=stats
 ```
 
 This is typically only needed if the endpoint is already public and we do not want to introduce a breaking change. Remember, if the endpoint is public and we remove an attribute, it is a breaking change. If you are iterating on an undocumented endpoint, return the minimal set of attributes and rely on the `expand` parameter to get more detailed information.
+
+## Paginating responses
+>> APIs often need to provide collections of data, most commonly in the `List` standard method.
+However, collections can often be arbitrarily sized, and also often grow over time, increasing lookup time as well as the size of the responses being sent over the wire.
+Therefore, it is important that collections be paginated.
```
lizokm, line 102:
```suggestion
>> APIs often need to provide collections of data, most commonly in the `List` standard method. However, collections can be arbitrarily sized, and tend to grow over time, increasing lookup time as well as the size of the responses being sent over the wire. This is why it's important for collections to be paginated.
```

src/docs/api/concepts.mdx
```diff
@@ -95,3 +95,49 @@ GET /api/0/projects/{project_slug}/?collapse=stats
...
+
+Paginating responses is a [standard practice for APIs](https://google.aip.dev/158), and we follow them at Sentry as well.
+We've seen an example of a `List` endpoint above, and these endpoints have two tell-tale signs:
```
lizokm, line 105:
```suggestion
Paginating responses is a [standard practice for APIs](https://google.aip.dev/158), which Sentry follows.
We've seen an example of a `List` endpoint above; these endpoints have two tell-tale signs:
```

src/docs/api/concepts.mdx
```diff
@@ -95,3 +95,49 @@ GET /api/0/projects/{project_slug}/?collapse=stats
...
+		"slug": "team2",
+  }
+]
+
+```
+1. The endpoint returns an array, or multiple, objects instead of just one.
+2. The endpoint can sometimes end in a plural (s), but more importantly, it does __not__ end in an identifier (`*_slug`, or `*_id`).
+
+To paginate a response at Sentry, you can leverage the [`self.paginate`](https://github.com/getsentry/sentry/blob/24.2.0/src/sentry/api/base.py#L463-L476) method as part of your endpoint.
+You can find multiple [examples of this](https://github.com/getsentry/sentry/blob/24.2.0/src/sentry/api/endpoints/api_applications.py#L22-L33) in the code base, and it looks something like:
```
lizokm, line 126:
```suggestion
You can find multiple [examples of this](https://github.com/getsentry/sentry/blob/24.2.0/src/sentry/api/endpoints/api_applications.py#L22-L33) in the code base. They'll look something like:
```

src/docs/api/concepts.mdx
```diff
@@ -95,3 +95,49 @@ GET /api/0/projects/{project_slug}/?collapse=stats
...
+        request=request,
+        queryset=queryset,
+        order_by="name",
+        paginator_cls=OffsetPaginator,
+        on_results=lambda x: serialize(x, request.user),
+    )
+```
+
+The example above uses an offset type paginator, but you are free to use whatever paginator type suits your endpoint needs.
+There are some [existing types](https://github.com/getsentry/sentry/blob/24.2.0/src/sentry/api/paginator.py) that you can leverage out of the box, or you can extend the base class [`BasePaginator`](https://github.com/getsentry/sentry/blob/24.2.0/src/sentry/api/paginator.py#L48) and implement your own.
```
lizokm, line 143:
```suggestion
The example above uses an offset type paginator, but feel free to use whatever paginator type suits your endpoint needs.
There are some [existing types](https://github.com/getsentry/sentry/blob/24.2.0/src/sentry/api/paginator.py) that you can leverage out of the box, or you can extend the base class [`BasePaginator`](https://github.com/getsentry/sentry/blob/24.2.0/src/sentry/api/paginator.py#L48) and implement your own.
```

lizokm (approved):
Looks great!

src/docs/api/concepts.mdx
ykamo001: done!

PR #1178, opened 2024-03-04
chore(metrics): Clarify SDK implementation details

markushi:
Clarifies
* when normalization should be applied
* that the flush interval should be 5s
* the `before_emit_metric` hook

### DIFF
--- a/src/docs/sdk/metrics.mdx
+++ b/src/docs/sdk/metrics.mdx
@@ -268,8 +268,8 @@ class Set:
 
 ## Serialization
 
-Metrics are emitted in regular flush intervals (every 10 seconds, with a random offset that
-is detemined once per startup) as envelope items.  The item type is `statsd` and the serialization
+Metrics are emitted in regular flush intervals (every 5 seconds, with a random offset that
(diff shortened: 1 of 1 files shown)


src/docs/sdk/metrics.mdx
```diff
@@ -279,6 +279,29 @@ The format is documented as part of relay: [relay_metrics](https://getsentry.git
...
+    return True
+
+sentry_sdk.init(
+    _experiments={
+        "enable_metrics": True,
+        "before_emit_metric": before_emit_metric
+    }
+)
+```
+
```
markushi, line 303: @cleptric If I remember correctly PHP also has a `before_send_metrics` hook. I guess this could be useful for other SDKs as well?

sl0thentr0py (approved):

src/docs/sdk/metrics.mdx
markushi: Yes, that's exactly the reason why `before_send_metrics` would make little sense for Java. Let's skip this in the docs right now and see if we get any community feedback.

PR #1179, opened 2024-03-05
docs(environment): Add devservices script to setup

s1gr1d:
...as `devservices` need to be started before running the `devserver`


joshuarli (approved):
thanks!

PR #1186, opened 2024-03-08
Update overview.mdx

stephanie-anderson:
We added support for Protobuf a few weeks back. This only works on the new Span ingestion endpoint though. Leaving it to Neel to decide, whether this remark should already be added.


sl0thentr0py:
the protobuf support is for otlp, not our envelope endpoint

PR #1190, opened 2024-03-12
chore: Expand DSN docs to cover subdomain semantics.

markstory:
Explain what the various subdomains are in a DSN and explain the backwards compatibility for US accounts


AbhiPrasad (approved):
Thanks Mark!

PR #1203, opened 2024-03-25
feat: Add migration deployment documentation

markstory:
Expand database migration docs to also cover migration deployment as it is now tool based instead of process based.


src/docs/database-migrations.mdx
```diff
@@ -376,3 +376,60 @@ Renaming columns is dangerous and will result in downtime. The reason this occur
...
+</Alert>
+
+We support two kinds of migrations in our saas deployments:
+
+- Deploy time migrations
+- Post-deploy migrations
+
+### Deployment Migrations
+
+Deployment migrations are run in each region and tenant before code is deployed. Deployment migrations are expected to finish quickly and all statements must complete within 5 seconds. If a migration could take longer because a large number of rows is being operated on, it should be deployed as a post-deploy migration instead. Deployment migrations are ideal for:
```
wedamija, line 394: I don't totally remember, but I thought our timeout here was 2 seconds

src/docs/database-migrations.mdx
```diff
@@ -376,3 +376,60 @@ Renaming columns is dangerous and will result in downtime. The reason this occur
...
+* Adding new tables and columns.
+* Adding indexes to most tables.
+* Removing columns and tables - as long as you follow the processes outlined
+  above.
+
+### Post-deploy migrations (is_dangerous) 
+
+Post-deploy migrations are run manually by engineers **after** a migration has been deployed **to all** regions. During deployment, post-deploy migrations are marked as complete (faked) with django's fake migration behavior. When a post-deploy migration is run, it is run against all regions and tenants. Post-deploy migrations are triggered manually by engineers.
+
+Post-deploy migrations are ideal for:
```
wedamija, line 405: Could we add a warning here to never delete/add columns, or do any other ddl changes other than adding indexes and constraints via post-deploy migration? This is an easy way people can accidentally cause an outage.

I actually wonder if we should build this into the tool as well.

wedamija (approved):
Thanks for updating this!

src/docs/database-migrations.mdx
markstory: We have 5s in the [current production](https://github.com/getsentry/getsentry/blob/master/getsentry/conf/settings/regionsilo.py#L1409-L1410) config.

PR #1209, opened 2024-03-27
feat(metrics): Mobile aggregator memory footprint

philipphofmann:
Add a recommendation for the memory footprint of the aggregator for mobile SDKs.


markushi (approved):
LGTM!

PR #1213, opened 2024-04-03
Meric Normalization

cleptric:
Add the new replacement Regex and replacement map for tag values.


cleptric:
`\w` should match `a-z`, `A-Z`, 0-9, and `_`, so you can use `[^a-zA-Z0-9_]+` instead.

philipphofmann:
@cleptric, I opened a PR to add this info https://github.com/getsentry/develop/pull/1219.

PR #1215, opened 2024-04-06
docs(backup): mention skip-user-creation command for a clean database

L-U-C-K-Y:
Hi all

Not sure, but I think we should use this command to migrate an empty database that is intended to be used as a restore target?


aldy505 (approved):
cc @azaslavsky

hubertdeng123 (approved):

hubertdeng123:
@L-U-C-K-Y Could you rebase master here? Looks like some required CI jobs have been added

PR #1219, opened 2024-04-09
Metric Normalization note on w

philipphofmann:
Explain that \w could include unwanted chars.

### DIFF
--- a/src/docs/sdk/metrics.mdx
+++ b/src/docs/sdk/metrics.mdx
@@ -68,9 +68,7 @@ Normalization shall be done for `statsd` envelope items with unicode support if 
 * **Tag Keys:** regex `[^a-zA-Z0-9_\-.\/]+` with no replacement character.
 * **Tag Values:** See below
 
-**Note:** We can't use `\w` instead of `a-zA-Z0-9` because on some Regex engines `\w` includes
-chars such as `ä`, `ö` or `é`.
-
+{/* Internal note: We can't use `\w` instead of `a-zA-Z0-9` because on some Regex engines `\w` includes chars such as `ä`, `ö` or `é`. */}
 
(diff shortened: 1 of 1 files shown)


cleptric (commented):
Let's update all regexes above instead. It was like this before, but we thought this makes it simpler. Turns out, it does not 😬

src/docs/sdk/metrics.mdx
```diff
@@ -63,11 +63,15 @@ class Aggregator:
...
-* **Tag Keys:** regex `[^\w\-.\/]+` with no replacement character.
+* **Namespaces and Units:** regex `[^a-zA-Z0-9_]+` with no replacement character.
+* **Metric Keys/Names:** regex `[^a-zA-Z0-9_\-.]+` with replacement character `_`.
+* **Tag Keys:** regex `[^a-zA-Z0-9_\-.\/]+` with no replacement character.
 * **Tag Values:** See below
 
+**Note:** We can't use `\w` instead of `a-zA-Z0-9` because on some Regex engines `\w` includes
+chars such as `ä`, `ö` or `é`.
+
+
```
cleptric, line 74: Let's remove this; it adds more confusion than good.

cleptric (approved):

src/docs/sdk/metrics.mdx
philipphofmann: I intended to prevent someone from making the same mistake again. I changed it to an internal note only as a code comment so that it won't be visible in the docs, but only when you edit the .mdx file. Do you still think it is confusing, @cleptric? If yes, I can ditch it.

cleptric: It's fine, thanks!

PR #1222, opened 2024-04-12
Add expected User-Agent format to SDK docs

kahest:
related: https://github.com/getsentry/team-sdks/issues/61

### DIFF
--- a/src/docs/sdk/overview.mdx
+++ b/src/docs/sdk/overview.mdx
@@ -215,6 +215,7 @@ We recommend always sending the following headers:
 
 - `content-type`
 - `content-length`
+- `user-agent`
 
 The following additional headers are permitted as per CORS policy:
 
@@ -228,6 +229,24 @@ The following additional headers are permitted as per CORS policy:
 - `authorization`
 - `content-encoding`
 - `transfer-encoding`
+
+### User Agent
+
+All SDKs are expected to report their name and version via the `user-agent` header. 
+The following format should be used (unless required or recommended differently by the platform, 
+e.g. for browser SDKs):
+
+`{sdk-name}/{sdk-version}`
+
+Additional information about the runtime, operating system, and others can be 
+appended as comments in parentheses, like so:
+
+`{sdk-name}/{sdk-version} ({runtime-name} {runtime-version}; {os-name} {os-version})`
+
+There is no expectation towards the presence or order of fields. The user agent 
+must not contain PII or otherwise sensitive data. In general it should not contain 
+any information that is not already present in the payload.
+
 
 ## Request Compression
 
--- a/src/docs/sdk/overview.mdx
+++ b/src/docs/sdk/overview.mdx
@@ -238,6 +238,12 @@ e.g. for browser SDKs):
 
 `{sdk-name}/{sdk-version}`
 
+For example:
+- `sentry.python/1.45.0`
+- `sentry.php/4.7.0`
+- `sentry-ruby/5.17.3`
+- `sentry.cocoa/8.24.0`
+
 Additional information about the runtime, operating system, and others can be 
 appended as comments in parentheses, separated by `; ` (semicolon and space), like so:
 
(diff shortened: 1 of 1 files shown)


src/docs/sdk/overview.mdx
```diff
@@ -229,6 +230,24 @@ The following additional headers are permitted as per CORS policy:
...
+All SDKs are expected to report their name and version via the `user-agent` header. 
+The following format should be used (unless required or recommended differently by the platform, 
+e.g. for browser SDKs):
+
+`{sdk-name}/{sdk-version}`
+
+Additional information about the runtime, operating system, and others can be 
+appended as comments in parentheses, like so:
+
+`{sdk-name}/{sdk-version} ({runtime-name} {runtime-version}; {os-name} {os-version})`
```
mwarkentin, line 244: Should we define "; " as the expected separator between optional info? Will make it easier to parse UA logs programmatically.

src/docs/sdk/overview.mdx
```diff
@@ -229,6 +230,24 @@ The following additional headers are permitted as per CORS policy:
 - `content-encoding`
 - `transfer-encoding`
 
+### User Agent
+
+All SDKs are expected to report their name and version via the `user-agent` header. 
+The following format should be used (unless required or recommended differently by the platform, 
+e.g. for browser SDKs):
+
+`{sdk-name}/{sdk-version}`
```
jjbayer, line 239: Can we list some examples here? IMO it would make sense if the user agent is always prefixed by `sentry-`, not sure if that's included in the `{sdk-name}`.

src/docs/sdk/overview.mdx
```diff
@@ -229,6 +230,24 @@ The following additional headers are permitted as per CORS policy:
 - `content-encoding`
 - `transfer-encoding`
 
+### User Agent
+
+All SDKs are expected to report their name and version via the `user-agent` header. 
+The following format should be used (unless required or recommended differently by the platform, 
+e.g. for browser SDKs):
```
jjbayer, line 237: I assume this is true?

```suggestion
e.g. for browser SDKs, where the user agent header must be set by the browser itself):
```

src/docs/sdk/overview.mdx
kahest: most SDKs I checked use `sentry.{name}`, ruby uses `sentry-ruby`. if we define a requirement for `sentry-` we'd need to adapt a bunch of SDKs - I'm gonna add the examples I found

jjbayer: 👍 As long as they have `sentry` in the name I think that's fine. I was worried about the user agent showing up as `python/1.2.3`, which would make it hard to attribute to sentry.

src/docs/sdk/overview.mdx
kahest: confirmed by @Lms24. I believe this is the relevant `Transport` impl which doesn't set a `user-agent`:

https://github.com/getsentry/sentry-javascript/blob/ee4091e8850dac6f6cd8970e814f2c7f74d1a74f/packages/types/src/transport.ts#L4

Lms24: correct! Also, [this code](https://github.com/getsentry/sentry-javascript/blob/ee4091e8850dac6f6cd8970e814f2c7f74d1a74f/packages/browser/src/transports/fetch.ts/#L24-L42) shows that we only add headers additionally to the request that are specified by users in `Sentry.init({transportOptions: headers: {...}})`. So if users want to add headers to the request we'll do that but we don't add any by default.

cleptric: @Lms24 what are we doing for server side JS?

Lms24: @cleptric Good point, we [don't set](https://github.com/getsentry/sentry-javascript/blob/ee4091e8850dac6f6cd8970e814f2c7f74d1a74f/packages/node/src/transports/http.ts/#L127-L138) a user-agent either.

cleptric (approved):

PR #1223, opened 2024-04-15
Cache integration

cleptric:


src/docs/sdk/cache.mdx
```diff
@@ -0,0 +1,53 @@
...
+## Span conventions
+
+### Span Operations
+
+| Span OP | Description |
+|:--|:--|
+| `cache.get` | A single item or multiple items are retrieved from the cache |
+| `cache.put` | A single item or multiple items are written to the cache |
+| `cache.remove` | A single item or multiple items are removed from the cache |
+| `cache.flush` | The entire content of the cache is deleted |
```
cleptric, line 17: No strong opinion here, just made the OPs used more concise.

src/docs/sdk/cache.mdx
```diff
@@ -0,0 +1,53 @@
...
+| Data Key | Type | Description | Conditions
+|:--|:--|:--|:--|
+| `cache.hit` | bool | cache hit or miss | required only on read-operations (`cache.get`)
+| `cache.key ` | array | the key(s) involved in the operation | required only on operations that have a key
+
+The following data attributes are only to be set if exposed by the instrumented cache abstraction or retrievable with minimal overhead.
+
+| Data Key | Type | Description | Conditions
+|:--|:--|:--|:--|
+| `cache.item_size ` | int | the size of the item/items read/written/deleted in `bytes` | only on operations that have a key
```
cleptric, line 30: Retrieving the size might be expensive, hence I would opt to make this optional.

src/docs/sdk/cache.mdx
```diff
@@ -0,0 +1,53 @@
...
+| `cache.ttl ` | int | the time to life in `seconds` | only on operations that have a key
+| `network.peer.address ` | string | The hostname of the cache instance |
+| `network.peer.port ` | int | the port used by the cache instance | 
+
+Additionally, the `span.status` can be used to signal if the operation was successful (`success`) or failed (`error`).
```
cleptric, line 35: I'm not sure we can actually use `span.status`, as it was tightly coupled with the HTTP status. TBD, else we'll use `cache.success`.

src/docs/sdk/cache.mdx
antonpirker: `cache.get_item` is used in Python for around 11 months right now, so we need to also support this on the Sentry backend to be backwards compatible.

src/docs/sdk/cache.mdx
antonpirker: The `errro` status is not yet defined here: https://develop.sentry.dev/sdk/event-payloads/span/ 
Do we need to tell anyone on ingest side if we use the new `errror` status?

src/docs/sdk/cache.mdx
DominikB2014: Would retrieving the size be a flag set in the sdk config?

src/docs/sdk/cache.mdx
DominikB2014: To add to @antonpirker point, we document the existing django cache operations here, and it's what ingestion expects rn.

DominikB2014:
@cleptric question about the screenshot, why is there a db query inside the `cache.get`? The `cache.hit` is `true`, so I would assume that the db query would not execute. Or are we using a `db` as a cache here?

cleptric:
This example assumes the DB is used as a cache indeed, using a `cache` table.

src/docs/sdk/cache.mdx
cleptric: We could make this an opt-in config on the integration level, in case the used abstraction does not provide this data.

I'm mostly concerned about the need to serialize none scalar values to gather their size.

src/docs/sdk/cache.mdx
cleptric: Yep, I'll clarify this

PR #1227, opened 2024-04-17
feat: add Linux distributions to the os context

supervacuus:
Users of the Native SDK would prefer to identify and search for Linux distributions their events come from: https://github.com/getsentry/sentry-native/issues/943.

After a preliminary implementation (https://github.com/getsentry/sentry-native/pull/963 and https://github.com/getsentry/relay/pull/3443), I think it is time to nail down the specification of the context attributes.

I consider the current naming of the attributes to be preliminary. I didn't want to force the naming scheme from `/etc/os-release` onto the attributes here (but maybe we should?), and the selection of the attributes is also a bit arbitrary. 

Only `name` and ` version` make sense as indexable entries because the support of other attributes varies wildly between distributions (not only in availability but sometimes also in meaning). Most users probably want to know where issues occur on a specific distribution version rather than necessarily seeing the URL of a distro's issue tracker. 

Nonetheless, we could add all other attributes we find in `/etc/os-release` (in which case I would use the original names. The current mapping is:

| os context name  | /etc/os-release name | Comment
| ------------- | ------------- | ------------- |
| `name` | `ID`  | the actual `NAME` attribute is not stable and changes in some distros between versions
| `version` | `VERSION_ID` | the `VERSION` attribute often includes version aliases
| `pretty_name` | `PRETTY_NAME` |

I am also unsure whether exposing `distribution` as an object rather than prefixed flat attributes is preferred. All these decisions are easy to revert right now, whereas they can be painful once events are out there.

Cc: @kahest

### DIFF
--- a/src/docs/sdk/event-payloads/contexts.mdx
+++ b/src/docs/sdk/event-payloads/contexts.mdx
@@ -273,6 +273,14 @@ output of the `uname` syscall.
 system. For some well-known runtimes, Sentry will attempt to parse `name` and
 `version` from this string, if they are not explicitly given.
 
+`distribution`
+
+: _Optional_. An object that provides meta-data for Linux distributions. Contains the following keys:
+
+- `name`: A name that is stable for each distribution (examples: `ubuntu`, `rhel`, `alpine`).
+- `version`: _Optional_. Typically identifies at least the major release version number. Distributions with rolling releases only, will not provide a version.
+- `pretty_name`: _Optional_. Typically provides the full name, full version, and release alias (examples: `Ubuntu 22.04.4 LTS`, `Raspian GNU/Linux 10 (buster)`).
+
 ### Example OS Context
 
 The OS Context for the 3 major OSs should look like this:
@@ -295,8 +303,12 @@ The OS Context for the 3 major OSs should look like this:
   "linux": {
     "type": "os",
     "name": "Linux",
-    "version": "5.10.6",
-    "build": "arch1-1"
+    "version": "6.1.82(99.168.amzn2023.x86_64)",
+    "distribution": {
+      "name": "amzn",
+      "version": "2023",
+      "pretty_name": "Amazon Linux 2023.4.20240401"
+    }
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


src/docs/sdk/event-payloads/contexts.mdx
```diff
@@ -273,6 +273,14 @@ output of the `uname` syscall.
 system. For some well-known runtimes, Sentry will attempt to parse `name` and
 `version` from this string, if they are not explicitly given.
 
+`distribution`
+
+: _Optional_. An object that provides meta-data for Linux distributions. Contains the following keys:
+
+- `name`: A name that is stable for each distribution (examples: `ubuntu`, `rhel`, `alpine`).
```
Dav1dde, line 280: Is there an exhaustive list of identifiers we can link to?

supervacuus: The official documentation contains most (all?) of the identifiers: https://www.freedesktop.org/software/systemd/man/latest/os-release.html.

I added [test data](https://github.com/getsentry/sentry-native/tree/c18b5ff8953e602af4cf603583e5b62f384a613f/tests/fixtures/os_releases) to the sentry-native repo, of which I can quickly gather an exhaustive list from a representative subset (not to link to, but to compare with the docs).

It would be a good idea to link the `os-release` docs from `develop`, but then it makes sense to show the current mapping, too, right? I think it boils down to whether we want to use their identifiers. 

If we use them, we can say that whatever is in `distribution` is a straight translation from `/etc/os-release`; we link to the official docs, and everyone can look them up or request the addition of new ones.

With any kind of mapping, we can leave it open if we gather distro-related data from other sources in the future, but we need to specify where data comes from individually for each entry.

Initially, I was in the latter camp because the most valuable identifiers might have less obvious names for people not familiar with `/etc/os-release`. But now I think that introducing our own identifier mapping could make things even less obvious.

Dav1dde: I think as long as we're compatible with the "official" identifiers we should link to them, once we diverge either document it here or update the link to where one can see the differences/list of values we use.

Dav1dde (approved):
Thanks for putting all the effort in, I'd just link to some source (you think is best) for the distribution names 👍

src/docs/sdk/event-payloads/contexts.mdx
Dav1dde: Oh yeah sorry, I meant the names like `ubuntu`, `rhel`, `alpine`. Didn't realize identifiers is ambiguous.

PR #1236, opened 2024-04-22
chore(lockreason): Update Lock Reason docs

kahest:
As discussed - Lock Reason is not a top level field but rather part of the threads interface. This PR aims to reflect that by removing it from the sidbar and Event Paylods index page, and adding a note to that effect

### DIFF
--- a/src/docs/sdk/event-payloads/lockreason.mdx
+++ b/src/docs/sdk/event-payloads/lockreason.mdx
@@ -6,6 +6,12 @@ Represents an instance of a held lock (monitor object) in a thread. Typically, t
 to determine which other thread is holding the lock in case the current thread is blocked. In languages like
 `Java` or `Kotlin` this typically will be represented by a `java.lang.Object` type.
 
+<Note>
+
+Lock reasons are always part of a thread. They cannot be declared as a top-level event property. 
+
+</Note>
+
 ## Attributes
 
 `type`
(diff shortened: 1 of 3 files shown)


src/docs/sdk/event-payloads/lockreason.mdx
```diff
@@ -6,6 +6,12 @@ Represents an instance of a held lock (monitor object) in a thread. Typically, t
 to determine which other thread is holding the lock in case the current thread is blocked. In languages like
 `Java` or `Kotlin` this typically will be represented by a `java.lang.Object` type.
 
+<Note>
```
romtsn, line 9: I think we could even embed the interface into this page, before the `Examples` section, wdyt?

romtsn (approved):
thanks for doing this, I didn't give much thought where to put this, but definitely does not belong to the top level.

src/docs/sdk/event-payloads/lockreason.mdx
kahest: I think this would make the page more complex than necessary, also from what I see we usually don't do that for embedded interfaces. But I'll add a link to the Threads interface for easier navigation

romtsn: We kinda do embed them, e.g. [stackframes](https://develop.sentry.dev/sdk/event-payloads/stacktrace/) or [contexts](https://develop.sentry.dev/sdk/event-payloads/contexts/), but this is also fine

kahest: You're right, we do on some pages - IMO in those the embedded interfaces are more essential/required and the parent interface is mostly empty/a collection, so it's a bit different. I think the link is fine

PR #1238, opened 2024-04-22
docs(frontend): Update "Using Hooks" docs to reflect the latest state of the repo

ryan953:


src/docs/frontend/using-hooks.mdx
```diff
@@ -20,61 +20,38 @@ We don't prefer libraries that use or don't use hooks. Instead favor libraries
...
 state.
 
 The `useEffect` hook is more complex and you need to be careful to track your
 dependencies and ensure that subscriptions are canceled through a cleanup
-callback. Complex chained applications of `useEffect` should be avoided and
-'controller' components should remain class-based at this time.
-
-Likewise, the `useReducer` hook overlaps with state-management which currently
-has an undecided future. We would like to avoid *yet another* state management
-pattern, so at this time avoid `useReducer`.
```
ryan953, line 39: Unsubstantiated claim about an `undecided future`.

`useReducer` looks solid: https://react.dev/reference/react/useReducer
things like `useOptimistic` looks like they're still experimental: https://react.dev/reference/react/useOptimistic
I don't see deprecation markers, and don't expect it for this one.

src/docs/frontend/using-hooks.mdx
```diff
@@ -20,61 +20,38 @@ We don't prefer libraries that use or don't use hooks. Instead favor libraries
...
+callback.
 
 ### Using context
 
-As we plan our path away from Reflux, the `useContext` hook provides a simpler
-implementation option to share state and behavior. When you need to create new
```
ryan953, line 44: we're not really planning a path away... but certainly we should, and for new cases it's very likely not needed at all. I don't expect new uses to appear and would discourage it.

src/docs/frontend/using-hooks.mdx
```diff
@@ -20,61 +20,38 @@ We don't prefer libraries that use or don't use hooks. Instead favor libraries
...
-props for your `AsyncComponent`.
-
-### Don't rewrite for hooks
-
-While hooks can be ergonomic in new code we should *avoid* rewriting existing
-code to leverage hooks. Rewriting takes time, exposes us to risk and provides
-very little end-user value.
-
-If you need to rework a component to consume hooks from a library, then also
-consider converting from a class to a function component.
```
ryan953, line 80: Removed language that supported class-based components. Our hooks like `useOrganization` are pervasive and we shouldn't introduce any doubt: all new code should be functional components. But that's not really a topic for this file, more of a topic for frontend/index.mdx

evanpurkhiser (approved):

PR #1242, opened 2024-04-22
docs(frontend): Add a page for Using Styled Components

ryan953:
Two things are in here so far:
1. moved the stuff from index.mdx into the file file, with some formatting and heading changes
2. added a big section called Dynamic Props and some alternative ideas…they all ‘work’ but some of these i don’t think we have examples of in the app yet. Check it out and see how agreeable each suggestion is. I was going for breadth

I'd followup with some more sections in another PR:
- Naming conventions for styled components. ie "Wrapper = styled(...)" is weak.
- Something like ‘subclassing’ or overriding styles
- More notes about values/utils that are in theme (like `overflowEllipsis` for example)


src/docs/frontend/using-styled-components.mdx
```diff
@@ -0,0 +1,221 @@
...
+const Numeric = styled('span')`
+  font-variant-numeric: tabular-nums;
+`;
+
+const SideBySide = styled('div')`
+  display: flex;
+  gap: ${space(2)};
+  flex-wrap: wrap;
+  align-items: flex-start;
+`;
```
gggritso, line 26: 👍🏻 maybe worth mentioning that even styled components deserve good names. e.g., `SideBySide` is a great name, something like `Container` or `StyledPanel` are less good

src/docs/frontend/using-styled-components.mdx
```diff
@@ -0,0 +1,221 @@
...
+  &[data-is-error="true"] {
+    color: ${p => p.theme.error400};
+  }
+  
+  &:disabled {
+    color: ${p => p.theme.gray200};
+  }
+`;
+
+return <Label disabled={false} data-is-error={true}>There is an error</Label>;
```
gggritso, line 103: Styling based on `data-` attributes seems spooky, no? It works, but it's a pretty surprising approach! There's not much precedent for it. Is this really better than regular props?

src/docs/frontend/using-styled-components.mdx
```diff
@@ -0,0 +1,221 @@
...
+  --color: ${p => p.theme.red400};
+`;
+const Info = styled('div')`
+  --color: ${p => p.theme.blue400};
+`;
+
+const label = <Label>There is an error</Label>;
+return isError
+  ? <Error>{label}</Error>
+  : <Info>{label}</Info>;
```
gggritso, line 128: I like this approach but the docs confused me a bit. This would work without CSS variables through the cascade, right? This example feels like it's more about using conditional rendering (choosing the correct wrapper) than about CSS variables

src/docs/frontend/using-styled-components.mdx
```diff
@@ -0,0 +1,221 @@
...
+  <Grid>
+    <HomeLink style={{gridArea: 'logo'}} />
+    <SearchForm style={{gridArea: 'search'}} />
+    <Navigation style={{gridArea: 'sidebar'}} />
+    <DataTable style={{gridArea: 'main'}} />
+  </Grid>
+);
+```
+
+### Imperatively Set Styles
```
gggritso, line 166: Underrated! Scary to jump out of React's default rendering model, but this comes up a lot!

gggritso (commented):
🙌🏻

src/docs/frontend/using-styled-components.mdx
```diff
@@ -0,0 +1,221 @@
+---
+title: Using Styled Components
+---
+
+To apply styles to HTML we use a css-in-js library called [Emotion](https://emotion.sh/). This allows us to define styles in the same file where the component is defined and used. Styles are scoped to avoid global selectors and having to manually specify classnames at strings. You can also leverage the theme object to reference common values and utilities.
```
michellewzhang, line 5:
```suggestion
To apply styles to HTML, we use a CSS-in-JS library called [Emotion](https://emotion.sh/). This allows us to define styles in the same file where the component is defined and used. Styles are scoped to avoid global selectors and having to manually specify classnames as strings. You can also leverage the `theme` object to reference common values and utilities.
```

src/docs/frontend/using-styled-components.mdx
```diff
@@ -0,0 +1,221 @@
...
+`;
+
+return <Label disabled={false} isError={true}>There is an error</Label>;
+```
+
+Choose one of the strategies below instead.
+
+### Use Data Selectors
+
+Data selectors can be an easy way to set styles based on simple/pre-defined state. When defining these attributes don't forget that browser have built-in selectors like `:disabled`, `:focus`, `:first-child` and so on.
```
michellewzhang, line 86:
```suggestion
Data selectors can be an easy way to set styles based on simple/pre-defined state. When defining these attributes, don't forget that browsers have built-in selectors like `:disabled`, `:focus`, `:first-child` and so on.
```

src/docs/frontend/using-styled-components.mdx
ryan953: yeah, i'll take this section out. it's not a strong example of what i was going for... maybe better in a future 'how to override/subclass components' section.

src/docs/frontend/using-styled-components.mdx
ryan953: I think the downside of regular props is that it has more of a runtime cost to re-serialize the styles and inject them into the page. So not only are we re-rendering the react component, but also messing with the styles, creating a lot of redundant-ish classes which have mostly the same properties inside. 

Using data selectors avoids the runtime css changes.

Also I've (rarely) seem examples where the props are passed through to be rendered, like this: 
```
const Thing = styled(props => <span {...props} />)<{hello: string}>``;
return <Thing hello="world" />;
```
So this would just make that valid html

gggritso: That makes sense, I'm just surprised because I haven't seen any precedent of people using `data-` attributes for styling, that's all!

src/docs/frontend/using-styled-components.mdx
```diff
@@ -0,0 +1,198 @@
...
+    <DataTable style={{gridArea: 'main'}} />
+  </Grid>
+);
+```
+
+### Imperatively Set Styles
+
+For values that change frequently it's beneficial to use a component ref and set the values imperatively. This can be done without triggering react which can result in much more efficient updates.
+
+Note that you can set CSS values directly, or if the component is reading a css variable call `elem.style.setProperty('--foo', 'value')` instead.
```
michellewzhang, line 148:
```suggestion
Note that you can set CSS values directly, or if the component is reading a CSS variable, call `elem.style.setProperty('--foo', 'value')` instead.
```

src/docs/frontend/using-styled-components.mdx
```diff
@@ -0,0 +1,198 @@
...
+title: Using Styled Components
+---
+
+To apply styles to HTML, we use a CSS-in-JS library called [Emotion](https://emotion.sh/). This allows us to define styles in the same file where the component is defined and used. Styles are scoped to avoid global selectors and having to manually specify classnames as strings. You can also leverage the `theme` object to reference common values and utilities.
+
+The best styles are the ones you don’t write - whenever possible use existing components.
+
+## Basics
+
+To use Style Components pull in `styled` and apply your css. For consistent spacing you should also use `space()`:
```
michellewzhang, line 11:
```suggestion
To use Style Components pull in `styled` and apply your CSS. For consistent spacing you should also use `space()`:
```

michellewzhang (approved):
left suggestions for minor capitalization things i noticed but overall looks good and comprehensive👍

src/docs/frontend/using-styled-components.mdx
ryan953: I'm gonna followup on this. I don't really know what convention we should push.. there's lots of bad examples too, but i know whatever we write will be optimistic/idealistic for the future.

PR #1244, opened 2024-04-23
Add SDK features to event payload

philipphofmann:
Add a new field features to the SDK event payload, which is handy for tracking the adoption of features.

This is for https://github.com/getsentry/team-sdks/issues/83.

### DIFF
--- a/src/docs/sdk/event-payloads/sdk.mdx
+++ b/src/docs/sdk/event-payloads/sdk.mdx
@@ -54,6 +54,13 @@ have all enabled integrations, including default integrations. Default
 integrations are included because different SDK releases may contain different
 default integrations.
 
+`features`
+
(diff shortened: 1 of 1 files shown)


cleptric (changes requested):
This should become a context, which is much simpler to add to SDKs compared to making changes to the transaction.

philipphofmann:
> This should become a context, which is much simpler to add to SDKs compared to making changes to the transaction.

@cleptric, I don't understand what exactly you mean by making changes to the transaction. Why is this a problem? Extending the SDK interface on Cocoa, for example, is straightforward. 

Adding SDK features to the SDK interface, which contains the SDK name, SDK version, SDK integrations, and SDK packages, is a coherent approach. That's why I chose it over the context.

cleptric (approved):

PR #1248, opened 2024-04-24
Queue Integration

cleptric:
https://develop-git-queue-module.sentry.dev/sdk/performance/modules/queue/

**Producer**

image (4)

**Consumer**

image (5)


antonpirker:
Right now we also include the name of the library that does the enqueuing in the op. See:
https://github.com/getsentry/sentry-python/blob/antonpirker/django-caching-module/sentry_sdk/consts.py#L267-L273

Should we continue to do so or is `messaging.system` span data sufficient for the backend to do some differentiation. I am not sure if this important for the backend.

PR #1255, opened 2024-04-28
docs(self-hosted): recommend users to rerun install.sh for restart

aldy505:
Closes [#2996](https://github.com/getsentry/self-hosted/issues/2996)

/cc @hubertdeng123 @azaslavsky @stayallive

### DIFF
--- a/src/docs/integrations/discord/index.mdx
+++ b/src/docs/integrations/discord/index.mdx
@@ -32,7 +32,7 @@ discord.bot-token: "<bot token>"
 After you update the <code>config.yml</code>, restart your Sentry server to continue the setup process.
 
 <Alert title="Note" level="info">
-    It is preferred to re-run the <code>./install.sh</code> script, instead of restarting all Docker containers. See the <Link to="/self-hosted/#configuration">configuration section</Link> for more information.
+    After changing configuration files, re-run the <code>./install.sh</code> script, to rebuild and restart the containers. See the <Link to="/self-hosted/#configuration">configuration section</Link> for more information.
 </Alert>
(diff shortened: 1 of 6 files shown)


stayallive:
I'm trying to think on how to formulate this better, but I feel like the current suggestion is a bit "soft" as in that it's more of a suggestion while instead it's more of an requirement. Especially if config has been changed.

What about smth more like:

```
After changing configuration files re-run the <code>./install.sh</code> script to rebuild and restart the containers. See the <Link to="/self-hosted/#configuration">configuration section</Link> for more information.
```

(I would also make them all the same, now there are 3/4 versions of exactly the same thing worded slightly different)

azaslavsky:
> I'm trying to think on how to formulate this better, but I feel like the current suggestion is a bit "soft" as in that it's more of a suggestion while instead it's more of an requirement. Especially if config has been changed.
> 
> What about smth more like:
> 
> ```
> After changing configuration files re-run the <code>./install.sh</code> script to rebuild and restart the containers. See the <Link to="/self-hosted/#configuration">configuration section</Link> for more information.
> ```
> 
> (I would also make them all the same, now there are 3/4 versions of exactly the same thing worded slightly different)

Agreed on both counts. We should replicate the exact same callout, and use language like "strongly recommended".

aldy505:
What about now? @stayallive @azaslavsky

azaslavsky (approved):
LGTM!

aldy505:
Ping @hubertdeng123

hubertdeng123 (approved):

PR #1257, opened 2024-04-29
docs(self-hosted): add NGINX as a reverse proxy example

oioki:


aldy505 (approved):
Nice. Although it's very minimal, at least it's here.

It'd be good if you can put up some examples regarding rate limiter middleware (since SH don't have the concept of quotas and billings) and expose only ingest endpoints as specified [here](https://develop.sentry.dev/self-hosted/reverse-proxy#expose-only-ingest-endpoint-publicly).

azaslavsky (approved):
Thank you for adding this! :)

PR #1266, opened 2024-05-10
feat: Add multiregion overview docs

markstory:
Move these out of notion into an easier to find long-term home. More multiregion docs will be added as they are completed.

### DIFF
--- a/src/docs/architecture.mdx
+++ b/src/docs/architecture.mdx
@@ -56,9 +56,9 @@ In order to offer customers data residency we needed to co-locate the bulk of an
 
 These constraints led our design to having multiple 'silo modes'. 
 
-<aside>
-☝ The term ‘silo’ was chosen as other applicable terms (region, zone, partition, cluster) are all taken.
-</aside>
+<Alert level="info">
+    ☝ The term ‘silo’ was chosen as other applicable terms (region, zone, partition, cluster) are all taken.
+</Alert>
 
  Our architecture includes two silo modes:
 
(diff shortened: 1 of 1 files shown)


src/docs/architecture.mdx
```diff
@@ -26,3 +26,54 @@ For more information read [Path of an event through Relay](https://getsentry.git
...
+- Our solution should afford the ability for customers to provide their own hardware for a region in the future.
+- Create a solution that allows single-tenants to join up with saas to help reduce operational overhead and increase our ability to deliver updates to single-tenants.
+- Our solution would need to ‘scale down’ to self-hosted and development environments. It should offer the same functionality in all deployment environments.
+
+In addition to the goals we had several constraints:
+
+- Saas Customers will continue to login and signup at `sentry.io`. They will not need to log into each organization individually.
+- Existing Saas Customers should not be required to update their DSN values. If a customer migrates to a new region, it is ok if they also need to update their DSNs.
+- Generally API URLs should not be required to change. Existing customers can use `[region.sentry.io](http://region.sentry.io)` style URLs to improve latency and ensure data residency requirements are met, but they don’t need to. Customers in new regions may be required to use `region.sentry.io` style URLs.
```
mrduncan, line 49:
```suggestion
- Generally API URLs should not be required to change. Existing customers can use `region.sentry.io` style URLs to improve latency and ensure data residency requirements are met, but they don’t need to. Customers in new regions may be required to use `region.sentry.io` style URLs.
```

Preview doesn't like the combination of code and link

mrduncan (approved):

Issue #1254, opened 2024-04-27
GitHub integration missing callback URL


aldy505:
`${url-prefix}/extensions/github/setup/` should also be included as one of callback URL, otherwise the app will fail. Specifying only the `${url-prefix}/auth/sso/` would fail the installation.

👇🏻 This works

![image](https://github.com/getsentry/develop/assets/7274326/18e221a4-51a5-45f1-82e0-844ec149520c)

Code in question: https://github.com/getsentry/develop/blob/3402f8228985e501a3bd8d87af72e416df6acf23/src/docs/integrations/github.mdx?plain=1#L20-L26

aldy505:
Another case: https://discord.com/channels/621778831602221064/1235674927265484930/1235887106048397392

xHeaven:
Thanks a lot, this solved the integration issue on my self-hosted instance. Honestly, I would've never figured it out by myself.

aldy505:
Closing since https://github.com/getsentry/develop/pull/1235 has been merged

PR #1235, opened 2024-04-22
Extend github integration docs

lucasra1:
When trying to setup the github integration on my self-hosted sentry environment I noticed that it did not work. On further investigation I found out, that github threw an error. I got this error back from github [Github Doc](https://docs.github.com/en/apps/oauth-apps/maintaining-oauth-apps/troubleshooting-authorization-request-errors#redirect-uri-mismatch).

So I found out that I had to add the sentry github setup url to the callback list of my Github App. Extended Documentation to reflect this.

PR #1267, opened 2024-05-12
fix(github): add missing callback url

aldy505:
Closes https://github.com/getsentry/develop/issues/1254


aldy505:
Turns out this is a dupe of #1235

PR #1269, opened 2024-05-12
docs(self-hosted): external storage configurations

aldy505:
Sorry for the ping but I need your feedbacks on this. @hubertdeng123 @azaslavsky @stayallive


src/docs/self-hosted/external-storage.mdx
```diff
@@ -0,0 +1,89 @@
+---
+title: External Storage
```
stayallive, line 2: Maybe we should generalize to "Data Storage" or something.

This way this document can explain where the data is stored by default and can list alternatives if there are any.

Could have the following sections:

- Sentry (with a general explanation about postgres, clickhouse and kafka maybe)
	- Filestore (Uploads, Replays)
		- Database
		- Object Storage
	- Nodestore (Event data)
		- Database
- Vroom (Profiles)
	- Docker volume
	- Object Storage

We should probably either rename those section to what the specifically store or explain that in the intro because "Vroom" is not very descriptive but if it's explained that that component is responsible for (ingest and) storing profiling data it makes a lot more sense.

Maybe with until someone else also chimes in before rewriting the whole thing in case I'm off base with this outline but this sounds like a document I would love to have had when I started my self-hosted adventures 👍

src/docs/self-hosted/external-storage.mdx
```diff
@@ -0,0 +1,89 @@
...
+  
+  I got some time before Monday to write up some docs about setting up an S3 storage for selfhosted instance, but I can't decide whether I should put it under a big "External Services" page, in which people can include external postgres, external redis, and that kind of things; or should I put it under a page called "External Storage"?
+
+  There. Please help me decide this. I'll delete this comment afterwards -->
+
+<Alert title="Note" level="info">
+    After changing configuration files, re-run the <code>./install.sh</code> script, to rebuild and restart the containers. See the <Link to="/self-hosted/#configuration">configuration section</Link> for more information.
+</Alert>
+
+<!-- Should we add a description about what "external storage" is? -->
```
hubertdeng123, line 15: I think we can safely assume that people can find that out for themselves if they are looking at this page

src/docs/self-hosted/external-storage.mdx
```diff
@@ -0,0 +1,89 @@
+---
+title: External Storage
+---
+
+<!-- Hello! If you're reading this, you're in luck because I can't decide whether to make.. wait let me copy the text from Discord.
+  
+  I got some time before Monday to write up some docs about setting up an S3 storage for selfhosted instance, but I can't decide whether I should put it under a big "External Services" page, in which people can include external postgres, external redis, and that kind of things; or should I put it under a page called "External Storage"?
```
hubertdeng123, line 7: External storage sounds fine. I don't think we should recommend people to use external postgres, redis, etc as that can introduce a lot of issues for people trying to set that up unless they really know what they're doing.

src/docs/self-hosted/external-storage.mdx
```diff
@@ -0,0 +1,89 @@
...
+    After changing configuration files, re-run the <code>./install.sh</code> script, to rebuild and restart the containers. See the <Link to="/self-hosted/#configuration">configuration section</Link> for more information.
+</Alert>
+
+<!-- Should we add a description about what "external storage" is? -->
+
+## Filestore
+
+Filestore handles storing attachment, sourcemap, and replays. Filestore configuration for Sentry should be configured on the `sentry/config.yml` file.
+
+### S3 backend
```
hubertdeng123, line 21: Just to make sure, have you tried the Azure/s3 compatible backend without issues? We're using GCS so wanted to make sure

src/docs/self-hosted/external-storage.mdx
aldy505: I agree with "don't think we should recommend people to use external postgres, redis, etc", but if there are some people who wishes to do that... I don't know if there's any other better way to say to them that "they're on their own"

src/docs/self-hosted/external-storage.mdx
aldy505: Yes! See my patch https://github.com/getsentry/self-hosted/compare/master...teknologi-umum:sentry:teknum-patch

src/docs/self-hosted/external-storage.mdx
```diff
@@ -0,0 +1,89 @@
+---
+title: External Storage
+---
+
+<!-- Hello! If you're reading this, you're in luck because I can't decide whether to make.. wait let me copy the text from Discord.
+  
+  I got some time before Monday to write up some docs about setting up an S3 storage for selfhosted instance, but I can't decide whether I should put it under a big "External Services" page, in which people can include external postgres, external redis, and that kind of things; or should I put it under a page called "External Storage"?
+
+  There. Please help me decide this. I'll delete this comment afterwards -->
```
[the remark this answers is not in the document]
aldy505: > because realistically we can't offer very good support for it, and folks will be left disappointed.

I know this, but I think the current S3 support is good enough for selfhosted. Users can always pull out their own Django plugins though. Like @stayallive's S3 Nodestore plugin https://github.com/stayallive/sentry-nodestore-s3

> we need to be very clear that this is provided as a rough best effort template

I agree. Need to take some time to come up with good enough copywriting for this lol.

src/docs/self-hosted/external-storage.mdx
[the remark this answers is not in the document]
aldy505: I think it's something that's not strictly on the same filesystem as the sentry self-hosted instance. But it also excludes if you want to use something like NAS or external bind-mount storage to store Sentry data. For files, it's the blob storage provided by each cloud provider. For databases, it's external database that's either managed or unmanaged, but it should be separate to the sentry self-hosted instance. Do you have any suggestion on how to better phrase this out?

src/docs/self-hosted/external-storage.mdx
```diff
@@ -0,0 +1,84 @@
+---
+title: External Storage
+---
+
+In some cases, storing Sentry data on-disk is not really something people can do. Sometimes, it's better if they can offload it into some bucket storage (like AWS S3 or Google Cloud Storage).
```
aldy505, line 5: This seems a bit confusing, but I think adding another page later after this about "Unsupported Workflows" in which we can specify more about what kind of things that we can't offer support to (external Redis, external Postgres, installing third party plugins for extending some stuff).

See @azaslavsky's comment here https://github.com/getsentry/develop/pull/1269#discussion_r1598919946

src/docs/self-hosted/external-storage.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+Vroom is the service that handles profiling. By default the data for profiling is saved on local filesystem. On self-hosted deployment, this should be done by overriding the `SENTRY_BUCKET_PROFILES` environment variable. It's also possible that additional environment variables should be added, depending on the backend of choice.
+
+### Google Cloud Storage backend
+
+You will need to set `GOOGLE_APPLICATION_CREDENTIALS` environment variable. For more information, refer to the [Google Cloud documentation for setting up authentication](https://cloud.google.com/storage/docs/reference/libraries#setting_up_authentication).
+
+```bash
+gs://my-bucket
+```
```
aldy505, line 59: I haven't tested this. I only know how to configure this. Can you guys test this out on your dogfood instance?

PR #1270, opened 2024-05-13
docs(trace-view): Updating docs referencing the trace view.

Abdkhan14:
We made the new trace view generally available and we can load spans in it.

### DIFF
--- a/src/docs/sdk/research/performance/index.mdx
+++ b/src/docs/sdk/research/performance/index.mdx
@@ -269,13 +269,7 @@ Sentry's ingestion model was not designed for nested transactions within a servi
 
 In practice, SDKs have no way of preventing transactions from becoming nested. The end result is possibly surprising to users, as each transaction starts a new tree. The only way to relate those trees is through the `trace_id`.
 
-Sentry's billing model is per event, be it an error event or a transaction event. That means that a transaction within a transaction generates two billable events.
-
-In SDKs, having a transaction within a transaction will cause inner spans to be "swallowed" by the innermost transaction surrounding them. In these situations, the code creating the spans will only add them to one of the two transactions, causing instrumentation gaps in the other.
-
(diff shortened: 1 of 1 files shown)


src/docs/sdk/research/performance/index.mdx
```diff
@@ -290,6 +284,12 @@ with transaction("index-page"):
                 create_thumbnail(result)
 ```
 
+Sentry's billing model is per event, be it an error event or a transaction event. That means that a transaction within a transaction generates two billable events.
+
+In SDKs, having a transaction within a transaction will cause inner spans to be "swallowed" by the innermost transaction surrounding them. In these situations, the code creating the spans will only add them to one of the two transactions, causing instrumentation gaps in the other.
+
+To visualize all transactions that share a `trace_id`, a user can visit the Trace View page in **Sentry > Performance**. This page gives an overview of the trace by displaying transactions using hierarchical UI and allows users to load child spans. To learn more about the Trace View, checkout our [docs](https://docs.sentry.io/concepts/key-terms/tracing/trace-view/).
```
lizokm, line 291:
```suggestion
To visualize all transactions that share a `trace_id`, a user can visit the Trace View page in **Sentry > Performance**. This page gives an overview of the trace by displaying transactions using hierarchical UI and allows users to load child spans. To learn more about the Trace View, check out our [docs](https://docs.sentry.io/concepts/key-terms/tracing/trace-view/). 
```

lizokm (approved):
This looks fine to me in terms of words. My only concern is that we're updating the billing model and the billing unit will be spans not transactions so we'll have to update that at some point

Abdkhan14:
> This looks fine to me in terms of words. My only concern is that we're updating the billing model and the billing unit will be spans not transactions so we'll have to update that at some point

@lizokm yes I'm aware of this. We'll update the docs when we get rid of transactions.

PR #1279, opened 2024-05-15
docs(self-hosted): Add missing slash in caddy reverse-proxy path_regexp example

lukasbableck:
This took me wayyy to long to realize...


aldy505 (approved):
Welp, I'm sorry lol (the git blame was on me). Thanks for the fix!

aldy505:
cc @hubertdeng123 @azaslavsky

PR #1281, opened 2024-05-17
feat(performance): Web Vitals transactions and spans documentation

edwardgou-sentry:
Adds some dev docs around webvital transaction and span conventions


cleptric (approved):
Thanks!

PR #1282, opened 2024-05-17
feat(insights): Add documentation for Queries module

gggritso:
👋🏻 the Query Source stuff is an omission, but I can add that later both here, and on the page that describes the options the SDK should support.


antonpirker (approved):
Nice!

PR #1288, opened 2024-05-21
test(ui): Add examples of overriding default contexts

scttcper:


evanpurkhiser (approved):

ryan953 (approved):
Put it higher in the file?

I would argue that the order goes:
- render() first
- query for stuff in the result

PR #1293, opened 2024-05-27
docs(app-starts): Add SDK requirements

narsaynorath:
Reflects what conditions are required for ingestion of app start metrics in relay to support the app start module.

I had to split out the Android specific ops because iOS uses `app.start.cold` or `app.start.warm` and varies the description to identify different operations.


cleptric (approved):
Thanks!

PR #1296, opened 2024-05-30
docs(analytics): Add info about testing locally

roggenkemper:
this pr adds a line about testing analytics events locally


mrduncan (approved):
🙌

PR #1302, opened 2024-06-06
Added missing categories

antonpirker:
Added three missing categories, sorted the list of categories and formatted the doc.


antonpirker:
As this is a tiny change, I am merging it right away.

PR #1307, opened 2024-06-11
doc(metrics): Metric tags can only have a single value

jjbayer:
Neither Relay nor storage support multi-value tags.

Fixes https://github.com/getsentry/relay/issues/3691.

### DIFF
--- a/src/docs/sdk/metrics.mdx
+++ b/src/docs/sdk/metrics.mdx
@@ -43,7 +43,7 @@ exist:
 * **Background Aggregation:** SDKs are encouraged to defer flushing and aggregation
   into a background thread.  For SDKs where `fork()` is to be expected, the background
   thread needs to ensure it restarts after fork (eg: python).
-* **Tagging:** metrics are tagged by key value pairs. Both keys and values are limited to strings,
+* **Tagging:** metrics can be tagged with a unique key and an optional value for each key. Both keys and values are limited to strings,
   but the SDK can expose other types if it has a reasonable way to stringify these.
 
 In abstract terms the aggregator is recommended to look a bit like this:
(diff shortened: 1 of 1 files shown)


Dav1dde (approved):
Technically this still allows the interpretation of having two pairs `(a, 1)` and `(a, 2)` for the same metric.

jjbayer:
> Technically this still allows the interpretation of having two pairs (a, 1) and (a, 2) for the same metric.

@Dav1dde what do you suggest?

src/docs/sdk/metrics.mdx
```diff
@@ -43,9 +43,8 @@ exist:
 * **Background Aggregation:** SDKs are encouraged to defer flushing and aggregation
   into a background thread.  For SDKs where `fork()` is to be expected, the background
   thread needs to ensure it restarts after fork (eg: python).
-* **Tagging:** metrics are tagged by key value pairs, where each key can have more
-  than one value.  Both keys and values are limited to strings, but the SDK can
-  expose other types if it has a reasonable way to stringify these.
+* **Tagging:** metrics are tagged by key value pairs. Both keys and values are limited to strings,
```
iker-barriocanal, line 46: LGTM.

re: David's suggestion, maybe the following:
```suggestion
* **Tagging:** metrics are tagged by unique key value pairs. Both keys and values are limited to strings,
```
but I think it's ok to merge as-is.

iker-barriocanal (approved):

src/docs/sdk/metrics.mdx
Dav1dde: `(a, 1)` and `(a, 2)` are still unique pairs.

Dav1dde, line 46:
```suggestion
* **Tagging:** metrics can be tagged with a unique key and an optional value for each key. Both keys and values are limited to strings,
```

Maybe something like this, since there can be *no* tags at all and tags *without* values.

PR #1315, opened 2024-06-21
feat: Remove DDM

matejminar:
This PR removes `Sentry Developer Metrics` section as most of it is already available on docs.sentry.io in a more up-to-date version. Some parts there were still relevant has been moved to other parts of the develop docs (metrics abstraction).

Closes https://github.com/getsentry/sentry/issues/71971


src/docs/sdk/metrics.mdx
```diff
@@ -6,8 +6,7 @@ sidebar_order: 13
 Sentry supports the emission of pre-aggregated metrics information via the `statsd`
 envelope item.  This emission bypasses relay-side sampling and assumes a client side
 aggregator.  In the product these are referred to as "custom metrics" and internally
-the system sometimes carries the name "ddm" ([delightful developer
-metrics](/delightful-developer-metrics/)).  The functionality sits on top of the
+the system sometimes carries the name "DDM". The functionality sits on top of the
```
shellmayr, line 9: Is there a reason for keeping this around when any documentation of DDM is removed?

shellmayr (approved):

vgrozdanic (approved):

src/docs/sdk/metrics.mdx
matejminar: Yeah, I thought about it. Some people still internally use word DDM and these docs are used for both internal and external readers so I figured there's no harm in keeping it to explain that this is actually what's known as _DDM_. No strong opinions though, this section might get removed completely as we move to the new spans world.

PR #1316, opened 2024-06-24
SDK Philosophy - Write less code

HazAT:
Another addition for our SDK Philosophy

### DIFF
--- a/src/docs/sdk/philosophy.mdx
+++ b/src/docs/sdk/philosophy.mdx
@@ -57,3 +57,7 @@ Please check <Link to="/sdk/data-handling">Data Handling</Link> for more detail.
 ## Don’t forget the big picture
 
 Being correct is important but even more important is doing the right thing. Yes, we are building libraries and with that comes much responsibility. But in the end, we build a product that should solve a problem users have in their application. This principle relates to many other things mentioned on this page - but most importantly, it should act as a reminder that we only succeed if we manage to solve a problem end-to-end.
+
+## Write less code
+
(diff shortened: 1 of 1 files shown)


bruno-garcia (approved):
:shipit:

src/docs/sdk/philosophy.mdx
```diff
@@ -57,3 +57,7 @@ Please check <Link to="/sdk/data-handling">Data Handling</Link> for more detail.
 ## Don’t forget the big picture
 
 Being correct is important but even more important is doing the right thing. Yes, we are building libraries and with that comes much responsibility. But in the end, we build a product that should solve a problem users have in their application. This principle relates to many other things mentioned on this page - but most importantly, it should act as a reminder that we only succeed if we manage to solve a problem end-to-end.
+
+## Write less code
+
+Only do in SDKs what can only be done in SDKs. Data quality is one of the most important things. Without it, we can't build a great product experience. The purpose of the SDK is to instrument the user's application, collect context around the event type we want to capture, and send it to Sentry. Any extensive business logic that does changes to the data should be avoided; the data should be sent as raw as possible. Business logic in SDKs is complex, hard to maintain and even harder to change and becomes baked into the data forever. We have way more flexibility on the server.
```
AbhiPrasad, line 63:
```suggestion
The role of an SDK is to instrument the user's application, collect context around the event types we want to capture, and send event data with useful context to Sentry. We should avoid extensive business logic within the SDK that transforms this data as it can get complex, hard to maintain and makes permanent changes to the wire format we send events. We should aim to have data collected by the SDK be as raw as possible. We have more flexibility and control on the server to manipulate the data more effectively.
```

AbhiPrasad (approved):

cleptric (approved):

lforst:
Love it. The one question I had relates to "preemptive" filtering. Considering this philosophy, should we have code in the SDK, for example, to filter out known noisy errors?

HazAT:
@lforst well we do filtering on the server already - but there are still cases where we need to do filtering in the SDK because the technology we have on the server today is not there yet to make the best possible decision for data quality -> Tailbased Sampling / Buffering

Until we have that, we need to make some tradeoffs - but the ultimate goal is that SDKs collect raw data and send it.

PR #1317, opened 2024-06-24
fix: Improve ngrok setup docs

markstory:
Clarify how ngrok should be setup for siloed and non-siloed modes.

Refs getsentry/sentry#14423

### DIFF
--- a/src/docs/backend/development-server.mdx
+++ b/src/docs/backend/development-server.mdx
@@ -27,3 +27,31 @@ This will do literally nothing except bring up the web workers. You are now resp
 * Starting Celery workers using `sentry run worker` to run most kinds of background jobs.
 
 Use `sentry run --help` to see what you might be missing by not running devserver.
+
+
+## Using ngrok
+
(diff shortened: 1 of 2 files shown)


src/docs/backend/development-server.mdx
```diff
@@ -27,3 +27,31 @@ This will do literally nothing except bring up the web workers. You are now resp
...
+
+The development server can be operated behind ngrok for when you need to receive traffic from the outside world. This comes up frequently when working with integrations. Start off by defining a configuration file for `ngrok`:
+
+```yaml
+version: '2'
+authtoken: <YOUR-NGROK-AUTHTOKEN>
+tunnels:
+  acme-org
+    proto: http
+    hostname: $yourname.ngrok.dev
```
vartec, line 42: nit: I'd use `<yourname>` or just `yourname`, because with `$` it kind of looks like you'd be able to use an env var, which won't actually work in a YAML.

vartec (approved):

PR #1320, opened 2024-06-26
feat: Add docs for missing instrumentation context

chargome:
Note: Should only be merged once https://github.com/getsentry/sentry-javascript/pull/12639 is released.


AbhiPrasad (approved):
Realized I forgot to come back to this!

Let's merge this in now that we have released this.

PR #1330, opened 2024-07-09
fix(environment): restructure subheadings and clarify ingest setup

mifu67:
The subsubheadings 'Frontend Only & Backend Only'/'Enabling HTTPS'/'Ingestion Pipeline (Relay)' are not specific to Getsentry, so it makes more sense for them to be structured under the Sentry subhead.

Additionally clarify which settings should be used when configuring the devserver for error ingestion vs metrics ingestion.

### DIFF
--- a/src/docs/environment/index.mdx
+++ b/src/docs/environment/index.mdx
@@ -202,7 +202,7 @@ environment, you can create `getsentry/conf/settings/devlocal.py` and put the
 configuration option overrides there. This module will be automatically imported
 by `dev.py` if it exists.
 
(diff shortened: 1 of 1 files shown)


ceorourke:
What do you think about moving the "Running the Getsentry Development Server" section to just below "Running the Development Server"? So it'd be 1. Sentry 2. Getsentry 3. Running Siloed Instances

I'm guessing it used to be in that order before the siloed instances part got added because the gs part says "Just like running sentry (see above)," but above is now about silos.

ceorourke (approved):

PR #1334, opened 2024-07-15
Specify extension and format

otbutz:
`update-ca-certificates` is a bit picky and will silently ignore files without the `.crt` extension.


aldy505 (approved):

aldy505:
@hubertdeng123 @azaslavsky

PR #1336, opened 2024-07-18
feat(self-hosted): Add docs for self-hosted errors only

hubertdeng123:
Steps to tell users how to enable this mode of self-hosted Sentry

### DIFF
--- a/src/docs/self-hosted/index.mdx
+++ b/src/docs/self-hosted/index.mdx
@@ -107,6 +107,31 @@ By default Sentry sends anonymous usage statistics to the Sentry team. It helps 
 
 You can disable this feature by adding `SENTRY_BEACON = False` into `sentry.conf.py` file.
 
+### Errors-only Self-hosted Sentry
+
+This feature is experimental and now in beta. Starting from 24.8.0+ onwards, users will have the option to choose between two types of self-hosted Sentry deployments.
+
+**Errors Only**
+This is an opt in version of self-hosted Sentry that is much more lightweight than the feature complete version. This includes features around Sentry's bread and butter, issues.
+1. Issues
+2. Alerts
+3. Integrations
+4. Dashboards
+5. Releases
+6. Discover
+
+**Feature Complete**
+This is the default version of self-hosted Sentry that includes many of the features that we offer on SaaS. This includes everything offered in the errors only version but also additional main features listed below!
+1. Traces
+2. Profiles
+3. Replays
+4. Insights (Requests, Queries, Assets, etc)
+5. User Feedback
+6. Performance
+7. Crons
+
+In order to enable errors only self-hosted, you'll need to update your [.env file](https://github.com/getsentry/self-hosted/blob/master/.env) to include `COMPOSE_PROFILES=errors-only`.
+
 ## Configuration
 
 You very likely will want to adjust the default configuration for your Sentry installation as well. These facilities are available for that purpose:


src/docs/self-hosted/index.mdx
```diff
@@ -107,6 +107,31 @@ By default Sentry sends anonymous usage statistics to the Sentry team. It helps
 
 You can disable this feature by adding `SENTRY_BEACON = False` into `sentry.conf.py` file.
 
+### Errors-only Self-hosted Sentry
+
+This feature is experimental and now in beta. Starting from 24.8.0+ onwards, users will have the ability to choose between two distinct types of self-hosted Sentry deployments, tailored to their specific needs and preferences.
```
azaslavsky, line 112: nit: first sentence should be its own bolded paragraph: "This is an experimental alpha feature. Use at your own risk!"

azaslavsky, line 112: nit: "24.8.0+ onwards" is redundant - either 24.8.0+ or 24.8.0 onwards

azaslavsky, line 112: nit: remove "tailored to their specific needs and preferences" imo

src/docs/self-hosted/index.mdx
```diff
@@ -107,6 +107,31 @@ By default Sentry sends anonymous usage statistics to the Sentry team. It helps
...
+
+**Errors Only**\n
+Opt for a sleek and efficient version of self-hosted Sentry that focuses on what we do best—tackling issues head-on. This lightweight option includes these features:
```
azaslavsky, line 116: This paragraph is written directly addressing the user, while the one in the next section is written in a more passive style. We should pick one and stick with it.

azaslavsky, line 116: nit: "best-tackling" -> "best: tackling"

azaslavsky, line 116: "includes only the following features"

src/docs/self-hosted/index.mdx
```diff
@@ -107,6 +107,31 @@ By default Sentry sends anonymous usage statistics to the Sentry team. It helps
...
+1. Issues
+2. Alerts
+3. Integrations
+4. Dashboards
+5. Releases
+6. Discover
+
+**Feature Complete**\n
+This is our default version of self-hosted Sentry that is most similar feature parity wise to our SaaS product. This includes everything offered in the Errors Only version, as well as many other amazing features.
```
azaslavsky, line 126: This should be two sentences.
"This is the default version of self-hosted Sentry. It includes most of the features available in our SaaS product."

azaslavsky, line 126: This is a developer doc, not marketing. "As well as many other amazing features" is laying it one way too thick imo.

"Feature Complete Sentry includes everything available in the Errors Only version, plus the following:"

src/docs/self-hosted/index.mdx
```diff
@@ -107,6 +107,31 @@ By default Sentry sends anonymous usage statistics to the Sentry team. It helps
...
+1. Traces
+2. Profiles
+3. Replays
+4. Insights (Requests, Queries, Assets, etc)
+5. User Feedback
+6. Performance
+7. Crons
+
+In order to enable errors only self-hosted, you'll need to update your [.env file](https://github.com/getsentry/self-hosted/blob/master/.env) to include `COMPOSE_PROFILES=errors-only`.
```
azaslavsky, line 135: This should go in the `Errors Only` section above. This section should read something like:

"To use Feature Complete Sentry, you'll need to ensure that you are not setting the `COMPOSE_PROFILES` environment variable (usually set via your `.env` file) when running Sentry via docker compose."

src/docs/self-hosted/index.mdx
azaslavsky: Also, I think this should be more technical and less markety. "Errors Only Sentry offers a small set of core features, with an eye on minimizing resource costs."

src/docs/self-hosted/index.mdx
hubertdeng123: yeah you're right, changed

src/docs/self-hosted/index.mdx
```diff
@@ -107,6 +107,37 @@ By default Sentry sends anonymous usage statistics to the Sentry team. It helps
...
 
+## Errors-only Self-hosted Sentry
+
+**Disclaimer: This is an experimental beta feature. This means that features and workflows are not completely tested, so use at your own risk!**
+
+Starting from 24.8.0+, users will have the ability to choose between two distinct types of self-hosted Sentry deployments.
+
+**Errors Only**
+
+Errors Only self-hosted offers a small subset of features with an emphasis on minimizing system resources while focusing on what we do best: tackling issues head-on. This lightweight option includes only the following features:
```
azaslavsky, line 118:
```suggestion
Errors Only self-hosted offers a small subset of features with an emphasis on minimizing system resources. This lightweight option includes only the following features:
```

src/docs/self-hosted/index.mdx
```diff
@@ -107,6 +107,37 @@ By default Sentry sends anonymous usage statistics to the Sentry team. It helps
...
+1. Issues
+2. Alerts
+3. Integrations
+4. Dashboards
+5. Releases
+6. Discover
+
+In order to enable errors only self-hosted, you'll need to update your [.env file](https://github.com/getsentry/self-hosted/blob/master/.env) to include `COMPOSE_PROFILES=errors-only`.
```
azaslavsky, line 126: This applies to existing install too, right? Will it apply in reverse (that is, if I insert this flag then take it off, will that cause issues?)? If not, we should be very clear about that.

azaslavsky (approved):

src/docs/self-hosted/index.mdx
hubertdeng123: Yes, applies for existing installs. If you insert this flag and use errors-only self-hosted and want to switch back to feature complete sentry there shouldn't be issues.

PR #1341, opened 2024-07-26
docs(self-hosted): provide product links to ommited features on errors-only

aldy505:
Since some people has been asking about what the feature is, but we don't exactly point them to the correct product documentation.

cc @hubertdeng123 @IanWoodard


hubertdeng123 (commented):
We will need to make these change sin https://github.com/getsentry/sentry-docs/blob/master/develop-docs/self-hosted/index.mdx instead

PR #1346, opened 2024-08-13
fix(build): fix on mac arm

joshuarli:
i can't build these docs locally on mac arm because @sentry/profiling-node 1.1.2 mispackaged executables like this:

```
node_modules/@sentry/profiling-node/lib/sentry_cpu_profiler-darwin-arm64-83.node:  ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, not stripped
```

this was fixed in @sentry/profiling-node 1.2.0


asottile-sentry (approved):
wcgw

